PR #4, opened 2017-08-23
Feature/lnd ipc renderer

jackmallers:
This removes the zap-nodejs library completely, no more express app needed. It leverages ipcRenderer within' electron. Makes the app much more friendly to non-technical users. Inching closer and closer to a great user experience


fresheneesz:
@jackmallers Could you update the Readme to reflect your changes? Documentation changes should be part of commits that change the installation process or functionality for things that have instructions that appear in the readme.

jackmallers:
@fresheneesz yeah that was mental error on my part. doing it now. thanks for pointing it out

jackmallers:
@fresheneesz updated

fresheneesz:
: D

PR #7, opened 2017-08-25
Fix Added location to lnd tls certificate file for windows

funyug:
Added location to lnd tls certificate file for windows


jackmallers:
Nice. Merged.

PR #14, opened 2017-08-30
Changed linux instruction for tls.cert

callmekurisu:
~/.lnd/tls.cert path did not work for tls.cert but /home/user/.lnd/tls.cert did


funyug:
The ~ refers to current user's directory same as /home/user/
Don't know why that didn't worked for you.

callmekurisu:
yeah i know. it was strange. I have all the paths setup correctly

jackmallers:
Hm, that's strange. I agree with @funyug and ~/.lnd works for me on my Linux machine.

Do any paths with `~` work for you?

callmekurisu:
Yeah others do. Will find keep looking for this path error. Closing this request. Domo!

PR #19, opened 2017-09-05
Escape space in Mac Lnd/ path in README

tbloncar:
This makes a small change to the README to escape the space in the path to `Lnd/`. It also parallels the Linux path and leverages the tilde to reference the home directory.


jackmallers:
Good call. Merged

PR #20, opened 2017-09-05
Format ticker USD price with dollar symbol and two decimal places

tbloncar:
This adds a utility for formatting a USD price so that it includes the dollar symbol and two decimal places. It uses this utility to format the ticker USD price.

Note: This was primarily an exercise in familiarizing. It may be worthwhile to leverage something like [accounting](https://github.com/openexchangerates/accounting.js) to do this sort of thing. This would [also handle](https://github.com/openexchangerates/accounting.js/commit/9d65e654a7cd4db0248bbeb99388642b3517f8fb) the native [toFixed rounding issue](https://stackoverflow.com/questions/10015027/javascript-tofixed-not-rounding).

Before:

screen shot 2017-09-05 at 12 26 46 am

After:

screen shot 2017-09-05 at 12 23 09 am


jackmallers:
Nice! Merged.

PR #26, opened 2017-09-09
LND backend refactoring

joaodealmeida:
Some code refactoring and new Lnd endpoints.

@JimmyMow  take a look!


app/lnd/config/index.js
```diff
@@ -5,5 +5,5 @@
 export default {
   lightningRpc: `${__dirname}/rpc.proto`,
   lightningHost: 'localhost:10009',
-  cert: '/Users/jmow/Library/Application Support/Lnd/tls.cert'
+  cert: '/Users/joaoalmeida/Library/Application\ Support/Lnd/tls.cert'
```
jackmallers: How should we handle this? Should remove anyones username probably

app/lnd/methods/invoicesController.js
```diff
@@ -0,0 +1,84 @@
...
+
+
+/**
+ * Attemps to look up an invoice according to its payment hash
+ * @param  {[type]} lnd   [description]
+ * @param  {[type]} rhash [description]
+ * @return {[type]}       [description]
+ */
+export function lookupInvoice(lnd, { rhash }) {
+  return new Prommise ((resolve, reject) => {
```
jackmallers: double "m" in this Promise

app/lnd/methods/networkController.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+
+/**
+ * Returns a description of the latest graph state from the point of view of the node
+ * @param  {[type]} lnd [description]
+ * @return {[type]}     [description]
+ */
+export function describeGraph(lnd) {
+  return new Promise ((resolve, reject) => {
```
jackmallers: Some small inconsistencies with space before `((resolve, reject) => ...`

I don't have a preference but let's pick no space or space and keep a standard

app/lnd/config/index.js
thisconnect: `os.homedir()` https://nodejs.org/api/os.html#os_os_homedir  ?

app/lnd/methods/networkController.js
thisconnect: `prettier` would kill the space! https://www.npmjs.com/package/prettier

app/lnd/config/index.js
joaodealmeida: isn't os.userInfo().username better in this case ? @thisconnect

jackmallers: @joaodealmeida `os.userInfo().username` works for me locally

joaodealmeida: I got it working here as well, just wanted to make sure what was the correct function to use here. But I guess I'll go with userInfo().username since its working pretty well.

jackmallers: cool

app/lnd/config/index.js
```diff
@@ -2,8 +2,10 @@
 // Mac OS X: /Users/user/Library/Application Support/Lnd/tls.cert
 // Linux: ~/.lnd/tls.cert
 // Windows: TODO find out where cert is located for windows machine
+import { userInfo } from 'os'
+
 export default {
   lightningRpc: `${__dirname}/rpc.proto`,
   lightningHost: 'localhost:10009',
-  cert: '/Users/joaoalmeida/Library/Application\ Support/Lnd/tls.cert'
+  cert: '/Users/'+ userInfo().username+'/Library/Application\ Support/Lnd/tls.cert'
```
jackmallers: again, super picky, but this should be \`/Users/${userInfo().username+}/Library/Application\ Support/Lnd/tls.cert\`

:)

app/lnd/config/index.js
```diff
@@ -2,8 +2,10 @@
 // Mac OS X: /Users/user/Library/Application Support/Lnd/tls.cert
 // Linux: ~/.lnd/tls.cert
 // Windows: TODO find out where cert is located for windows machine
+import { userInfo } from 'os'
+
 export default {
   lightningRpc: `${__dirname}/rpc.proto`,
   lightningHost: 'localhost:10009',
-  cert: '/Users/jmow/Library/Application Support/Lnd/tls.cert'
+  cert: '/Users/'+ userInfo().username+'/Library/Application\ Support/Lnd/tls.cert'
```
jackmallers: This is super picky haha bit this line should use es6 like this:

`/Users/${userInfo().username}/Library/Application\ Support/Lnd/tls.cert`

:)

PR #32, opened 2017-09-21
Feature/listpayments

jackmallers:


jackmallers:
Will QA again tomorrow and do a final review/touch up before merge

PR #39, opened 2017-09-25
Feature/setalias

jackmallers:


jackmallers:
add setAlias to the back end

PR #40, opened 2017-09-25
feature(eslint-resolve-paths-from-app-dir): use 'app' path as default…

VonIobro:
… for imports

get rid of those ../../../../../paths^^

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -34,6 +34,9 @@
   ],
   "settings": {
     "import/resolver": {
+      "node": {
+        "paths": ["app"]
+      },
       "webpack": {
         "config": "webpack.config.eslint.js"
       }
--- a/app/routes/activity/components/components/Invoice/Invoice.js
+++ b/app/routes/activity/components/components/Invoice/Invoice.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import Moment from 'react-moment'
 import 'moment-timezone'
 import { FaBolt, FaClockO } from 'react-icons/lib/fa'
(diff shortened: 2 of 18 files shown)


jackmallers:
Nice! Do you wanna replace all annoying paths in this PR or merge the fix (in `.eslintrc`) and I can handle replacing all annoying paths later? @VonIobro

VonIobro:
I can replace all the annoying paths in this PR. No problem.

VonIobro:
The above commits have been tested and are ready to be merged. The paths for the .scss files have not been changed because of the problem below.

There seems to be a problem with webpack/css-loader/sass-loader not recognizing the resolved paths. While preparing to `npm run test`, `npm run build-renderer` fails when resolved-paths for scss imports: 
```
 ERROR in ./node_modules/css-loader?{"modules":true,"importLoaders":1,"localIdentName":"[name]__[local]__[hash:base64:5]"}!./node_modules/sass-loader/lib/loader.js!./app/routes/wallet/components/Wallet.scss
    Module build failed: 
    @import 'variables.scss';
    ^
          File to import not found or unreadable: variables.scss.
```

jackmallers:
Beautiful @VonIobro, nice first contribution! Yeah if you wanna create an issue we can take care of that issue at some point.

I actually don't mind styled components myself. One day when this project is in a much more stable place and goes through a large refactor we can make that switch

thisconnect:
this is non standard esmodule path resolving correct? in case you want to use another builder this might not work anymore

PR #48, opened 2017-09-28
feature(show-frame): show browser window frame

VonIobro:
This shows the title bar so that user can reposition the app in their OS easily.


jackmallers:
I've never liked the look of the default frame, I removed it from the app for design purposes. I do agree the app should be resizable and draggable. Lets make this a little more visually appealing

PR #51, opened 2017-09-29
Feature/LND_Neutrino-packaging

joaodealmeida:
This PR has the goal to make it easier for an average user to use the application. An user should click on an executable to open the app. This PR handles all the dependencies behind the application such as homebrew, go, glide and lnd installations as well as reporting the live status of lnd to the application.

- [x] macOS.
- [x] Linux (lacks test)
- [ ] Windows
- [ ] Create executables with [electron-builder](https://github.com/electron-userland/electron-builder)

to run:
$ yarn
$ ./node_modules/.bin/electron-rebuild
$ npm run dev


joaodealmeida:
To deal with the sudo problem when installing go, I would suggest to manually install it.
Something like [this ](https://golang.org/doc/install)

app/main.dev.js
```diff
@@ -54,45 +54,80 @@ app.on('window-all-closed', () => {
...
+
+    if (!results.length) {
+      // Alert user that LND is starting
+      console.log('STARTING LND')
+
+      // Start LND
+      neutrino = spawn(
+        'lnd',
+        [
+          '--bitcoin.active',
```
VonIobro: I've been using lnd without the `--bitcoin.active` flag, just `--bitcoin.testnet`. After perusing this PR, I decided to try adding the bitcoin.active flag. It makes my previous onchain (testnet) tx-history disappear because I think it uses a different folder for the wallet.dat.

joaodealmeida: Do you have your lnd updated ?

joaodealmeida:
This PR can be closed since it has been replaced by PR #59 .

PR #77, opened 2017-11-22
Linux support

torkelrogstad:
This commit adds support for Linux, this was the only thing I had to change to get it up and running.

### DIFF
--- a/app/lnd/config/index.js
+++ b/app/lnd/config/index.js
@@ -2,10 +2,26 @@
 // Mac OS X: /Users/user/Library/Application Support/Lnd/tls.cert
 // Linux: ~/.lnd/tls.cert
 // Windows: TODO find out where cert is located for windows machine
-import { userInfo } from 'os'
+import { userInfo, platform } from 'os'
+import { join } from 'path'
+
+let loc
+switch (platform()) {
+  case 'darwin':
+    loc = 'Library/Application\ Support/Lnd/tls.cert'
+    break
+  case 'linux':
+    loc = '.lnd/tls.cert'
+    break
+  case 'win32':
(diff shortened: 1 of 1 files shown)


JimmyMow:
Nice! Thank you @torkelrogstad, all of us have Macs and totally got outta whack for multiple OS support. I am gonna be super picky with code style haha hope you don't mind. If you want I can make the changes myself, lemme know.

app/lnd/config/index.js
```diff
@@ -2,10 +2,17 @@
 // Mac OS X: /Users/user/Library/Application Support/Lnd/tls.cert
 // Linux: ~/.lnd/tls.cert
 // Windows: TODO find out where cert is located for windows machine
-import { userInfo } from 'os'
+import { userInfo, platform } from 'os'
+
+var loc;
```
JimmyMow: let's use `let` instead of `var`

we also have decided to avoid semicolons in this codebase 😄

app/lnd/config/index.js
```diff
@@ -2,10 +2,17 @@
...
+if (platform() == 'linux') {
```
JimmyMow: as long as we are gonna add support for other OS we should add `win` here as well. Something like:

```
switch (platform()) {
  case 'darwin':
    loc = path.join(homedir, 'Library/Application\ Support/Lnd/tls.cert')
    break
  case 'linux':
    loc = path.join(homedir, '.lnd/tls.cert')
    break
  case 'win32':
    loc = path.join(homedir, 'AppData', 'Local', 'Lnd', 'tls.cert')
    break
  default:
    break
}
```

torkelrogstad:
Better?

JimmyMow:
Yep! Nice, merging

PR #85, opened 2017-12-06
Lint

torkelrogstad:
This PR fixes a great deal of the complaints from `npm run lint`. The rest I'm not sure about yet, and have to look into. There's still also a complaint from `npm run test`, about a missing file. I believe this is because something got deleted earlier?

To avoid such things happening, and ensuring lint and test compatibility in the future, would it be an idea to set up continuous integration (CI) from for example [Jenkins](https://jenkins.io/solutions/github/)?


Archcry:
You can use Travis, it's free for opensource projects.

JimmyMow:
Thanks for the PR @torkelrogstad, merging. 

Yes, linting errors have gotten a little out of hand as we are trying to rush a testnet MVP to the people, thanks for this. I agree completely that as the source code is coming to a more stable position something like Travis is needed. I can make an issue for it?

torkelrogstad:
Making an issue would be great, I could give it a shot at setting it up.

PR #92, opened 2017-12-09
Generate witness addresses instead of p2pkh.

jimpo:
lnd requires inputs to funding transactions to spend from pay-to-witness outputs.


JimmyMow:
Nice. Wanna make the same change here? 

https://github.com/LN-Zap/zap-desktop/blob/master/app/routes/app/components/App.js#L17

Would be better if both changes are in the same PR

jimpo:
Thanks, done.

JimmyMow:
👍

PR #99, opened 2017-12-12
Drop package-lock.json in favor of yarn.lock

Empact:
package-lock.json is apparently defunct, was last updated in August.

It's problematic to support both systems because of divergence / drift
between the two lock files.


Empact:
Closing since c7ed8ece98198bdb76846a37f57cfd7dabd60693 is in

PR #100, opened 2017-12-12
Improve propTypes for Peer and Peers through use of shape

Empact:


JimmyMow:
Nice. Should probably do this throughout the application at some point. Also makes it easier to read code as data gets passed down through components

Issue #95, opened 2017-12-11
Integrate Travis


JimmyMow:
https://github.com/LN-Zap/zap-desktop/pull/85#issuecomment-350393176

@torkelrogstad here is the issue to set up continuous integration for the project

JimmyMow:
@torkelrogstad

torkelrogstad:
Nice
@JimmyMow I see that there's already a `.travis.yml` file in the repo. That's the file Travis uses for config purposes. Was Travis set up for this project previously?

JimmyMow:
I didn't personally, I assume it came with the boilerplate I used @torkelrogstad

auldsyababua:
I can try to take a stab at this next week if wanted. How do I get in on the slack channel?

JimmyMow:
@auldsyababua slack link should be in the README

NahomAgidew:
@JimmyMow Since there's already the config file, you just go to https://travis-ci.org/ and enable Travis for the project.

Empact:
First ever CI-checked PR 🤘 https://github.com/LN-Zap/zap-desktop/pull/140

JimmyMow:
Ayyyyy. Just don't wanna close this cause it took Travis 20 minutes to clear lol. Must be doing something wrong

Empact:
@JimmyMow propose we close this and improve the CI over time

JimmyMow:
@Empact agree. closing

PR #113, opened 2017-12-19
Fix lints so that test-all runs green

Empact:
With this we can set up CI on test-all for PR review, etc.

### DIFF
--- a/app/components/Network/CanvasNetworkGraph.js
+++ b/app/components/Network/CanvasNetworkGraph.js
@@ -64,9 +64,14 @@ class CanvasNetworkGraph extends Component {
 
     const simulationDataEmpty = !nodes.length && !links.length
     const networkDataLoaded = network.nodes.length || network.edges.length
+    const prevNetwork = this.props.network
 
-    // if the simulationData is empty and we have network data
-    if (simulationDataEmpty && networkDataLoaded) {
+    if (
+      // update the simulationData only if
+      // the simulationData is empty and we have network data
+      (simulationDataEmpty && networkDataLoaded) ||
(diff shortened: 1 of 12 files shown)


Empact:
#95

app/components/Network/CanvasNetworkGraph.js
```diff
@@ -112,7 +108,7 @@ class CanvasNetworkGraph extends Component {
 
     // remove active class
     d3.selectAll('.active-peer')
-      .each(function () {
```
JimmyMow: es6 arrow functions lexically bind `this`, so i did not use them here on purpose. may need to disable these lines for lint

JimmyMow:
wow this is awesome @Empact, you are killing it. Did you test the functionality locally? I might pull this branch down just to poke around and make sure things work

Empact:
Thanks! I did test a bit, not exhaustively. I actually just synced the testnet locally so I haven't been able to run transactions or anything through it.

I took a different approach on the d3 selectAll situation - according to the docs you can call `classed` etc on the selection itself, which is more succinct/declarative and should be more performant than iterating and re-selecting.

LMK if you find anything broken, would be good to test whatever that is.

JimmyMow:
Thanks @Empact. Been a bit busy, my apologies, I'm gonna pull this down after lunch and poke around. Will report back afterwards

Empact:
Green on jenkins as well FYI https://travis-ci.org/Empact/zap-desktop/builds/320447675

JimmyMow:
niiiiice @Empact! cloned it down and it's working for me locally. merging 🎉

PR #121, opened 2017-12-23
Update electron and electron-builder to latest release.

joaodealmeida:
This PR updates electron and electron-builders packages to the latest stable release.

Uses yarn.lock instead of package-lock adopted since  c7ed8ec commit.

(couldn't use the old PR since I've deleted the old branch, sorry for the spam 🙏)


Empact:
You'll need to update the `install-grpc` action in app/package.json - it has:

`--target=1.6.10`

PR #124, opened 2017-12-24
Fix that lnd might not be detected on Mac OS due to misaligned `ps` output

Empact:
On mac, the ps output can be misaligned which prevents lnd from being detected, e.g. when some VSZ field is longer than expected.

The default ps-node lookup is via ps lx, which includes:

```
  -l Display information associated with the following keywords:
    uid, pid, ppid, flags, cpu, pri, nice, vsz=SZ, rss, wchan,
    state=S, paddr=ADDR, tty, time, and command=CMD.
```

Here's some example misaligned output from `ps lx`

```
  501 19344     1   0   4  0  4376668   1804 -      S      ??    0:01.23 /System/Library/PrivateFrameworks/IMDPersistence.framework/IMAutomaticHistoryDeletionAgent.a
  501 33800     1   0   4  0 21481005020 172224 -      S      ??    1:35.79 /Applications/GitX.app/Contents/MacOS/GitX
  501 34520     1   0   4  0  4453292   3568 -      S      ??    0:04.03 /System/Library/CoreServices/navd
  501 34689   352   0  31  0  5668268  43616 -      S      ??    2:01.09 /Applications/Slack.app/Contents/Frameworks/Slack Helper.app/Contents/MacOS/Slack Helper --t
```

Many of these fields are unnecessary because we're looking up by name only,
so we can lookup using only the `-o pid,command` output.

https://github.com/neekey/table-parser/issues/11
https://github.com/neekey/ps/pull/64

### DIFF
--- a/app/lib/lnd/util.js
+++ b/app/lib/lnd/util.js
@@ -187,7 +187,7 @@ export const createMacaroonCreds = async macaroonPath => {
 export const isLndRunning = () => {
   return new Promise((resolve, reject) => {
     mainLog.info('Looking for existing lnd process')
-    lookup({ command: 'lnd' }, (err, results) => {
+    lookup({ command: 'lnd', psargs: 'x -o pid,command' }, (err, results) => {
       // There was an error checking for the LND process.
       if (err) {
         return reject(err)


Empact:
This might break ubuntu, according to this comment: https://github.com/neekey/ps/blob/master/lib/index.js#L119

Empact:
Updated to include specific output columns. This is tested to work on my machine.

JimmyMow:
@Empact we should test this on other machines just in case, or wait to see how https://github.com/neekey/ps/pull/69 gets handled

Empact:
Rebased, this is still an issue fyi. Will ping neekey again.

mrfelton:
We no longer attempt to detect wether or not lnd is already running on the user's machine as there is no longer a hard requirement that our internal instance of lnd must be the only one running.

PR #129, opened 2017-12-27
Allow submitting peer on Enter key press

pajasevi:
This should enhance user experience for a bit when adding new peers. No side-effects.


JimmyMow:
Nice

Issue #138, opened 2017-12-28
UX Enhancement: Remove error popup message after an action


thinkjanis:
The top red popup doesn't disappear after an error. When modifying input fields (performing an action) or switching to a new view, it should disappear.

pajasevi:
What about timeout on error message? 10 seconds maybe? Could be easy to implement.

thinkjanis:
Yes, good point @pajasevi. I was about to add it but the timeout may prevent the user from reading it, gets distracted etc. What if a user sends a transaction, the error is displayed, and the app is minimized? 

I am thinking the timeout should apply when the app window is active. Every time the app is minimized with the message, the timeout stops, and when maximized, resets or has 5 seconds remaining.

pajasevi:
Maybe we should rethink how the global error UI works. Right now it can only show one error, so more errors happening at the same time are a problem.

Solution would be adding errors to the list and displaying them to the user in more compact but still visible way (top of the UI, not fixed to the top of the app). This would allow displaying more errors without hurting the view and there would be no need for timeout.

I can take a look at.

Enhancement: propagate errors to system notifications maybe?

JimmyMow:
Gonna close this issue since #140 is merged. 

@pajasevi if you wanna open a new ticket to enhance global errors feel free. Probably a good idea

pajasevi:
Will do! Thanks :zap:

Issue #136, opened 2017-12-28
UI Bug: Make error message close button (X) noticeable with long messages


thinkjanis:
![screenshot from 2017-12-26 20 47 18](https://user-images.githubusercontent.com/31632325/34417889-9cd89594-ec03-11e7-9211-23e6f26235c5.png)

NahomAgidew:
@thinkjanis  I was thinking of removing the close button and make the error message auto-disappear after a few seconds.

thinkjanis:
Solved by #140

PR #140, opened 2017-12-28
Error message enhancements

pajasevi:
Solves #138 and #136 

`clearError` is automatically called after 10 seconds.
Styles were adjusted so that close icon is more visible and error message does not overflow over it.


JimmyMow:
she's a beauty, that PR #140. Nice 👍

PR #147, opened 2017-12-30
Show loading while connecting to a peer

pajasevi:
User doesn't know if anything is actually happening while trying to connect to a peer. It is better to display a loading icon and overlay over the form.

peer-loading


JimmyMow:
Going to close this since we no longer have a peers form anymore :)

Issue #145, opened 2017-12-29
Have a unified close button (X) style in all modals


thinkjanis:
![close_btn](https://user-images.githubusercontent.com/31632325/34440390-800e5ba0-ecbd-11e7-85c4-15fa1613d159.png)

JimmyMow:
You're on fire @thinkjanis, nice. Thank you

pajasevi:
I'm on it! :wink:

PR #148, opened 2017-12-30
Unified close icons

pajasevi:
I changed all close icons to FaClose (look way better than MdClose) except for GlobalError component which will be rewritten anyway.

I also added close icon to PeerForm component.

Second commit is just code cleaning.

Solves #145

### DIFF
--- a/app/components/ChannelForm/ChannelForm.js
+++ b/app/components/ChannelForm/ChannelForm.js
@@ -51,7 +51,7 @@ const ChannelForm = ({
       className={styles.modal}
     >
       <div onClick={closeChannelForm} className={styles.modalClose}>
-        <FaClose className={styles.close} />
+        <FaClose />
       </div>
 
       <header className={styles.header}>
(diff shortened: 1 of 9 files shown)


app/components/ChannelForm/ChannelForm.js
```diff
@@ -51,7 +51,7 @@ const ChannelForm = ({
       className={styles.modal}
     >
       <div onClick={closeChannelForm} className={styles.modalClose}>
-        <FaClose className={styles.close} />
```
Empact: Looks like styles.close is unused after this removal. You can drop it from the scss as well.

pajasevi: Right! Thanks for the tip, Webstorm didn't show me that.

pajasevi: It's not in the SCSS at all, that's why I removed it https://github.com/pajasevi/zap-desktop/blob/95872d8ea3ed76f50ece53173e0871f2f83269b6/app/components/ChannelForm/ChannelForm.scss

JimmyMow:
Nice @pajasevi, looks good to me. Merging

PR #153, opened 2018-01-02
Bring CI runtime down from ~19 to ~13 minutes

Empact:
This parallelizes the lint and test efforts, and caches more of the setup results, most notably apt.

Note this also changes the tests to cover Node 8 & 9 rather than 7 & 8, and relies on the default build environment on Travis rather than setting up g++ 4.8 from the ubuntu dev sources.

PR #154, opened 2018-01-04
Bring CI runtime down from ~19 to ~13 minutes

Empact:
This parallelizes the lint and test efforts, and caches more of the setup results, most notably apt.

Note this also changes the tests to cover Node 8 & 9 rather than 7 & 8, and relies on the default build environment on Travis rather than setting up g++ 4.8 from the ubuntu dev sources.

Previously open as #153


Empact:
I tried some of the other arrangements and this one seems best. Runtimes vary due to, e.g. variance in network latency, but it's on average a significant improvement.

Empact:
Here's another example build for the same code:
https://travis-ci.org/Empact/zap-desktop/builds/324852092

JimmyMow:
@Empact Awesome

PR #155, opened 2018-01-04
Drop node 7 support, declare minimum versions in package.json

Empact:
Node 7 was a dev release and is defunct now that 8 is LTS and 9 is the
current dev release. If we wanted to support earlier versions, we should
go all the way back to the prior LTS release series, 6, but I don't see
a reason to do that for new software.

Bump to npm 5 because that is the version that was released with node 8:
https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#8.0.0

This was prompted by the fact that current versions of yarn have dropped node 7
support.
https://yarnpkg.com/en/

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,7 @@
   "presets": [
     ["env", {
       "targets": {
-        "node": 7,
+        "node": 8,
         "browsers": "electron 1.7"
       },
       "useBuiltIns": true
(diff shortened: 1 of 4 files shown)


JimmyMow:
@Empact hmmmm I tried to resolve the conflicts via Github and it's not happy for some reason. Wanna fix those real quick and then I'll merge?

(Only reason there was merge conflicts is because your previous PR also got merged)

PR #156, opened 2018-01-04
Fix browserslist to provide the proper dependencies for electron 1.7

Empact:
Was not installed and was configured for electron 1.6.

With this fix, babel builds for electron 1.7 / chrome 58, rather than just node 7.


JimmyMow:
@Empact 👏

PR #158, opened 2018-01-05
Improve Zap installation instructions

thinkjanis:
Improve Zap installation instructions


thinkjanis:
Should I mention that Zap uses .lnd for config and that it should be deleted if the user encounters any problems?

JimmyMow:
@thinkjanis nice

PR #161, opened 2018-01-05
Extract blockExplorer util

Empact:
To handle shelling out to the transaction display

This was previously done inline via electron.shell.openExternal. Consolidating
it in the utility DRYs up the code both in terms of the base url and the path
building.

Note there is also code in reducers/info.js relating to:
infoSelectors.explorerLinkBase, but it seems to be unused currently.

### DIFF
--- a/app/components/Contacts/ClosingContact.js
+++ b/app/components/Contacts/ClosingContact.js
@@ -1,8 +1,7 @@
-import { shell } from 'electron'
 import React from 'react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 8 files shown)


JimmyMow:
Concept ACK

Looks awesome dude. 

`infoSelectors.explorerLinkBase` should be used, I've just been lazy. But when we have mainnet and testnet we'll want dynamic base for the links. 

Looks like the lint failed because of `Node 7` dependencies. Didn't we not update travis when we removed the `Node 7` dependency

PR #163, opened 2018-01-07
enhance(peer-form): auto fill host if pasted pubkey contains host inf…

stues:
PeerForm: Auto fill host if pasted pubkey contains host information

If pasted pubkey is in the form of \<pubkey\>@\<host\> the string will be splitted and the pubkey and host will be set accordingly.
This simplifies adding new peers since most pages use the format \<pubkey\>@\<host\> to publish their lightning node information.


JimmyMow:
We recently merged in a big change that removes the peer form. Gonna close this

PR #164, opened 2018-01-08
Fix that we had redundant lnd rpc.proto files

Empact:
Only one of these files was being used, and they were going out of step
relative to one another.

Also updated the def from:
https://github.com/lightningnetwork/lnd/blob/master/lnrpc/rpc.proto


JimmyMow:
I don't think this will work when we build. The reason there are 2 is because I was lazy and didn't set the correct path for build settings. 

What should be changed is this path: https://github.com/LN-Zap/zap-desktop/blob/master/package.json#L50

Empact:
Hmm... I tried building the app and running it but that fails with a security exception. Any tips?

screen shot 2018-01-12 at 02 28 00

Empact:
I'm working on getting code signing up to test the electron-builder output. Should hopefully have that together soon.

JimmyMow:
@Empact Wow dude you are the man. I was going to work on that super soon as like the cherry on top before MVP release, would be awesome if you take a peak over the weekend. Keep me posted!

mrfelton:
@Empact this needs to be updated to apply cleanly to master due to a number of recent changes around how we manage the lnd client configuration.

Empact:
Solved with #406. Thanks @mrfelton!

PR #168, opened 2018-01-09
Change "block chain" to "blockchain"

JukeboxRhino:
Historically, "block chain" was the term used to refer to Bitcoin's distributed ledger. However, usage turned away from that version to the combined "blockchain". In all news articles, forum posts, etc., people now use the single-word version.

[See the first sentence of this Wikipedia article](https://en.wikipedia.org/wiki/Blockchain#cite_note-primer-4)

### DIFF
--- a/app/components/LndSyncing/LndSyncing.js
+++ b/app/components/LndSyncing/LndSyncing.js
@@ -9,11 +9,11 @@ class LndSyncing extends Component {
       facts: [
         {
           title: 'The Lightning Network',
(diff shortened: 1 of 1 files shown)


Empact:
I'm negative on this, not because one or the other is better, but because it's not worth bothering to make a change over such a thing. "block chain" and "blockchain" both clearly communicate what they're intended to communicate.

JukeboxRhino:
@Empact That's a fair assessment. I've also just noticed that there's a new sync screen branch that would make this change obsolete.

It's nothing huge, just something that popped out at me when I started the client for the first time.

JimmyMow:
Cool. I'm gonna close this then. Hopefully we get the new syncing page in there sooner than later

PR #175, opened 2018-01-10
Move PayForm amount error message below the field

### LINKED ISSUE #174: Error text for currency in PayForm is off
The error text for the currency is off, it is ment to be to the right of the currency icon but that only works when there are many numbers entered, else the text overlaps the icon.
skarmavbild 2018-01-10 kl 13 35 45

A fix is to move it to the bottom as the other error message, which would result in the following:
skarmavbild 2018-01-10 kl 13 57 26

And with the suggested fix in #143 it would look like this:
skarmavbild 2018-01-10 kl 13 53 41

Empact:
This is consistent with the other field and avoids overlap formatting issues.

Fixes #174

### DIFF
--- a/app/components/Form/PayForm.js
+++ b/app/components/Form/PayForm.js
@@ -49,11 +49,6 @@ class PayForm extends Component {
         {showPayLoadingScreen && <LoadingBolt />}
 
         <section className={`${styles.amountContainer} ${isLn ? styles.ln : ''} ${showErrors.amount && styles.error}`}>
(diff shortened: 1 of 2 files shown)


Empact:
screen shot 2018-01-10 at 11 45 25

Empact:
screen shot 2018-01-10 at 11 49 46

Empact:
screen shot 2018-01-10 at 12 50 52

JimmyMow:
I like this change. Also I love that you include screenshots in your UI PRs, awesome.

PR #176, opened 2018-01-10
Size the PayForm amount input to hold at least 2 chars

Empact:
Because the "." of "0." is not propagated via the JS, so not in the size of
the amount string.


Empact:
Before:
screen shot 2018-01-10 at 11 50 55

After:
screen shot 2018-01-10 at 11 51 29

Issue #177, opened 2018-01-10
By default, show the peers in the Add Contact list


Empact:
Currently this space is empty, and showing the peers gives the user an easy way to open a channel with a random active/online node.

JimmyMow:
fixed in #184

PR #184, opened 2018-01-12
Fix/contacts form

JimmyMow:


JimmyMow:
fixes #177, UX feedback when submitting a manual form and search still works even if user includes the @host along with the pubkey when adding a contact

PR #188, opened 2018-01-13
Enable our eslint plugins, which are currently inactive

Empact:
This enables the following extensions:

    "plugin:react/recommended",
    "plugin:flowtype/recommended",
    "plugin:import/errors",
    "plugin:import/warnings",
    "plugin:jsx-a11y/strict",
    "plugin:promise/recommended"

And rules:

    "compat/compat": "error",
    "flowtype-errors/show-errors": "error",
    "flowtype-errors/enforce-min-coverage": ["error", 20]

### DIFF
--- a/app/lnd/methods/index.js
+++ b/app/lnd/methods/index.js
@@ -25,6 +25,7 @@ export default function (lnd, meta, event, msg, data) {
         .then((infoData) => {
           event.sender.send('receiveInfo', infoData)
           event.sender.send('receiveCryptocurrency', infoData.chains[0])
+          return infoData
         })
         .catch(() => event.sender.send('infoFailed'))
       break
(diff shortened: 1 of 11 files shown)


JimmyMow:
Concept ACK

Will review and test the code in a bit 👍

app/lnd/methods/index.js
```diff
@@ -42,13 +43,15 @@ export default function (lnd, meta, event, msg, data) {
...
-          networkController.queryRoutes(lnd, meta, { pubkey: invoiceData.destination, amount: invoiceData.num_satoshis })
-            .then((routes) => {
-              event.sender.send('receiveInvoiceAndQueryRoutes', routes)
-            })
-            .catch(error => console.log('getInvoiceAndQueryRoutes queryRoutes error: ', error))
-        })
+        .then(invoiceData =>
+          networkController.queryRoutes(lnd, meta, {
+            pubkey: invoiceData.destination,
+            amount: invoiceData.num_satoshis
```
Empact: Incidentally I think we should adopt a practice of explicitly labelling variables by their units, ala `_satoshis`.

JimmyMow: yeah we need to come up with some unit standard going forward, especially because we want to allow unit switching within the app

Issue #193, opened 2018-01-16
UI Bug: Dropdown menus visible over modal overlay


dfattlar:
Steps to reproduce:

1. Open Wallet or Contact Tab
2. Open Filter Dropdown
3. Activate Modal (Contact -> Add Contact, Wallet -> Receive Payment)
4. Observe the Filter Dropdown elements are still visible and clickable over the modal overlay and sometimes over the modal itself.

<img width="486" alt="screen shot 2018-01-15 at 4 17 56 pm" src="https://user-images.githubusercontent.com/4843270/34966686-4f791a54-fa12-11e7-8348-a249ff0185fe.png">
<img width="594" alt="screen shot 2018-01-15 at 4 24 17 pm" src="https://user-images.githubusercontent.com/4843270/34966691-51d758e2-fa12-11e7-85dc-3e813f5d170d.png">
<img width="540" alt="screen shot 2018-01-15 at 4 31 04 pm" src="https://user-images.githubusercontent.com/4843270/34966695-555b1ac6-fa12-11e7-80a2-055792b60ba2.png">

JimmyMow:
#194 

closing

PR #194, opened 2018-01-16
fix(filter-over-modal): Add z-index to open ReactModal Overlay

dfattlar:
This is related to Issue #193.

I was looking at a better way to track z-indexes as was mentioned in [this comment](https://github.com/LN-Zap/zap-desktop/pull/183#issuecomment-357128547), and came across the bug.

After checking the react-modal issues, their accepted fix was manually adding a z-index to the modal overlay class, see [this react issue](https://github.com/reactjs/react-modal/issues/390). 

Increasing the z-index keeps the modal at the forefront and blocks clicking on the filter dropdown until the modal is closed. I added the update to app.global.scss so that the z-index would be applied to the modal across the app. Doing so seems to fix the issue in all locations where <ReactModal> is used.

Separately, I found some better ways of documenting/tracking the z-index usage via variables. Will hopefully have a PR for that later this week.

### DIFF
--- a/app/app.global.scss
+++ b/app/app.global.scss
@@ -192,5 +192,9 @@ body {
   stroke: #88D4A2;
   stroke-width: 15;
   stroke-dasharray: 100;
-  animation: dash 2.5s infinite linear;  
+  animation: dash 2.5s infinite linear;
+}
+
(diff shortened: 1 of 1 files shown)


JimmyMow:
Nice @dfattlar! Excited for that future PR :)

This looks good, merging

PR #198, opened 2018-01-17
Feature(z-index-vars): Use function to apply z-index so they can easi…

dfattlar:
…ly be tracked as vars.

Following up on comments from #183, this update uses a function to apply the z-index in each .scss file. The function needs to iterate over a map of z-index values, so all of the z-indexes end up in a single list and can be tracked more easily. 

There are a couple ways of doing this, [a few discussed here](https://www.sitepoint.com/better-solution-managing-z-index-sass/). This approach seemed the best since it allows for scaling with nesting values - where the top level is the component name.

Possible improvements:
- Using _zindex.scss partial to track z-index so that the variables.scss file doesn't get overcrowded. A drawback to this would be having to import the partial into all files that use z-index.
- Use a more descriptive name for the z function? I saw the it used that way a few places and just decided to keep it.

### DIFF
--- a/app/app.global.scss
+++ b/app/app.global.scss
@@ -196,5 +196,5 @@ body {
 }
 
 .ReactModal__Overlay.ReactModal__Overlay--after-open {
-  z-index: 50;
+  z-index: z("react-modal","open");
 }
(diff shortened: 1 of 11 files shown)


mrfelton:
This was integrated via #831

PR #208, opened 2018-01-20
Fix: adds error message to contact form invalid manual form input 

### LINKED ISSUE #191: UI: Manual Add Contact Submit Button doesn't respond if the format is incorrect
As a first time user this was a bit confusing to me, and I assumed my connection was broken because there was no feedback when I was trying to add a NodeID without an "@ip:port."  I'd like to add a simple error message under the form input.

waseem999:
This PR adds a red error message to incorrectly formatted inputs to the manual form field.  The fix is modeled after the payform input error.  It involved a little refactoring by moving the manual form search query from local state to the redux store.  

This resolves #191


app/components/Contacts/ContactsForm.js
```diff
@@ -10,29 +10,30 @@ class ContactsForm extends React.Component {
...
-      editing: false,
-      manualFormInput: ''
+      editing: false
     }
   }
 
   render() {
     const {
       contactsform,
+      contactsform: { showErrors },
```
JimmyMow: since we are declaring `contactsform` in the line above maybe this is unnecessary?

below we can just do `contactsform.showErrors`

JimmyMow:
i'm looking to post some builds very soon, so I'm going to merge. this is great, I can update some clean up on a separate PR if need be. 

thanks @waseem999!

PR #213, opened 2018-01-24
fix(electron security): Bump electron to 1.6.16

Empact:
To fix protocol handler vulnerability
https://electronjs.org/blog/protocol-handler-fix

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -10,8 +10,8 @@
     "url": "https://github.com/LN-Zap/zap-desktop"
   },
   "scripts": {
-    "postinstall": "npm rebuild --runtime=electron --target=1.6.10 --disturl=https://atom.io/download/atom-shell --build-from-source",
-    "install-grpc": "cd node_modules/grpc && git submodule update --init && npm run electron-build -- --target=1.6.10"
+    "postinstall": "npm rebuild --runtime=electron --target=1.6.16 --disturl=https://atom.io/download/atom-shell --build-from-source",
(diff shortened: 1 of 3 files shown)


JimmyMow:
@Empact why not update the package.json?

Empact:
Yeah, on that - I'm also switching to 1.6.16 in light of your latest change.

app/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "scripts": {
     "postinstall": "npm rebuild --runtime=electron --target=1.6.6 --disturl=https://atom.io/download/atom-shell --build-from-source",
```
JimmyMow: @Empact also thoughts on updating `--target` to the electron version we are packaging?

Empact: Yeah fixed in the current code.

Empact:
@JimmyMow thoughts on moving back to the electron 1.7 series? What build command was failing here: 66c97eb3c41191d0980591ab170460019787214a

JimmyMow:
@Empact we can move back to the `1.7` series. It was actually a `grpc` issue. `grpc@1.7.3` combined with electron was getting `404` on packaging for linux and was packaging incorrectly.

switching to `grpc@1.7.1` fixed this, I think we can go back to `1.7` series

Empact:
Cool will do in a separate PR

JimmyMow:
Yeah we can certainly improve our automated testing. Catching something like this would be tricky but an overall improvement is probably a priority todo at some point

PR #225, opened 2018-01-25
Feature/lnd error messages

Vindexus:
These are meant to make it easier for new devs to set up the project.


Empact:
How about a user-visible (rather than logged-only) failure.
`dialog.showMessageBox` should serve.

PR #280, opened 2018-02-04
fix(lndsync): syncPercentage check and top margin on pager

lepipele:
Just few small fixes now that I can compile project on my machine. Here is how screen looked before:

before

And then after these small fixes:

after

`syncPercentage` obviously still updates to `0%` as soon as property is updated to number (no longer isNan).


lepipele:
I've used code from @Vindexus pull request: #225 to help with debugging... and it's included in pull request.

JimmyMow:
Have you tested this on multiple screen sizes? Not just the default?

Also, travis failed here. Would you mind getting this to pass? Thanks!

lepipele:
Yes. The thing is - even initial version works OK as soon as you resize - because it has to do with text height in 1st tab.

I've fixed lint problems... should be good to merge now. 

Regarding that - any way we can speed up these Travis builds? They are taking forever.

lepipele:
OK, those tests that are failing have to do with the fact that build server doesn't have `lnd`. So this new behavior of providing information in console that application is failing to start because `lnd` is missing would require to either:

1. Update test not to fail because more information in log is better
2. Code update that instead of showing this in console would show it in main window

JimmyMow:
Closing since there is not a new sync screen

PR #296, opened 2018-02-10
Fix/activity header selection and hitbox area

DataCourier:
Fixes:
- activity header selection (disabled with css)
- activity header hitbox size (what used to be margin is now clickable, graphics unchanged)


JimmyMow:
Nice 👍

Issue #326, opened 2018-03-08
Users should be able to easily connect to remote lnd server


mrfelton:
Currently, there are a number of hardcoded assumptions that the user is running an lnd node on the localhost.

Users should be able to easily override this to connect to an alternate lnd server, such as one that is running on a remote host.

Issue #324, opened 2018-03-08
Handshake failed with fatal error SSL_ERROR_SSL: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE.


Empact:
This is with the current master of Lnd, as run by Zap.

```
CERT EXISTS, STARTING GRPC
[0] E0307 18:10:13.136523000 140735655744320 ssl_transport_security.c:970] Handshake failed with fatal error SSL_ERROR_SSL: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE.
[0] error:  { Error: Connect Failed
[0]     at ClientReadableStream._emitStatusIfDone (/Users/ben/Development/Bitcoin/zap-desktop/app/node_modules/grpc/src/client.js:255:19)
[0]     at ClientReadableStream._readsDone (/Users/ben/Development/Bitcoin/zap-desktop/app/node_modules/grpc/src/client.js:221:8)
[0]     at readCallback (/Users/ben/Development/Bitcoin/zap-desktop/app/node_modules/grpc/src/client.js:283:12) code: 14, metadata: Metadata { _internal_repr: {} } }
[0] status:  { code: 14,
[0]   details: 'Connect Failed',
[0]   metadata: Metadata { _internal_repr: {} } }
[0] error:  { Error: Connect Failed
[0]     at ClientReadableStream._emitStatusIfDone (/Users/ben/Development/Bitcoin/zap-desktop/app/node_modules/grpc/src/client.js:255:19)
[0]     at ClientReadableStream._readsDone (/Users/ben/Development/Bitcoin/zap-desktop/app/node_modules/grpc/src/client.js:221:8)
[0]     at readCallback (/Users/ben/Development/Bitcoin/zap-desktop/app/node_modules/grpc/src/client.js:283:12) code: 14, metadata: Metadata { _internal_repr: {} } }
[0] TRANSACTION STATUS:  { code: 14,
[0]   details: 'Connect Failed',
[0]   metadata: Metadata { _internal_repr: {} } }
[0] 2018-03-07 18:10:13.161 [DBG] BTCN: waiting to see if more peers will respond with cfheaders
```

Empact:
Here's a temp workaround from @jackmallers:
```
process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'
[...]

Issue #167, opened 2018-01-09
Allow custom lnd configuration


JukeboxRhino:
Allowing for custom configuration would allow users to

- Listen on custom ports for p2p connections, rpc and the REST server.
- Increase the maximum number of pending incoming channels per peer
- Enable/disable autopilot to allow users to run Zap without having it open channels for them, potentially locking up funds they would prefer remain off the lightning network

There are quite a few other configuration options, but these are the ones I feel are most important for users. If this were to be implemented into the GUI, these are the only options I would let users change.

JimmyMow:
Nice, I think this would be cool. Gonna assign this to myself

sneurlax:
You can use Zap for Litecoin's Lightning Network by changing some parameters...  If you consider extending generic support, I know that Decred and Vertcoin are both compatible (they all ultimately use `btcd`.)

JimmyMow:
@sneurlax other currencies would need to add LND support. 

And then ultimately if they would like to have Zap production builds with all dependencies packaged in they would need neutrino support as well, which even LTC doesn't have.

I am down to support any coin though

mrfelton:
We do now support custom lnd connections as well as configuration of key settings such as autopilot and node alias. We may look to add additional configuration options in the future.

I'm going to close this issue but If there are any further settings that you think should be added please open a new issue for each one individually so that we can discuss the specifics. Thanks.

PR #327, opened 2018-03-08
Support custom lnd connections

mrfelton:
This PR exposes the lnd connection params as environment variables that make it simple to run Zap with an external lnd node.

I've implemented dotenv (https://github.com/motdotla/dotenv) to enable the settings to be controlled by a `.env` file placed at the project root.

Addresses #326 and #324


Empact:
Also #167

app/main.dev.js
```diff
@@ -257,12 +253,14 @@ app.on('ready', async () => {
 
   sendGrpcDisconnected()
   // Check to see if an LND process is running
-  lookup({ command: 'lnd' }, (err, results) => {
+  lookup({ command: 'node' }, (err, results) => {
```
Empact: Why `node` here? The executable is `lnd`, `node` seems overly broad.

mrfelton: Good catch @Empact - that was left over from when I was testing some thing locally, before I introduced the `ZAP_LND_DISABLE` flag. Have removed and updated the PR

mrfelton:
Superseded by https://github.com/LN-Zap/zap-desktop/pull/351 so closing this PR.

Issue #254, opened 2018-01-28
Error on startup "JavaScript error occurred in the main process"


coolpick:
Started up the first time without a problem.  I attempted to let Chrome open a lightning address with Zap, but it opened a second instance.  This error showed.  When I re-launched both to re-test, I started getting this error.  After dismissing error, lightning fails to connect to anything, is indefinitely loading.

Before the error:
- Attempted to open lightning url via chrome
- Second instance was opened
- Manually created peer, double clicked and two were showing as pending

Issue could be related to any of the above, or completely random.  Have not been able to find a log file anywhere.

About 10 minutes later, the error has gone away.  I'm leaning toward it being related to the port or a windows process.  A task hanging.

![zap](https://user-images.githubusercontent.com/1341176/35485353-673db356-042c-11e8-835b-1848243b3ac7.png)

Ceratodontiformes:
Same error here. Also Win (7).
- Trying to set up channel.
- Zap becomes "slow"
- Close Zap
- Open Zap -> Error
- open/close/open/close -> always error
- reboot -> still error

- after several minutes -> reopened -> error gone.

// thx for the awsome project

cavapoo2:
Similar error: 
```
Uncaught Exception:
Error: Connect Failed
    at ClientReadableStream._emitStatusIfDone (/home/febe/lightning/zap/zap-desktop/app/node_modules/grpc/src/client.js:255:19)
    at ClientReadableStream._readsDone (/home/febe/lightning/zap/zap-desktop/app/node_modules/grpc/src/client.js:221:8)
    at readCallback (/home/febe/lightning/zap/zap-desktop/app/node_modules/grpc/src/client.js:283:12)
```

I running simnet (btcd) and 2 lnd nodes (lnd version 0.3.0-alpha). I think i need to learn lightning code in more detail before playing around with this.

mrfelton:
Closing as a duplicate of https://github.com/LN-Zap/zap-desktop/issues/234

Issue #247, opened 2018-01-28
When I close Zap and start it again, I get following Javascript error


larspohlmann:
Uncaught Exception:
Error: Connect Failed
    at ClientReadableStream._emitStatusIfDone (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:255:19)
    at ClientReadableStream._readsDone (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:221:8)
    at readCallback (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:283:12)

--

MacOs 10.13.1

A reboot resolves the issue

PlanBit:
Same. Than restarted the pc and everything was fine.

Pearse500:
Same and good to know a restart gets it back. was uninstalling and reinstalling to get back to working.

mrfelton:
Can someone confirm if this issue is still reproducible with the latest code in the master branch?

mrfelton:
Closing this story as a duplicate of https://github.com/LN-Zap/zap-desktop/issues/234

Issue #236, opened 2018-01-27
Error: A JavaScript error occurred in the main process


TomasHo:
Uncaught Exception:
Error: Connect Failed
    at ClientReadableStream._emitStatusIfDone (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:255:19)
    at ClientReadableStream._readsDone (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:221:8)
    at readCallback (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:283:12)

TomasHo:
Happy to provide more info if needed.

1hiro:
I get this javascript error on startup.  

Uncaught Exception:
Error: Endpoint read failed
    at ClientReadableStream._emitStatusIfDone (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:255:19)
    at ClientReadableStream._receiveStatus (/Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:233:8)
    at /Applications/Zap.app/Contents/Resources/app.asar/node_modules/grpc/src/client.js:705:12

JimmyMow:
@1hiro @TomasHo this is a duplicate of #234 

let's take this discussion over there. I gave a potential hack-fix for the time being

Issue #207, opened 2018-01-19
"Endpoint read failed" Javascript Error when connection to LND is lost bug


helgabutters:
I think this error happens when you kill `lnd` while Zap is currently talking with `lnd`. During my testing with #205 and #206 I was killing `lnd` at different points to see the response and this error popped up on three random occasions.

### Error Log:

```
[0] status:  { code: 0,
[0]   details: '',
[0]   metadata: Metadata { _internal_repr: {} } }
[0] end
[0] error:  { Error: Endpoint read failed
[0]     at ClientReadableStream._emitStatusIfDone (/home/pnd/zap-desktop/app/node_modules/grpc/src/client.js:255:19)
[0]     at ClientReadableStream._receiveStatus (/home/pnd/zap-desktop/app/node_modules/grpc/src/client.js:233:8)
[0]     at /home/pnd/zap-desktop/app/node_modules/grpc/src/client.js:705:12 code: 14, metadata: Metadata { _internal_repr: {} } }
[0] TRANSACTION STATUS:  { code: 14,
[0]   details: 'Endpoint read failed',
[0]   metadata: Metadata { _internal_repr: {} } }
```

### Screenshot:
![uncaughtexception](https://user-images.githubusercontent.com/8001978/35138126-00756dc6-fcbb-11e7-8beb-a1653be2b6b7.png)

JimmyMow:
Yeah so this is unexpected behavior, if LND gets killed then Zap obviously won't work.

Curious what you'd expect to see UX wise? Right now we don't do anything, we just let LND die

helgabutters:
My thought at this point would be on a timeout/unsuccessful connection, present the red error message box at the top of the UI. Simple suggestion for now on wording would be "Backend connection failed. Retrying in X sec" where X would be whatever value is currently used (or not) for the retry.

mrfelton:
Handling for if/when Lnd crashes was updated in 0.4.x to provide feedback to the user, present the error message, and send the user back to the launchpad area.

PR #329, opened 2018-03-10
Bump to grpc 1.10.0

Empact:
The impetus is to improve error reporting for GRPC connection errors,
which was introduced in 1.8.4. Security and other fixes are included
as well.
https://github.com/grpc/grpc-node/releases/tag/v1.8.4

Here's an example of the new `emitStatusIfDone` error reporting:
screen shot 2018-03-10 at 1 27 28 pm

This error is commonly reported so additional info could be helpful: #236 #141 #234 #254 #207 #247 #324

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -15,7 +15,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "grpc": "1.7.1",
+    "grpc": "1.10.0",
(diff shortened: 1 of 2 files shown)


Empact:
The error reporting justification may be weak, but I'm always in favor of upgrading core dependencies.

PR #330, opened 2018-03-11
Sort the contacts form relative to peer's contactability

Empact:
Private peers can't be acted on, so show them below the contactable peers.


Empact:
Now sorts alphabetically as well, mainly to avoid random reordering of the list during actions:

screen shot 2018-03-12 at 2 36 35 am

Empact:
Now sorts across all nodes, rather than just the results passed in. :P

PR #332, opened 2018-03-12
Make LND errors visible to users via a message box

Empact:
This ensures the user can see the error, and that Zap will not be running if LND quits.


Empact:
Going to merge this as a precaution for error visibility, we can back it out if it's a pain.

PR #335, opened 2018-03-13
Configure defaultremotedelay to match c-lightning, etc.

Empact:
Absent this, attempts to open channels with c-lightning nodes fail with:
"to_self_delay X greater than 432"

https://github.com/lightningnetwork/lnd/pull/788
https://github.com/ElementsProject/lightning/issues/1110

### DIFF
--- a/app/main.dev.js
+++ b/app/main.dev.js
@@ -151,6 +151,7 @@ const startLnd = (alias, autopilot) => {
   const neutrinoArgs = [
     '--bitcoin.active',
     '--bitcoin.testnet',
+    '--bitcoin.defaultremotedelay=432',
     '--bitcoin.node=neutrino',
     '--neutrino.connect=188.166.148.62:18333',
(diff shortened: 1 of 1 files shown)


Empact:
This option was added back in December: https://github.com/lightningnetwork/lnd/commit/0815bd4eb0c72b6104963622c038928dc1bb8941

JimmyMow:
We should close this yeah? I think the implementations plan on being compatible.

If not I like this idea and would be interested in merging it

Empact:
Yeah the current behavior is to scale the delay based on channel size, which seems like safe / more reasonable behavior, even if it results in some incompatibility between nodes. 
https://github.com/lightningnetwork/lnd/commit/43d7dd3d99b7cd9b889e8baa02f2d47670bff379#diff-ec4b517f146700b3546a7af2224b6191R357

PR #338, opened 2018-03-15
Update list of supported cipher suites for compatibility with lnd 0.4.0

### LINKED ISSUE #324: Handshake failed with fatal error SSL_ERROR_SSL: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE.

mrfelton:
Update the list of supported cipher suites to use with openssl to
include those supported by lnd, which are defined as:

tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305

See https://github.com/lightningnetwork/lnd/blob/master/lnd.go#L80-L85

Fix #324

### DIFF
--- a/app/lnd/lib/lightning.js
+++ b/app/lnd/lib/lightning.js
@@ -3,6 +3,21 @@ import path from 'path'
 import grpc from 'grpc'
 import config from '../config'
 
+// Default is ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384
+// https://github.com/grpc/grpc/blob/master/doc/environment_variables.md
+//
+// Current LND cipher suites here:
+// https://github.com/lightningnetwork/lnd/blob/master/lnd.go#L80
+//
+// We order the suites by priority, based on the recommendations provided by SSL Labs here:
+// https://github.com/ssllabs/research/wiki/SSL-and-TLS-Deployment-Best-Practices#23-use-secure-cipher-suites
+process.env.GRPC_SSL_CIPHER_SUITES = process.env.GRPC_SSL_CIPHER_SUITES || [
+  'ECDHE-ECDSA-AES128-GCM-SHA256',
+  'ECDHE-ECDSA-AES256-GCM-SHA384',
+  'ECDHE-ECDSA-AES128-CBC-SHA256',
+  'ECDHE-ECDSA-CHACHA20-POLY1305'
+].join(':')
+
 const lightning = (rpcpath, host) => {
   const lndCert = fs.readFileSync(config.cert)
   const credentials = grpc.credentials.createSsl(lndCert)


Empact:
Thanks, looks good. Could you add a comment something like this one, for reference:
```
// Default is ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384
// https://github.com/grpc/grpc/blob/master/doc/environment_variables.md
// Current LND cipher suites here:
// https://github.com/lightningnetwork/lnd/blob/master/lnd.go#L80
```

And I'm unsure whether we should include the default suites, given they aren't accepted by LND. I guess they're harmless given they'll be excluded during negotiation. Thoughts? /cc @jackmallers

mrfelton:
@Empact  I have updated the code to include the comment. Regarding wether the defaults should be included or not - I did test without and it works ok with the most recent version of lnd. The reason I opted to include them however was that I figured that they may be needed in order to continue to support older version of lnd.

Empact:
Fair enough - another question came to mind: should we sort the suites in an order to indicate our preference for stronger suites? Until recently LND led with the 384 bit ciphers:
https://github.com/lightningnetwork/lnd/commit/f7eeea71e206a514ee649060e903f72c9d4a8c46

Empact:
FWIW, I don't see why we would maintain compatibility with < LND 0.4.0.

mrfelton:
Ok, so as we don't want to support older versions of lnd I have removed the default suites -which I think makes sense, although could make the upgrade process more involved for some users.

With regards to the ordering, I'm not sure TBH. The way it is now matches what they are doing in lnd.

Empact:
Fair enough on ordering, I don't have a strong sense of the relative strength so leaving as is is fine.

That brings to mind that we should defer to the existing environment variable, if set, so that users can override / set their own ordering if desired.

mrfelton:
Good call. Have updated to honour existing values of the env var.

JimmyMow:
@mrfelton nice!

I agree that I don't see the need to support LND < 0.4. We certainly won't be releasing anything with LND < 0.4, I think having 0.4+ support only makes sense.

Looks good to me though, just wondering if `ECDHE-ECDSA-CHACHA20-POLY1305` is needed? I can test locally in a bit. Can anyone confirm it's necessary?

JimmyMow:
yeah looks like lightning-app doesn't support `ECDHE-ECDSA-CHACHA20-POLY1305` as well: https://github.com/lightninglabs/lightning-app/blob/v2/public/grpc-client.js#L9

mrfelton:
@JimmyMow any idea why `ECDHE-ECDSA-CHACHA20-POLY1305` is included in lnd?

Empact:
Roasbeef's answer from the LND Slack:

```
¯\_(ツ)_/¯

doesn't really matter
just needs to have a non empty intersection
```

https://lightningcommunity.slack.com/archives/C6BDA6DGE/p1521492965000083

mrfelton:
Ok, then IMO we should keep our list of supported suites inline with those supported by lnd.

Empact:
SSL Labs likes them in this order:

* ECDHE-ECDSA-AES128-GCM-SHA256
* ECDHE-ECDSA-AES256-GCM-SHA384
* ECDHE-ECDSA-AES128-CBC-SHA256
https://github.com/ssllabs/research/wiki/SSL-and-TLS-Deployment-Best-Practices#23-use-secure-cipher-suites

When pretty much agrees with Mozilla's list from 2016::

* ECDHE-ECDSA-AES128-GCM-SHA256
* ECDHE-ECDSA-AES256-GCM-SHA384
* ECDHE-ECDSA-CHACHA20-POLY1305
https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility

Based on the above, how about ordering them like so:

* ECDHE-ECDSA-AES128-GCM-SHA256
* ECDHE-ECDSA-AES256-GCM-SHA384
* ECDHE-ECDSA-AES128-CBC-SHA256
* ECDHE-ECDSA-CHACHA20-POLY1305

mrfelton:
Nice find @Empact - have updated the PR to reflect.

PR #340, opened 2018-03-16
Show the connected peers at the top of the Add Channel list by default

Empact:
These we have an active connection with so are most likely to result in successful channel opening

### DIFF
--- a/app/reducers/contactsform.js
+++ b/app/reducers/contactsform.js
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect'
 import filter from 'lodash/filter'
+import partition from 'lodash/partition'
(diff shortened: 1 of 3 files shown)


Empact:
This should be ideally be re-worked to not block on the peer lookup on startup, because it is not a strictly required addition of information. That said, if LND is available, the peer list should be available, and the data access improvement applies to the other lookups that aren't strictly necessary for initial display.

mrfelton:
needs a rebase

mrfelton:
I'm rebased and updated this PR in https://github.com/LN-Zap/zap-desktop/pull/1076. Closing as a duplicate

PR #342, opened 2018-03-16
Fix the transaction scrollbar to consistently reveal all the transactions

### LINKED ISSUE #311: Cannot scroll to the bottom of content #UI
For example, when you have 10+ transactions in the transaction list you can scroll only to a certain point and cannot see the last 1-3 transactions (depending on the viewport height). Scrollbars appear to not have reached their bottom but content is cut. The issue is observed on Win10 64bit.

scrollbar-issue-cannot-scroll-to-the-bottom

Empact:
Fixes #311

### DIFF
--- a/app/routes/activity/components/Activity.scss
+++ b/app/routes/activity/components/Activity.scss
@@ -32,7 +32,6 @@
 
 .activities {
 	background: $white;
-	height: 100%;
 
 	.header {
 	  background: $spaceblue;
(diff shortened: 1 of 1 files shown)


Empact:
screen shot 2018-03-16 at 12 08 10 pm

Empact:
This uses calc() - better practice might be to switch to flexbox.

screen shot 2018-03-16 at 12 09 30 pm

odb366:
Great fix @Empact . I've tested it and it works. However, I see horizontal scrollbar now when width is less than ~1000px. It can be removed with `overflow-x: hidden;` on the activities container.

Empact:
@odb366 could you take another look? It's looking good to me now.

odb366:
@Empact Yep, looks good. Today I'll fix the scrolling issue for the networking area.

PR #347, opened 2018-03-21
fix for the scrollbar issue in the network list

odb366:
fix for the same issue as #311 but for the network list.


Empact:
I don't see this issue locally - could you post before and after?

odb366:
# **Before**
before-1

before-2

# **After**
after-1

after-2

Empact:
Thanks!

PR #349, opened 2018-03-22
Exclude expired requests and funding transactions from 'All'

Empact:
So this contradicts the naming 'All', but I would argue these two activity types are relatively unhelpful to the user and deserve to be separated out.

"The theory being that channel funding is relevant to the
building of the network, but more of an implementation
detail than a classic user-relevant data-point."


mrfelton:
@Empact can you rebase this PR please?

JimmyMow:
Closing this for #697

PR #359, opened 2018-04-04
Feature/new network form

JimmyMow:


odb366:
Some UX feedback on the new network form:

- there is considerable lag when opening the form (win10/64bit)


- initially when the form is triggered it is empty. Users would like to have some examples of what to search for. Most of them won't know how a node/channel address looks like. Displaying a list of nodes will give them directions for what to search. 

- we could use `autofocus` attribute on the input form `AddChannel__header`, so the users can start typing straight away rather than pointing to the field and then clicking. It saves user effort

- `AddChannel__header` uses some blue gradient as background which diverges from the ZAP style. Better use darker non-gradient background like `#252832`

- we can include a magnifying glass icon so we can denote a little bit better that this is a search field.

- When trying to connect to a peer the "Add funds to network" form is badly formatted. The header and subheader are center aligned, while the rest of the fields are left aligned. This creates more visual search effort for the users. The subheader text should be left-aligned. Also, the input fields don't have any borders. This makes them read only text for the users. If they don't click on the field they won't understand it is a input one. This is minimalistic but not usable. (Same for Pay and Request forms).

Anyway, great work ;)

JimmyMow:
thanks for the feedback @odb366!

1) When did you pull the branch? The lag should be removed via this commit `08d263c`. I myself have no lag

2) We could default to showing some nodes. Maybe just a form placeholder is enough?

3) Autofocus there is tricky because there are many forms on the same page. We could add autofocus but for time and priorities sake I did not. I agree with you though

4) That's actually going to be the style of all our searches going forward. I'm open to changing them if the are universally dislike by I'm going to implemented them throughout the app first

5) I like the magnifying glass idea

6) I agree. Others wanted center aligned, I will bring it up in slack :p

odb366:
1. Pulled on the `41dfede` commit. However, after reload of the app it doesn't lag anymore. Don't know why.
2. We already have a placeholder, and it is not enough. Seeing the results and filtering them will be a great example of what to do.
3. Ok
4. Will discuss this in slack
5. Yep
6. Already proposed redesigned modal following usability laws.

PR #363, opened 2018-04-05
fix(reformat-request-form): fixed bad formatting

odb366:
Redesigned the request form so that it complies with usability laws and fits well on the default window size of ZAP.

zap-request-form

### DIFF
--- a/app/components/Form/Request.js
+++ b/app/components/Form/Request.js
@@ -39,21 +39,6 @@ const Request = ({
       </header>
 
       <div className={styles.content}>
-        <section className={styles.memo}>
-          <div className={styles.top}>
-            <label htmlFor='memo'>Memo</label>
-          </div>
-          <div className={styles.bottom}>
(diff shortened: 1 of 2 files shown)


Empact:
Is there a particular reason for leading with the amount?

odb366:
In slack, at #desktop-design we discussed the request screen and @brandon shared a wireframe where the amount field was on top. We kind-a agreed to switch the fields there.

Empact:
@odb366 Could you fix the test failure? It looks related.

Given the amount is required and the memo is not, I think leading with it is reasonable.

odb366:
@Empact Done. Should I rebase?

JimmyMow:
utACK https://github.com/LN-Zap/zap-desktop/pull/363/commits/193006aa0d048b97ac425ebf02dc352cdff73f90

I like the changes. Merging!

PR #364, opened 2018-04-06
fix(reformat-pay-form): fixed bad formatting

odb366:
Redesigned the pay form so that it complies with usability laws and fits well on the default window size of ZAP. Feedback is welcome.

zap-pay-form

### DIFF
--- a/app/components/Form/Pay.js
+++ b/app/components/Form/Pay.js
@@ -106,12 +106,12 @@ class Pay extends Component {
             <div className={styles.bottom}>
               <textarea
                 type='text'
-                placeholder='Payment request or bitcoin address'
+                placeholder='Paste payment request or bitcoin address here'
                 value={payInput}
                 onChange={event => setPayInput(event.target.value)}
                 onBlur={onPayInputBlur}
                 id='paymentRequest'
-                rows='2'
+                rows='4'
               />
               <section className={`${styles.errorMessage} ${showErrors.payInput && styles.active}`}>
                 {showErrors.payInput &&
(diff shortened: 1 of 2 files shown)


Empact:
What usability laws do you mean?
utACK 0182d4b

odb366:
zap-pay-form-inactive

Above you can see the current Pay form (without usability laws implemented) for illustration.

Here's which one's I think it violates:

- Before, the input fields didn't had borders, and looked like plain text. This breaches [Jacob Nielsen's consistency heuristic](https://www.nngroup.com/articles/ten-usability-heuristics/), which means we're not consistent with majority of other interfaces by not having the standard representation of input fields with borders. This leads to user confusion, and errors.

- At default window resolution the form was too long and the pay button was cutoff (bug). At full screen the submit button was too far away from the input fields. This surely leads to increased selection times of the submit button according to [Fitts's Law](https://www.interaction-design.org/literature/book/the-glossary-of-human-computer-interaction/fitts-s-law). Now, the width is fixed, content is centered and the button is closer to the form saving user effort and time.

- The [labels were too far away from the input fields](https://www.nngroup.com/articles/form-design-white-space/) which could lead users of thinking labels are separate things from the input fields. According to [Gestalts law](http://graphicdesign.spokanefalls.edu/tutorials/process/gestaltprinciples/gestaltprinc.htm) of proximity if elements are visualized closely to each other the brain perceives them as related. However, the more distant they are, the less they will be perceived as a group.

JimmyMow:
utACK https://github.com/LN-Zap/zap-desktop/commit/0182d4b351b059cff9c999ff382f30b5e3b77068

merging

Issue #389, opened 2018-05-10
Autofocus for certain input fields


odb366:
We need to add `autoFocus` attribute to the 1st input field of Pay, Request, Search and Connect to peer screens.

Autofocus saves user effort by automatically selecting the desired input field of a form.

PR #390, opened 2018-05-10
feature(autofocus): implemented autoFocus where it is useful

odb366:
Fix for issue #389 
Used the `autoFocus` prop instead of `ref={input => input && input.focus()}` for setting up the autofocus attribute where needed.
I've tested it and it works fine on my end (win10).

### DIFF
--- a/app/components/Contacts/AddChannel.js
+++ b/app/components/Contacts/AddChannel.js
@@ -97,6 +97,7 @@ const AddChannel = ({
           value={contactsform.searchQuery}
           onChange={event => searchUpdated(event.target.value)}
           // ref={input => input && input.focus()}
+          autoFocus
         />
         <span onClick={closeContactsForm} className={styles.closeIcon}>
           <Isvg src={x} />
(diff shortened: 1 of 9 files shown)


odb366:
The test fails because lint thinks that [autoFocus](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/HEAD/docs/rules/no-autofocus.md) can cause [usability or accessibility problems](https://w3c.github.io/html/sec-forms.html#autofocusing-a-form-control-the-autofocus-attribute). However, the W3C recommendation is too general, and in our case I don't see any usability/accessibility issues.  

According to my knowledge `autoFocus` can cause [usability/accesibility problems](https://stackoverflow.com/questions/2180645/is-automatically-assigning-focus-bad-for-accessibility) when:
- the page is long and users are unintentionally moved down the page. 
- there are a lot of input fields and users want to start with other than the focused one.
- `autoFocus` is applied on buttons and can force accidental form submissions.

So we have usability/accessibility issues when the page is long or have lots of inputs or it is used on a button. 

In ZAP all of the forms consist of 1 or 2 input fields which are immediately visible (users don't have to scroll down). No long pages, no complex forms. Therefore, implementing `autoFocus` we improve the usability rather than decrease it. 

My advice is to disregard these lint errors and use `autoFocus` if we don't have any other issues with it.

PR #395, opened 2018-05-22
fix(channels): do not show suggested nodes on empty search

### LINKED ISSUE #394: Unable to edit channel search query if no results are found
At the bottom of the channel list there is an input box that lets you filter the channel list based on node alias or pubKey. If you type something into this box that returns no results, the input disappears and the message `Hmmm, looks like you don\'t have any channels yet. Here are some suggested nodes to open a channel with to get started` displays at the top of the channel list.

There is no way to recover from this short of restarting the app (you can't remove the invalid search query since the input is no longer visible).

mrfelton:
Do not show the suggested nodes text block when a channel search does not return any results. Instead, in this case retain the search query input filter on the screen so that the user can go back and edit their search query.

Fix #394

### DIFF
--- a/app/components/Contacts/Network.js
+++ b/app/components/Contacts/Network.js
@@ -29,7 +29,8 @@ class Network extends Component {
         filter,
         selectedChannel,
         loadingChannelPubkeys,
-        closingChannelIds
+        closingChannelIds,
+        channels
       },
       currentChannels,
       balance,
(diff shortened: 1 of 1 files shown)


JimmyMow:
utACK

@Empact this actually fixes a pretty bad bug. let's merge this unless you see anything you don't like

Empact:
Prefer we keep the `> 0` for explicitness, but that's a nit so let's merge.

PR #397, opened 2018-05-22
fix(channels): filter channel list on node alias

### LINKED ISSUE #396: Filtering on node name/alias not working
Trying to filter on channel name does not work (always shows no results). Only filtering on node id works. The placeholder text in the channel search input says "search by alias or pubkey" so searching on an alias should work in addition to the pubkey.

mrfelton:
Update channel filter handlers to correctly consider the node alias.

Fix #396

### DIFF
--- a/app/reducers/channels.js
+++ b/app/reducers/channels.js
@@ -386,6 +386,17 @@ const filtersSelector = state => state.channels.filters
 const filterSelector = state => state.channels.filter
 const nodesSelector = state => state.network.nodes
 
+const channelMatchesQuery = (channel, nodes, searchQuery) => {
+  const node = nodes.find(n => channel.remote_pubkey === n.pub_key)
+  const query = searchQuery.toLowerCase()
+
+  const remoteNodePub = (channel.remote_node_pub || '').toLowerCase()
+  const remotePubkey = (channel.remote_pubkey || '').toLowerCase()
+  const displayName = (node ? node.alias : '' || '').toLowerCase()
+
+  return remoteNodePub.includes(query) || remotePubkey.includes(query) || displayName.includes(query)
+}
+
 channelsSelectors.channelModalOpen = createSelector(
   channelSelector,
   channel => (!!channel)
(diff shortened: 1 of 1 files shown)


app/reducers/channels.js
```diff
@@ -386,6 +387,15 @@ const filtersSelector = state => state.channels.filters
 const filterSelector = state => state.channels.filter
 const nodesSelector = state => state.network.nodes
 
+const displayNodeName = (channel, nodes) => {
+  const node = find(nodes, n => channel.remote_pubkey === n.pub_key)
```
Empact: I think you can use regular `find` here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

app/reducers/channels.js
```diff
@@ -386,6 +387,15 @@ const filtersSelector = state => state.channels.filters
...
+
+  if (node && node.alias.length) { return node.alias }
+  return channel.remote_pubkey ? channel.remote_pubkey.substring(0, 10) : channel.remote_node_pub.substring(0, 10)
```
Empact: Is this redundant with the similar checks on the lines below? How about making this check alias-specific?

app/reducers/channels.js
```diff
@@ -386,6 +387,15 @@ const filtersSelector = state => state.channels.filters
...
+}
+
+const caseInsensitiveIncludes = (left = '', right = '') => left.toLowerCase().includes(right.toLowerCase())
```
Empact: I would inline this and DRY up another way - e.g. maybe call `filterChannel`/`filterPendingChannel` within the `channelArray` filter loop?

mrfelton:
@Empact could you give a re-review please? I've cleaned this up and simplified things a little.

Empact:
utACK 8d448d7

app/reducers/channels.js
```diff
@@ -386,6 +387,17 @@ const filtersSelector = state => state.channels.filters
 const filterSelector = state => state.channels.filter
 const nodesSelector = state => state.network.nodes
 
+const channelMatchesQuery = (channel, nodes, searchQuery) => {
+  const node = nodes.find(n => channel.remote_pubkey === n.pub_key)
+  const query = searchQuery.toLowerCase()
+
+  const remoteNodePub = get(channel, 'remote_node_pub', '').toLowerCase()
+  const remotePubkey = get(channel, 'remote_pubkey', '').toLowerCase()
+  const displayName = get(node, 'alias', '').toLowerCase()
```
Empact: nit: I think this would be nearly as simple without `lodash/get`

mrfelton:
Thanks @Empact - I've updated to remove the lodash dependency.

app/reducers/channels.js
```diff
@@ -386,6 +386,17 @@ const filtersSelector = state => state.channels.filters
 const filterSelector = state => state.channels.filter
 const nodesSelector = state => state.network.nodes
 
+const channelMatchesQuery = (channel, nodes, searchQuery) => {
+  const node = nodes.find(n => channel.remote_pubkey === n.pub_key)
+  const query = searchQuery.toLowerCase()
+
+  const remoteNodePub = (channel.remote_node_pub || '').toLowerCase()
+  const remotePubkey = (channel.remote_pubkey || '').toLowerCase()
+  const displayName = (node ? node.alias : '' || '').toLowerCase()
```
Empact: nit: could call `.toLowerCase` on the values themselves

Empact:
re-utACK e8e6e9f

app/reducers/channels.js
mrfelton: I'm unsure what you mean. The values may be undefined in which case calling `toLowerCase` on it may result in an error. The approach taken here ensures that `toLowerCase` is called on a string regardless of wether the value was set or not.

Empact: Roger, no need to change.

PR #398, opened 2018-05-23
fix(pay-form): ensure payment form shows on top

### LINKED ISSUE #284: Connect dialog remains open when request payment link (from web) is clicked
screenshot 2018-02-04 16 05 48

It should hide itself when that happens, or stay open but invisible and once payment is done, it can go back again.

For case like this I would use a popup stack and the payment would be just another popup that would be removed form the stack once completed, not messing with whatever was opened before.

mrfelton:
This fixes an issue where the payment form - initiated from an external request payment link - would display under the connect dialog if that was already open. This change ensures that the payment form always displays on the top most layer.

Fix #284

### DIFF
--- a/app/components/Form/Form.scss
+++ b/app/components/Form/Form.scss
@@ -3,7 +3,7 @@
 .container {
   position: absolute;
   top: 0;
-  z-index: 10;
+  z-index: 20;
   height: 100vh;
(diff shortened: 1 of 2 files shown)


Empact:
utACK e586dc5

PR #401, opened 2018-05-28
fix(passwords): convert password to Buffer

### LINKED ISSUE #400: Unable to create new wallet with 8 character password
Trying to create a wallet with an 8 character password fails with the following error:

```
[0] initWallet error:  { Error: 2 UNKNOWN: password must have at least 8 characters
[0]     at new createStatusError (/Users/tom/workspace/zap-desktop/app/node_modules/grpc/src/client.js:64:15)
[0]     at /Users/tom/workspace/zap-desktop/app/node_modules/grpc/src/client.js:583:15
[0]   code: 2,
[0]   metadata: Metadata { _internal_repr: {} },
[0]   details: 'password must have at least 8 characters' }
```

Additionally, attempting to unlock an existing wallet fails.

mrfelton:
The wallet password is defined as `bytes` in the rpc.proto spec and we are
passing it as a string, resulting in an incorrect password being set on a
new wallet and an inability to unlock an existing wallet.

Convert the wallet password to a Buffer before passing to `initWallet` and
`unlockWallet`, which is the default data type used to represent bytes in
the node grpc client. This is sures that the password is passed correctly
to lnd.

Fix #400

### DIFF
--- a/app/lnd/methods/walletController.js
+++ b/app/lnd/methods/walletController.js
@@ -112,7 +112,7 @@ export function genSeed(walletUnlocker) {
  */
 export function unlockWallet(walletUnlocker, { wallet_password }) {
   return new Promise((resolve, reject) => {
-    walletUnlocker.unlockWallet({ wallet_password }, (err, data) => {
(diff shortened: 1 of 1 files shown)


JimmyMow:
👏

PR #403, opened 2018-06-01
 Fix the local lnd already running usecase

Empact:
This fixes the "LND ALREADY RUNNING" case to have the same handling as the custom LND case.

Absent this fix, Zap is stuck at the loading page on startup with local LND running.


JimmyMow:
utACK

@Empact do you wanna fix the loading page on startup with local LND in a separate PR or add to this one?

Empact:
@JimmyMow will make another PR for that

PR #406, opened 2018-06-05
Consolidate lnd build resources

mrfelton:
This is an updated version of #164 in which I have moved the `rpc.proto` file into the top level `resources` directory as well as updating `electron-builder` to the latest version and ensuring that the resources (lnd binary and rpc.proto file) make it into the correct location in the final builds.


app/lnd/config/index.js
```diff
@@ -35,15 +52,15 @@ switch (plat) {
 }
 
 if (process.env.NODE_ENV === 'development') {
-  lndPath = join(__dirname, '..', '..', '..', 'resources', 'bin', plat, lndBin)
+  lndPath = join(appRootPath(), 'resources', 'bin', plat, lndBin)
```
JimmyMow: wanna dry this up by declaring `appRootPath` a const above? then we don't have to call the func multiple times.

JimmyMow:
Concept ACK

One tiny comment, let me know what you think @mrfelton 

cc @Empact

mrfelton:
Good call @JimmyMow - I've cleaned that up a little.

PR #417, opened 2018-06-14
Drop ModalRoot

Empact:
None of the associated modals are used, each being defunct. And we're providing all existing modals independently. Seems we have little use for this component.


app/routes/app/components/App.js
```diff
@@ -110,13 +97,10 @@ class App extends Component {
 }
 
 App.propTypes = {
-  modal: PropTypes.object.isRequired,
-  ticker: PropTypes.object.isRequired,
   form: PropTypes.object.isRequired,
   formProps: PropTypes.object.isRequired,
   closeForm: PropTypes.func.isRequired,
   error: PropTypes.object.isRequired,
-  info: PropTypes.object.isRequired,
```
Empact: One question is should I include these even if they are not used in the component itself, given the children sometimes or always do? I presume not.

JimmyMow: yeah no need. we used to manually pass objects like `info` into child components within `App.js` _but_ I've since started to refactor with a much cleaner approach by creating prop objects in Containers. Example: https://github.com/LN-Zap/zap-desktop/blob/master/app/routes/app/containers/AppContainer.js#L193

JimmyMow:
utACK

Ready to merge when you are

Issue #420, opened 2018-06-14
Error building dependencies with yarn on Arch Linux


dmp1ce:
I just downloaded the code and tried to build the dependencies on Arch Linux with `yarn`. I'm not sure if I'm missing something obvious.

Here is my build output.
https://bpaste.net/show/d9de030db9ca

mrfelton:
@dmp1ce did you / can you try running `npm run install-grpc` after running `yarn`?

mrfelton:
@dmp1ce if you could test the above patch which updates the grpc package that would be great ;)

Updated code is in this fork/branch https://github.com/mrfelton/zap-desktop/tree/feat/update-grpc

dmp1ce:
https://bpaste.net/show/d32c16d9afc0 from the branch `feat/update-grpc`. I didn't try the step `npm run install-grpc` because I assumed the `yarn` command failed. Should I just continue on with running `lnd` and running Zap?

dmp1ce:
`npm run install-grpc` also had errors. https://bpaste.net/show/c2bf6b2ea5e1

mrfelton:
@dmp1ce I've been working on updating various build dependencies for the project to resolve some build related issues that I was facing. It's possible that some of these updates could resolve your issues too.

Could you try with the code from this PR https://github.com/LN-Zap/zap-desktop/pull/426 (code is in this branch https://github.com/mrfelton/zap-desktop/tree/feat/update-tooling)

dmp1ce:
I'm not sure I'm building grpc-native-core correctly but I get these errors.

https://bpaste.net/show/659c8d50e916

mrfelton:
@dmp1ce Can you open a support request at https://github.com/grpc/grpc/issues/new please? They should be able to help you better.

nicolasnoble:
The actual problem is here:

https://github.com/LN-Zap/zap-desktop/blob/master/app/package.json#L14

Please don't do this. `electron-rebuild` isn't doing things correctly by skipping hooks, and will break grpc every now and then, like with this issue here. Moreover, the only advantage of using `electron-rebuild` is to automatically rebuild for the installed version of electron, so you don't need to specify the target, like you're doing. So effectively, you're getting into a situation where:

-) You are creating more problems than it's worth by using `electron-rebuild`.
-) You don't get the only advantage that using `electron-rebuild` is supposed to provide.

[...]

PR #423, opened 2018-06-14
fix(grpc): update grpc to v1.12.3

mrfelton:
Update grpc to v1.12.3 in order to fix issues build issues due to the fact
that 1.12.1 doesn't contain node-pre-gyp as a bundled dependency.

See https://github.com/grpc/grpc-node/issues/365

This may be linked to the issue reported in #420 whose build logs (https://bpaste.net/show/d9de030db9ca) show the warning

> warning "grpc@1.12.1" is missing a bundled dependency "node-pre-gyp". This should be reported to the package maintainer.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -15,7 +15,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "grpc": "1.12.1",
+    "grpc": "^1.12.3",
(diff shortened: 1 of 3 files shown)


Empact:
utACK 47773b4

Issue #415, opened 2018-06-12
Missing password validation during onboarding


odb366:
During onboarding process users are asked to create a pass for unlocking ZAP. The pass should be 8 or more characters, but users are able to create one with less than 8 characters and proceed. Inline validation is missing. Consequently, on the last step of the onboarding they cannot proceed, and are unable to start ZAP.

Update: the user can even proceed without entering the confirmation password on this screen.

![screenshot 32](https://user-images.githubusercontent.com/14116101/41280696-41780ca4-6e38-11e8-919f-7719e00b024e.png)

odb366:
Wow, how did I missed that. Thanks @mrfelton. Fix incoming.

odb366:
Another usability issue is that we don't have a back button on this screen. Should I add back functionality to the autopilot step?

JimmyMow:
No. That's more complex than you think. We'd have to shut down our LND process we just started. No back button there

odb366:
@JimmyMow Ok

odb366:
Issue fixed, closing

odb366:
The issue is reopened because it is possible to proceed with 2 different passwords ..

PR #424, opened 2018-06-15
fix(onboarding-pass-validation): added password validation

odb366:
Fixes issue #415 
Now, users are not allowed to proceed with the onboarding process if their wallet pass is less than 8 chars.


app/components/Onboarding/NewWalletPassword.scss
```diff
@@ -35,3 +35,10 @@
     visibility: visible;
   }
 }
+
+.helpMessage {
```
JimmyMow: currently we highlight the bottom border of the input that isn;t correct and make the border `$red`. we also make the error text `$red`. can you make those style changes?

app/components/Onboarding/NewWalletPassword.js
```diff
@@ -35,6 +36,9 @@ const NewWalletPassword = ({
       >
         Passwords do not match
       </p>
+      <p className={`${styles.helpMessage} ${passwordMinChars && styles.visible}`}>
```
JimmyMow: this visible class doesn't apply to `helpMessage` right? It looks like it's nested inside `errorMessage` in the `scss` file.

Also this logic seems to apply `visible` when the passwords match and the char count is 8 or over. Shouldn't it be the other way around?

JimmyMow: When I'm testing this PR the error message shows up before I start typing anything

JimmyMow:
@odb366 I was just testing this and added some comments/feedback

Empact:
Could you also squash + rebase the commits?

app/components/Onboarding/NewWalletPassword.js
odb366: The help message is intended to always stay visible for better usability. I've just forgotten to clean it up the code.
If the help message is hidden some users will try to create passwords with less than 8 chars and then fail. They will be forced to rethink and resubmit their pass which is considerable user effort. We can prevent such cases by leaving the help message always visible. See [this article](https://www.nngroup.com/articles/password-creation/).

app/components/Onboarding/NewWalletPassword.scss
odb366: Since we use input field with all borders making just the bottom one red will look strange. I haven't seen this pattern in sites/apps, so in my opinion it will not be consistent in usability point of view. I've made all borders and error text to get $red when error occurs. Is that Ok?

odb366:
Thanks for the feedback guys. However, I can't commit because i got the following error. Even when I swich to my local master branch which is up-to-date i still got this error: 
```
 webpack.config.renderer.prod.js:1:1
  ‼    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-named-as-default-member
  ‼    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-named-as-default
  ‼    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-duplicates
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/extensions
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/default
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-unresolved
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/namespace
  ×    8:34  Missing file extension for "csp-html-webpack-plugin"             import/extensions
  ×    8:34  Unable to resolve path to module csp-html-webpack-plugin.        import/no-unresolved

  webpack.config.renderer.dev.js:1:1
  ‼    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-duplicates
  ‼    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-named-as-default
  ‼    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-named-as-default-member
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/no-unresolved
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/extensions
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/default
  ×    1:1   Resolve error: Cannot find module add-asset-html-webpack-plugin  import/namespace
  ×   17:32  Unable to resolve path to module add-asset-html-webpack-plugin.  import/no-unresolved
  ×   17:32  Missing file extension for "add-asset-html-webpack-plugin"       import/extensions
  ×   18:34  Missing file extension for "csp-html-webpack-plugin"             import/extensions
```

mrfelton:
@odb366 you need to run `yarn` whenever you pull new code from github in order to install any new dependencies that might have been added.

odb366:
@mrfelton Thanks for your lightning quick answer :)

odb366:
Latest commit improves the password validation by highlighting the always visible help message `Password must be at least 8 characters long` in red when both passwords are of same length but less than 8 characters. This way the validation message won't appear immediately when you start typing in the `Confirmation Password` field.

The same way inline validation about password match is fixed in this commit. It won't appear immediately after typing in the second field, but when the users has typed equal amount of characters as  in the first field.

error-message-design2

JimmyMow:
Tested ACK 1ef2c96

Nice @odb366 🎊

Empact:
utACK 1ef2c96

PR #426, opened 2018-06-15
Update tooling

mrfelton:
This PR updates several components from the build system in order to speed up and prevent hanging processes during the `yarn` install process.

The most notable changes are:

 - Updates electron-builder to the latest version which resolves some local build issues
 - Updates a number of the supporting dependencies to more recent versions
 - Updates Webpack to v4.x
 - Small optimisation for the travis builds to prevent them from compiling grpc twice (shaves several minutes of the builds)

Full details and reasoning for the updates are detailed in the individual commit messages.


Empact:
Tested ACK f50dd34

PR #428, opened 2018-06-15
Update electron to v2.x

mrfelton:
This PR updates the app to use Electron 2.x:

- Update all electron packages to the latest v2.x compatible versions
- Implement a Content-Security-Policy to define where scripts & content etc can be loaded from

Everything seems to work as expected in my limited testing.

I have set `ELECTRON_DISABLE_SECURITY_WARNINGS` on travis to prevent security warnings related to CSP from failing the tests (the warnings are because when the code is compiled in dev mode we have `devtool: 'eval',` set in `webpack.config.renderer.dev.dll.js` which results in code that includes `eval` statements. Because of this, we must include `unsafe-eval` in our CSP.

Note: This initial CSP that I have defined includes three following remote URLs:

*content:*
 - https://api.coinmarketcap.com
 - http://zap.jackmallers.com
 - https://testnet-api.smartbit.com.au;

*fonts/styles:*
 - https://fonts.googleapis.com
 - https://s3.amazonaws.com
 - https://fonts.gstatic.com

This policy could do with some more scrutiny.

Ultimately, we may want to have different polocies for dev vs production builds.


app/app.html
```diff
@@ -2,9 +2,15 @@
 <html>
   <head>
     <meta charset="utf-8">
+    <meta http-equiv="Content-Security-Policy" content="
+      default-src 'self';
+      connect-src 'self' http://localhost:* ws://localhost:* https://api.coinmarketcap.com http://zap.jackmallers.com https://testnet-api.smartbit.com.au;
+      script-src 'self' http://localhost:* 'unsafe-eval' 'unsafe-inline';
+      font-src 'self' data: http://localhost:* https://fonts.googleapis.com https://s3.amazonaws.com https://fonts.gstatic.com;
```
Empact: https://s3.amazonaws.com is pretty broad, how about narrowing that down?

app/app.html
```diff
@@ -2,9 +2,15 @@
 <html>
   <head>
     <meta charset="utf-8">
+    <meta http-equiv="Content-Security-Policy" content="
+      default-src 'self';
+      connect-src 'self' http://localhost:* ws://localhost:* https://api.coinmarketcap.com http://zap.jackmallers.com https://testnet-api.smartbit.com.au;
```
Empact: https://zap.jackmallers.com/?

Empact:
Concept ACK

app/app.html
mrfelton: Currently, we pull the list of suggested nodes from https://zap.jackmallers.com/suggested-peers

app/app.html
mrfelton: I don't know that it is possible to be more specific than the host level in a CSP.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/font-src

app/app.html
Empact: How about using http instead of https in the CSP?

app/app.html
Empact: Fair enough then we would have to expose s3 assets via our own url. This works as an improvement without that.

app/app.html
mrfelton: When webpack devserver runs, it runs over http, so all the calls to localhost go over http. It is possible to tell webpack devserver to run using https however when you do that it generates and uses a self signed certificate which electron doesn't know about or trust - so electron will not run correctly as it warns about the unknown/invalid certificate

Empact: I mean specifically for accessing `zap.jackmallers.com`. Agree `http://localhost` is fine.

mrfelton: Ahh yes. I had that change on local already but wasn't included in what I pushed for some reason. Have updated now.

Empact:
Needs rebase

mrfelton:
Rebased

PR #429, opened 2018-06-15
Don't run onboarding if LND is already running locally

Empact:
The onboarding settings are only relevant if the user wants to connect to a
remote LND or have Zap run LND. If the LND is already running locally, the user
is capable of separately configuring it.

This does not support the case of connecting to a remote LND while running LND
locally, but I can't think of a reason why someone would do such a thing. In
such a case, they could quit LND, start Zap, then re-start LND.

### DIFF
--- a/app/main.dev.js
+++ b/app/main.dev.js
@@ -274,8 +274,25 @@ app.on('ready', async () => {
 
   sendGrpcDisconnected()
 
-  // Let the application know onboarding has started.
-  sendStartOnboarding()
+  console.log('LOOKING FOR LOCAL LND')
+  // Check to see if an LND process is running.
+  lookup({ command: 'lnd' }, (err, results) => {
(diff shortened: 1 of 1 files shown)


app/main.dev.js
```diff
@@ -295,6 +312,7 @@ app.on('ready', async () => {
     console.log('SAVED CONFIG TO:', store.path, 'AS', store.store)
 
     if (options.connectionType === 'local') {
+<<<<<<< Updated upstream
```
mrfelton: Looks like a merge conflict here.

Empact: FIxed

mrfelton:
Tested ACK https://github.com/LN-Zap/zap-desktop/pull/429/commits/0ce5b04fe51e81bd1a4915c334cba12ccf1cb29b

PR #433, opened 2018-06-17
More reliable lnd log parsing

### LINKED ISSUE #432: lnd log output parsing is unreliable (could cause app to hang at startup)
After the application starts up lnd it attempts to parse its log output in order to determine and react to the current status of lnd. In the code, we have this:

```javascript
  // Listen for when neutrino prints out data
  neutrino.stdout.on('data', (data) => {
    // Data stored in variable line, log line to the console
    const line = data.toString('utf8')
```

See https://github.com/LN-Zap/zap-desktop/blob/master/app/main.dev.js#L174-L177

However, data emitted from a stream's `data` event is actually a buffer of a fixed length which does not directly map to a single line of log output. So the assumption that the `data` variable is a single line of logout is incorrect.

This means that when we attempt to parse each line and do things like this:

```javascript
    // If the gRPC proxy has started we can start ours
    if (line.includes('gRPC proxy started')) {
      ...
```

The result can be unreliable, since the actual log line that contains the string `gRPC proxy started` could actually be split over 2 `data` events.

eg:

`data` event 1 = `...gRPC pr`
`data` event 2 = `oxy started...`

This could cause the app to miss important lines of lnd log output and result in the app hanging at startup.

mrfelton:
Pipe lnd stdout tream through `split2` in order to ensure that each data
event contains a single line of log output.

Fix #432

### DIFF
--- a/app/main.dev.js
+++ b/app/main.dev.js
@@ -13,6 +13,7 @@
 import { app, BrowserWindow, ipcMain, dialog } from 'electron'
 import path from 'path'
 import fs from 'fs'
+import split2 from 'split2'
 import { spawn } from 'child_process'
 import { lookup } from 'ps-node'
 import Store from 'electron-store'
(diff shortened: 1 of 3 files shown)


Empact:
utACK 1b702e9 - I'm assuming the utf8 handling is incidental / inconsequential.

mrfelton:
Yes, that is only needed when converting a Buffer to a string. `split2` handles that.

PR #434, opened 2018-06-17
Remove redundant install steps

mrfelton:
Remove the `install-grpc` installation step as it is redundant.

See #420


Empact:
Tested ACK 09fe241 - confirmed that grpc is built on package install now.

PR #436, opened 2018-06-17
Improve log readability and allow control over what is logged

mrfelton:
Currently, the log output is messy and hard to follow.

This PR aims to improve the application logging with a more comprehensive and flexible logging
solution. We leverage https://github.com/appscot/debug-logger for this.

We use two separate log handlers to provide separation between electron logs (namespace=`zap:main`) and and lnd logs (namespace=`zap:lnd`). This provides a good level of control over what log out put you want to see. Some examples:

Only show log messages in the `zap:main` namespace
```
DEBUG=zap:main npm run dev
```

Only show log messages of level `warn` and above
```
DEBUG_LEVEL=warn npm run dev
```

The default configuration will show `debug` and above for all messages in the `zap:*` namespace.

**BEFORE:**
screenshot 2018-06-17 18 45 30

**AFTER:**
screenshot 2018-06-17 18 43 56

PR #437, opened 2018-06-17
Convert remaining console statements to log or throw, enable no-console linter

Empact:
This follow up on #436 by converting remaining `console.log` statements to use the log or throw, and enabling the `no-console` linter to prevent addition in the future.

As a bonus, this allows us to enable the full `eslint:recommended` set, which is a good baseline.
https://eslint.org/docs/rules/

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,6 +6,7 @@
   },
   "extends": [
     "airbnb",
+    "eslint:recommended",
     "plugin:react/recommended",
     "plugin:flowtype/recommended",
     "plugin:import/errors",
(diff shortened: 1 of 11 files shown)


mrfelton:
@Empact This is a good improvement. My only real question is wether the `lndLog` is the best logging instance to use for this.

My original intention of the namespaces were as follows:

`zap:main` - things that happen from the Main process
`zap:lnd` - the raw lnd binary logs

(I also want to get `lnd:renderer` working at some point in order to enable console logging from the renderrer - but there is a little more to that).

Whilst the console.log statements that you have updated here do sort of relate to lnd, they are not coming directly from lnd - so in my eyes `zap:main` would be more appropriate.

Alternatively, we could introduce a new namespace so that we had for example:

`zap:main` - things that happen from the Main process
`zap:lnd` - logs that relate to lnd
`zap:lnd:binary` - the raw lnd binary logs

PR #440, opened 2018-06-18
Developer workflow improvements

mrfelton:
Following on from the work in https://github.com/LN-Zap/zap-desktop/pull/439 I've made several additional improvements to the developer workflow here.

The goal of this PR is to:
- provide consistency in development standards
- automate common development workflow tasks
- provide additional help for new contributors

The updates include:
 - Integration with [`prettier`](https://github.com/prettier/prettier) via [`prettier-eslint`](https://github.com/prettier/prettier-eslint)
- Automatic code formatting and linting in git `pre-commit` hook
- Automatic commit message format validation in git `commit-msg` hook
- Above commit hooks are opt-in (enable them by creating a `.opt-in` file (see updated docs)
- Multiple updates to documentation:

    **Readme (see https://github.com/mrfelton/zap-desktop/blob/build/developer-workflow/README.md)**
     - Update format to conform to `standard-readme` format
     - Tidy up header section
     - Add table of contents
     - Add security section
     - Add help section
     - List maintainers
     - Add contribution notes
     - Add contributors list

    **Contributing (see https://github.com/mrfelton/zap-desktop/blob/build/developer-workflow/CONTRIBUTING.md)**
     - Add table of contents
     - Add contribution faq
     - Add contribution guidelines

     **Advanced (see https://github.com/mrfelton/zap-desktop/blob/build/developer-workflow/ADVANCED.md)**
      - Add table of contents

PR #443, opened 2018-06-18
style(lint): lint and format all code

mrfelton:
Run all code through updated eslint/prettier setup

Follow on from https://github.com/LN-Zap/zap-desktop/pull/440


mrfelton:
Note: The travis build will fail as the code is reformatted to the updated rules from https://github.com/LN-Zap/zap-desktop/pull/440. That needs to be committed first.

PR #445, opened 2018-06-19
build: do not build grpc-node from source

mrfelton:
grpc-node provides node bindings that are specifically targeted towards
electrum. Make use of these rather than force compiling grpc on install.

See https://github.com/LN-Zap/zap-desktop/issues/420


Empact:
Tested ACK 556e541

For further reference: https://electronjs.org/docs/tutorial/using-native-node-modules#manually-building-for-electron

Issue #446, opened 2018-06-19
UX/UI Connect to node dialog has bad formatting and readability


odb366:
The connect to peer dialog lacks input field borders, which makes fields look like text. Also, it is lacking adequate container and white spacing.

![connect-to-peer](https://user-images.githubusercontent.com/14116101/41592032-1049a75a-73c4-11e8-8209-3335d6e97852.png)

PR #447, opened 2018-06-19
fix(submit-channel-layout): refactor form layout

odb366:
Fixes issue #446. 
- fixed form layout
- added input field border
refactored-add-channel-form


JimmyMow:
Tested ACK ad7a08c

PR #448, opened 2018-06-19
Fix build issues on AppVeyor

mrfelton:
Builds are failing on on AppVeyor. This is due to the fact that those builds are targeting node v7, which is not supported (as per package.json `engines`).

This PR updates the AppVeyor config to get the builds passing again by specifying node v8 and v10 as our build targets.

Additionally, I have updated the travis.yml file to also specify node v8 and v10 for consistency (was previously specifying node v8 and latest - which could result in broken builds when node decide to release a new version of node)

Eg passing tests visible at https://ci.appveyor.com/project/mrfelton/zap-desktop

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,8 +2,8 @@ os: unstable
 
 environment:
   matrix:
+    - nodejs_version: 10
     - nodejs_version: 8
-    - nodejs_version: 7
 
 cache:
   - "%LOCALAPPDATA%/Yarn"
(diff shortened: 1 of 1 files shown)


Empact:
This is great! But could you add back node? As it is passing currently, the additional testing can serve as a warning of future breakage, and can help us engage w the node developers if needed. If the build is failing in a disruptful way, we can add it to the allowed failures.

mrfelton:
Fair enough. I've added that back in. Originally I was looking to tell AppVeyor to use the latest node version, but I didn't find a way to do it so I hard set it to v10. Just figured it would be good to be consistent across our test infrastructure..

PR #449, opened 2018-06-19
docs(slack-link): replace the expired slack link with a link that does…

JimmyMow:
… not expire

I finally figured out how to get a slack invite that does not expire. I have updated the slack invite on `zap.jackmallers.com` and now with this PR our readme will be updated with an active slack invite as well.


mrfelton:
Nit - the commit type on this commit message should be `docs` instead of `fix. The `fix` scope should only be used for changes that actually fix bugs in the application code.

JimmyMow:
@mrfelton wow my bad I forgot about the PRs yesterday. Updated

PR #450, opened 2018-06-19
fix(line length): Limit line length to 100 chars

Empact:
150 is a bit too long


Empact:
Yeah I'm playing with lengths over here to see what works well. :P

PR #451, opened 2018-06-20
Ensure webpack build output is in colour

mrfelton:
pass the `--raw` flag to concurrently so that we see the raw output of our processes, rather than a version which has been mangled by concurrently. This ensures that the webpack build output is in colour more legible.

To test: run `npm run build`:

**before:**
screenshot 2018-06-20 17 57 49

**after:**
screenshot 2018-06-20 17 58 33

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "0.1.1",
   "description": "desktop application for the lightning network",
   "scripts": {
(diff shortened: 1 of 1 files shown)


JimmyMow:
Tested ACK `88f4fa2`

PR #456, opened 2018-06-21
Implement a more restrictive Content Security Policy for production builds

mrfelton:
Currently we hardcode our CSP directly into `app.html`. The current policy is geared towards development - it allows loading resources from localhost, inline scripts, and scripts that contain `eval`. Some of these settings are needed for our development builds, but not for production.

In this PR I have reworked things slightly to enable us to have differing CSP's for development vs production builds as well as more restrictive policy for development builds.

- First, we switch to using [`HtmlWebpackPlugin`](https://webpack.js.org/plugins/html-webpack-plugin/) to generate our final html file so that we can dynamically inject scripts into the page at compile time.

- We refactor the `app.html` page a little to remove all inline scripts and move these back to the relevant webpack configs.

- Next, we reimplement the CSP using [`CspHtmlWebpackPlugin`](https://github.com/slackhq/csp-html-webpack-plugin) so that we can define our CSP using dynamic code rather than hard coding meta tags directly on our html index page.

- Finally, we adjust our Content Security Policy on production so that it is more restrictive than when we run in development mode.

### DIFF
--- a/app/api/index.js
+++ b/app/api/index.js
@@ -1,7 +1,19 @@
 import axios from 'axios'
 
+// When running in development/hot mode we load the renderer js code via webpack dev server, and it is from there that
+// we ultimately initate requests to these remote resources. The end result is that the electron browser window makes a
+// request to localhost (to webpack dev server), which in turn makes a request to the remote resource. If the remote
+// resource in question has a restrictive `Access-Control-Allow-Origin` header, this may cause the electron browser
+// window to not allow loading the remote content.
+//
+// See https://enable-cors.org/
+//
+// In order to mitigate the CORS issue, we instead access these remote resources through a local proxy that we have
+// defined on the webpack dev server.
+const scheme = process.env.HOT ? '/proxy/' : 'https://'
+
 export function requestTicker(id) {
-  const BASE_URL = `https://api.coinmarketcap.com/v1/ticker/${id}/`
+  const BASE_URL = `${scheme}api.coinmarketcap.com/v1/ticker/${id}`
   return axios({
     method: 'get',
     url: BASE_URL
@@ -17,7 +29,7 @@ export function requestTickers(ids) {
 }
 
 export function requestBlockHeight() {
-  const BASE_URL = 'https://testnet-api.smartbit.com.au/v1/blockchain/blocks?limit=1'
+  const BASE_URL = `${scheme}testnet-api.smartbit.com.au/v1/blockchain/blocks?limit=1`
   return axios({
     method: 'get',
     url: BASE_URL
@@ -25,7 +37,7 @@ export function requestBlockHeight() {
 }
 
 export function requestSuggestedNodes() {
-  const BASE_URL = 'https://zap.jackmallers.com/suggested-peers'
+  const BASE_URL = `${scheme}zap.jackmallers.com/suggested-peers`
   return axios({
     method: 'get',
     url: BASE_URL
--- a/webpack.config.renderer.dev.js
+++ b/webpack.config.renderer.dev.js
@@ -13,6 +13,8 @@ import webpack from 'webpack'
 import merge from 'webpack-merge'
 import { spawn, execSync } from 'child_process'
 import ExtractTextPlugin from 'extract-text-webpack-plugin'
+import HtmlWebpackPlugin from 'html-webpack-plugin'
+import AddAssetHtmlPlugin from 'add-asset-html-webpack-plugin'
 import baseConfig from './webpack.config.base'
 import { mainLog } from './app/utils/log'
 
@@ -217,6 +219,15 @@ export default merge.smart(baseConfig, {
 
     new ExtractTextPlugin({
       filename: '[name].css'
+    }),
+
+    new HtmlWebpackPlugin({
+      template: path.join(__dirname, 'app', 'app.html')
+    }),
+
+    new AddAssetHtmlPlugin({
+      filepath: path.join(__dirname, 'dll', 'renderer.dev.dll.js'),
+      includeSourcemap: false
     })
   ],
 
--- a/webpack.config.renderer.dev.js
+++ b/webpack.config.renderer.dev.js
@@ -15,6 +15,7 @@ import { spawn, execSync } from 'child_process'
 import ExtractTextPlugin from 'extract-text-webpack-plugin'
 import HtmlWebpackPlugin from 'html-webpack-plugin'
 import AddAssetHtmlPlugin from 'add-asset-html-webpack-plugin'
+import CspHtmlWebpackPlugin from 'csp-html-webpack-plugin'
 import baseConfig from './webpack.config.base'
 import { mainLog } from './app/utils/log'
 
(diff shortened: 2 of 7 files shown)


Empact:
Haven't tested yet, but this https://github.com/slackhq/csp-html-webpack-plugin#basic-usage suggests that we need to include a template in our html (`<meta http-equiv="Content-Security-Policy" content="">`) for population.

webpack.config.renderer.dev.js
```diff
@@ -234,6 +237,31 @@ export default merge.smart(baseConfig, {
 
     new ExtractTextPlugin({
       filename: '[name].css'
+    }),
+
+    new HtmlWebpackPlugin({
+      template: path.resolve(__dirname, 'app', 'app.html')
```
Empact: I think we want to use `path.join` here, as `path.resolve` has functionality we're not relying on: https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths

webpack.config.renderer.dev.js
```diff
@@ -234,6 +237,31 @@ export default merge.smart(baseConfig, {
...
+    }),
+
+    new AddAssetHtmlPlugin({
+      filepath: path.resolve(__dirname, 'dll', 'renderer.dev.dll.js'),
```
Empact: I think we want to use `path.join` here, as `path.resolve` has functionality we're not relying on: https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths

webpack.config.renderer.dev.js
```diff
@@ -234,6 +237,31 @@ export default merge.smart(baseConfig, {
...
+        "'self'",
+        'http://localhost:*',
+        'ws://localhost:*',
+        'https://api.coinmarketcap.com',
+        'https://zap.jackmallers.com',
+        'https://testnet-api.smartbit.com.au'
+      ],
+      'script-src': ["'self'", 'http://localhost:*', "'unsafe-eval'"],
+      'font-src': ["'self'", 'data:', 'http://localhost:*', 'https://fonts.googleapis.com', 'https://s3.amazonaws.com', 'https://fonts.gstatic.com'],
+      'style-src': ["'self'", 'blob:', 'https://fonts.googleapis.com', 'https://s3.amazonaws.com', 'https://fonts.gstatic.com', "'unsafe-inline'"]
```
Empact: Is it possible to remove `unsafe-inline` now?

webpack.config.renderer.prod.js
```diff
@@ -161,6 +163,19 @@ export default merge.smart(baseConfig, {
     new BundleAnalyzerPlugin({
       analyzerMode: process.env.OPEN_ANALYZER === 'true' ? 'server' : 'disabled',
       openAnalyzer: process.env.OPEN_ANALYZER === 'true'
+    }),
+
+    new HtmlWebpackPlugin({
+      template: path.resolve(__dirname, 'app', 'app.html')
```
Empact: Same

webpack.config.renderer.dev.js
mrfelton: No. Without it we get errors like this:

> setInnerHTML.js:44 Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' blob: https://fonts.googleapis.com https://s3.amazonaws.com https://fonts.gstatic.com". Either the 'unsafe-inline' keyword, a hash ('sha256-E3dZVGA62Z1nIA/nziyL8qEhz+nm68ehiFEWHvx0rnc='), or a nonce ('nonce-...') is required to enable inline execution.

mrfelton:
Have update to use `path.join`

Empact:
Just starting to test - I'm getting a deprecation warning fyi:

```
(node:34236) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
    at AddAssetHtmlPlugin.apply (/Users/ben/Development/Bitcoin/zap-desktop/node_modules/add-asset-html-webpack-plugin/lib/index.js:10:14)
    at webpack (/Users/ben/Development/Bitcoin/zap-desktop/node_modules/webpack/lib/webpack.js:37:12)
    at startDevServer (/Users/ben/Development/Bitcoin/zap-desktop/node_modules/webpack-dev-server/bin/webpack-dev-server.js:379:16)
    at processOptions (/Users/ben/Development/Bitcoin/zap-desktop/node_modules/webpack-dev-server/bin/webpack-dev-server.js:361:5)
    at Object.<anonymous> (/Users/ben/Development/Bitcoin/zap-desktop/node_modules/webpack-dev-server/bin/webpack-dev-server.js:504:1)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/ben/Development/Bitcoin/zap-desktop/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
```

mrfelton:
Yes, I gave a thumbs up on the relevant issue already - https://github.com/SimenB/add-asset-html-webpack-plugin/issues/128

Empact:
Also some warnings:
screen shot 2018-06-21 at 15 55 01

mrfelton:
I've been getting that randomly. Pretty sure it's not related to these changes though

mrfelton:
I have a working solution for this @Empact which involves setting up a proxy on webpack dev server and accessing the remote resources via that. See last commit above.

app/api/index.js
```diff
@@ -1,7 +1,11 @@
 import axios from 'axios'
 
+// When running in development/hot mode where the application code is served up via webpack dev server
+// access these remote resources through a proxy that has been defined on the dev server to prevent CORS issues.
```
Empact: Can you be more explicit in describing the cors issues here?

app/api/index.js
```diff
@@ -17,15 +21,15 @@ export function requestTickers(ids) {
...
+  const BASE_URL = `${scheme}testnet-api.smartbit.com.au/v1/blockchain/blocks?limit=1`
   return axios({
     method: 'get',
     url: BASE_URL
   }).then(response => response.data)
 }
 
 export function requestSuggestedNodes() {
-  const BASE_URL = 'https://zap.jackmallers.com/suggested-peers'
+  const BASE_URL = `${scheme}zap.jackmallers.com//suggested-peers`
```
Empact: Apparently unintended extra `/`

Empact:
@jackmallers would it be possible to set up CSP on zap.jackmallers.com? Seems we just need `Access-Control-Allow-Origin: *`

mrfelton:
It's a little hard to describe the issue succinctly but I have tried my best! I've added a link to https://enable-cors.org/ where more info about CORS can be found.

Empact:
Tested ACK 7ab31e1 - ran through some basic flows, watched the logs

PR #458, opened 2018-06-21
lint: Minimize disabling of linters

Empact:
Declare a specific linter where appropriate, remove when unneeded.


Empact:
Merging as trivial (comment only)

PR #459, opened 2018-06-21
style: Limit prettier line length to 100 chars, with 120 chars allowed for long strings and comments

Empact:
Prettier line length and eslint max-length are intentionally different here, as they apply different standards to the code. Prettier ignores long strings, comments, etc. max-length is applied to those lines.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,6 @@
 {
   "bracketSpacing": true,
-  "printWidth": 120,
+  "printWidth": 100,
   "semi": false,
   "singleQuote": true,
   "trailingComma": "none",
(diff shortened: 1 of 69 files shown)


.prettierrc
```diff
@@ -1,6 +1,6 @@
 {
   "bracketSpacing": true,
-  "printWidth": 150,
+  "printWidth": 100,
```
mrfelton: Should this be `120` too?

mrfelton: As in - there is a conflict here with `max-len` from `.eslintrc` which you have set to 120.

Which is it - 100 or 120?

Empact: These rules have different application actually - prettier applies to the code that's being generated from the AST, which ignores certain things like strings and comments. Explanation here:
https://github.com/prettier/eslint-config-prettier#max-len

So 100 is the soft target applied to code generated by prettier, while 120 is the hard limit enforced by eslint.

PR #460, opened 2018-06-21
lint: Remove redundant eslint rules

Empact:
Indentified via eslint-config-prettier's check script. Prettier applies standards that directly enforce these.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,8 +23,6 @@
     "node": true
   },
   "rules": {
-    "comma-dangle": ["error", "never"],
-    "semi": ["error", "never"],
     "prettier/prettier": "error",
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
     "react/no-did-mount-set-state": 0,
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "lint-styles": "npm run lint-styles-base -- app/*.scss app/components/*.scss",
     "lint-styles-fix-base": "npm run lint-styles-base -- --fix",
     "lint-styles-fix": "npm run lint-styles-fix-base -- app/*.scss app/components/*.scss",
+    "lint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
     "lint-ci": "npm run lint && npm run lint-styles && npm run flow",
     "package": "npm run build && build --publish never",
     "package-all": "npm run build && build -mwl",


package.json
```diff
@@ -11,14 +11,15 @@
     "dev": "cross-env START_HOT=1 npm run start-renderer-dev",
     "flow": "flow",
     "flow-typed": "rimraf flow-typed/npm && flow-typed install --overwrite || true",
-    "lint-base": "eslint --cache --format=node_modules/eslint-formatter-pretty",
```
mrfelton: why have you disabled caching? I found this to make linting quite slow?

.eslintrc
```diff
@@ -23,8 +23,6 @@
     "node": true
   },
   "rules": {
-    "comma-dangle": ["error", "never"],
-    "semi": ["error", "never"],
```
mrfelton: prettier does disable these rules as a precautionary measure - but there is an advantage to having them enabled (as long as you define them in a way that matches the prettier config) which is that it enables linting tools (eg atom eslint plugin) to pick up and highlight any issues. Without the eslint rule enabled we only get auto-formatting from prettier, and no linting from eslint.

package.json
Empact: Thanks that was unintentional due to needing to disable caching when changing rules (because caching limits the run to just changed files, when the new lints should be run against all files when the rules change).

.eslintrc
Empact: Interesting, I'm using `linter-eslint` on atom which lints for mismatches between the prettier output and the actual code via the `prettier/prettier` rules. Is that an option for you?

The concern with having rules present which are automatically disabled by `eslint-config-prettier` is that they are effect the rules are potentially misleading "dead code", given that `.prettierrc` governs these. Better to have a single declaration, in the spirit of DRY.

Empact: Here's an example of that output btw: https://travis-ci.org/LN-Zap/zap-desktop/jobs/395132416

mrfelton: Actually, I just tested this in atom and it does indeed highlight errors with the `semi` and `comma-dangle` issues via the `prettier/prettier` rules when the rules are not directly present in `.eslintrc`. Maybe I had some caching issue when I was testing previously.

Given that, I think this one is good to go.

unACK https://github.com/LN-Zap/zap-desktop/pull/460/commits/3c08966fd98d9ae26ec397ff73b70a48a80c4ea0

PR #461, opened 2018-06-21
docs: add missing convention link to CONTRIBUTING page

Empact:
Compare: https://github.com/LN-Zap/zap-desktop/blob/master/CONTRIBUTING.md
With: https://github.com/Empact/zap-desktop/blob/docs/convention-link/CONTRIBUTING.md

Note I link to: https://conventionalcommits.org/
An alternative target is: https://github.com/marionebl/commitlint/tree/master/%40commitlint/config-conventional

I find them both pretty difficult to parse, so if you guys have another link in mind, lmk.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,5 +92,6 @@ This project has eslint rules and pull requests should pass `npm run lint` befor
 
 Tests should try to be written for every feature/fix and pass `npm run test` before being merged. With the demand for the Lightning Network and Zap rising, rapid development will naturally leave some code untested but we should all try our best.
 
+[convention]: https://conventionalcommits.org/
 [issues]: https://github.com/LN-Zap/zap-desktop/issues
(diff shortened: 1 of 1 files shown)


mrfelton:
The angular convention has better documentation that is easier to parse:

https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines

But it is not 100% inline with the conventional standard.I'm not sure that there is any meaningful difference for our purposes though. I think they only differ on the conventions for the `scope` (the optional part in brackets that goes after the `type`)

Empact:
Hmm. Then I guess going with https://conventionalcommits.org/ is a good option given it links to the angular docs as well as the other one. Sound good?

mrfelton:
I think that is ok. It provides a good overview of the basics. I think the only other thing that is useful for people to know is what types are valid in addition to `feat` and `fix` - but I already included this information directly on the `CONTRIBUTING.md` doc for easy reference purposes.

Also, if you have the `commit-msg` git hook enabled and try to commit with an invalid type, it will list out all of the valid types in the error message.

PR #462, opened 2018-06-21
Optimise bundle size by only importing used react icons

mrfelton:
The final size of renderer.prod.js is about 4.75mb. 2.35mb of this comes from react-icons.

Update the react icon imports to import only the specific icons that we use. This reduces the final size of renderer.prod.js by about 50%

Compare the results!

**Before:**
screenshot 2018-06-21 19 49 09

**After:**
screenshot 2018-06-21 19 49 18


Empact:
utACK 0c7e75e This is beautiful. Thanks! We should maybe also lint against this sort of thing.

Issue #463, opened 2018-06-21
Redundant block explorer api backends


Empact:
https://testnet-api.smartbit.com.au/ is currently down, which affects wallet functionality.

We should integrate with 2-3 explorer to provide redundancy so that the wallet does not go down when the apis do.

Empact:
Options include:
https://www.blocktrail.com/tBTC
https://www.blockcypher.com/dev/bitcoin/

Empact:
https://testnet-api.smartbit.com.au/ is down again. Another argument for prioritizing this.

PR #465, opened 2018-06-21
fix: Don't catch api/index errors

Empact:
If we catch these errors and resolve to the error itself, then `await` for that
function resolves to the error value rather than throws. Because no api currently
handles the associated errors, it is better to be noisy about these errors, by
throwing rather than returning the error object.


Empact:
Prompted by failures due to the block explorer being down #463

PR #466, opened 2018-06-21
Feature/testnet prefix channel balance

JimmyMow:
We had `BTC` hardcoded next to the channel balance in our Network tab. This just replaces it with the dynamic `currencyName` prop


JimmyMow:
I guess they're not properly setup. That's the second time my commit message has slipped through

Empact:
nit: You could also squash and force-push the branch when you make fix-ups.

Empact:
utACK 147a719

PR #471, opened 2018-06-22
Feature/minimize retype seed

JimmyMow:
We currently make our user retype all 24 words in their seed. This switches to the industry standard 3 random words, redesigns the form and also refactors the reducer and recovery form

### DIFF
--- a/app/components/Onboarding/Onboarding.js
+++ b/app/components/Onboarding/Onboarding.js
@@ -163,7 +163,7 @@ const Onboarding = ({
             description="Enter your cipherseed passphrase (or just submit if you don't have one)"
             back={() => changeStep(5)}
             next={() => {
-              const recoverySeed = recoverFormProps.seedInput.map(input => input.word)
+              const recoverySeed = recoverFormProps.recoverSeedInput.map(input => input.word)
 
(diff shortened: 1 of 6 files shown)


JimmyMow:
Really struggling with these commit hooks lol I have no idea how this turned out to be a 2 commit PR

JimmyMow:
new form:

screen shot 2018-06-22 at 2 34 21 pm
screen shot 2018-06-22 at 2 33 39 pm

Empact:
How about squashing these commits and adding a describing the change in the commit message? Right now both commits say they're fixing lints.

JimmyMow:
@Empact squashed 2 bad commits and updated with 1 descriptive one

JimmyMow:
@Empact rebased

Empact:
Found the issue, see 728706c. Does that look good, or do you want to take a different approach?

JimmyMow:
@Empact that looks good.

The line you removed was supposed to make sure there were sufficient entries but I think `reEnterSeedInput[index]` (first part of the conditional in the `every` func) does that as well.

👍

Empact:
Tested ACK 530784a

PR #473, opened 2018-06-22
Do not bundle moment locale data data that we don't use

mrfelton:
More work to reduce the size of our bundle - this helps speed up the initial load time of the app.

We aren't using moment locales, yet they are being bundled with our app and causing significant bloat.

Removing these shaves an additional 136.63k of our final renderer.prod.js bundle size, bringing it down from 1.12mb to just under 1mb.

More info:
 - https://github.com/jmblog/how-to-optimize-momentjs-with-webpack
 - https://medium.com/@michalozogan/how-to-split-moment-js-locales-to-chunks-with-webpack-de9e25caccea

**before:**
2-react-icons

**after:**
3-moment-locale

### DIFF
--- a/webpack.config.base.js
+++ b/webpack.config.base.js
@@ -3,6 +3,7 @@
  */
 
 import path from 'path'
+import { IgnorePlugin } from 'webpack'
 import { dependencies as externals } from './app/package.json'
 
 export default {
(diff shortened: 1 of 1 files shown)


Empact:
utACK b3e5f7c

PR #474, opened 2018-06-22
Remove lodash dependency

mrfelton:
We only use 3 functions from lodash - `find`, `filter` and `isEmpty` - all of which can be done easily with straight ES6.

This PR removes lodash.

Shaves a nominal 34.57k of final renderer.prod.js bundle size.

**before:**
3-moment-locale

**after:**
4-lodash

### DIFF
--- a/app/components/Contacts/ContactModal.js
+++ b/app/components/Contacts/ContactModal.js
@@ -1,6 +1,5 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import find from 'lodash/find'
 import ReactModal from 'react-modal'
 import FaCircle from 'react-icons/lib/fa/circle'
 import MdClose from 'react-icons/lib/md/close'
@@ -47,7 +46,7 @@ const ContactModal = ({
   }
 
   // the remote node for the channel
-  const node = find(channelNodes, { pub_key: channel.remote_pubkey })
+  const node = channelNodes.find(node => node.pub_key === channel.remote_pubkey)
 
   return (
     <ReactModal
--- a/app/lnd/methods/channelController.js
+++ b/app/lnd/methods/channelController.js
@@ -1,10 +1,9 @@
-import find from 'lodash/find'
 import { listPeers, connectPeer } from './peersController'
 import pushopenchannel from '../push/openchannel'
 
 function ensurePeerConnected(lnd, pubkey, host) {
   return listPeers(lnd).then(({ peers }) => {
-    const peer = find(peers, { pub_key: pubkey })
+    const peer = peers.find(candidatePeer => candidatePeer.pub_key === pubkey)
     if (peer) {
       return peer
     }
--- a/app/reducers/contactsform.js
+++ b/app/reducers/contactsform.js
@@ -1,6 +1,4 @@
 import { createSelector } from 'reselect'
-import filter from 'lodash/filter'
-import isEmpty from 'lodash/isEmpty'
 
 import { tickerSelectors } from './ticker'
 import { btc } from '../utils'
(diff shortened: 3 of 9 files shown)


JimmyMow:
🎊 utACK

Empact:
Needs rebase

mrfelton:
rebased

app/components/Contacts/ContactModal.js
```diff
@@ -47,7 +46,9 @@ const ContactModal = ({
   }
 
   // the remote node for the channel
-  const node = find(channelNodes, { pub_key: channel.remote_pubkey })
+  const node = channelNodes.find(channel => {
+    channel.remote_pubkey
+  })
```
Empact: The new code doesn't replicate the earlier behavior. Here's what would work AFAICT:
```js
const node = channelNodes.find(node => (
  node.pubkey == channel.remote_pubkey
))
```
Note `{}` functions require explicit return, `()` returns the value of the expression.

app/lnd/methods/channelController.js
```diff
@@ -1,10 +1,11 @@
-import find from 'lodash/find'
 import { listPeers, connectPeer } from './peersController'
 import pushopenchannel from '../push/openchannel'
 
 function ensurePeerConnected(lnd, pubkey, host) {
   return listPeers(lnd).then(({ peers }) => {
-    const peer = find(peers, { pub_key: pubkey })
+    const peer = peers.find(() => {
+      pubkey
+    })
```
Empact: Same:
```js
const peer = peers.find(candidatePeer => (
  candidatePeer.pub_key === pubkey
))
```

app/reducers/payform.js
```diff
@@ -267,9 +265,9 @@ payFormSelectors.payFormIsValid = createSelector(
 
     return {
       errors,
-      amountIsValid: isEmpty(errors.amount),
-      payInputIsValid: isEmpty(errors.payInput),
-      isValid: isEmpty(errors)
+      amountIsValid: !errors.amount,
+      payInputIsValid: !errors.payInput,
```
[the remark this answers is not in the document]
JimmyMow: Errors looks like:

```
{ 
amount: "Amount must be more than 0",
payInput: "Must be a valid BTC address or Lightning Network request"
}
```

so something like:
```
return {
      errors,
      amountIsValid: !errors.amount,
      payInputIsValid: !errors.payInput,
      isValid: !errors.amount && !errors.payInput
}
```

would work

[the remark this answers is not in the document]
Empact: Ok sounds like this is good - I like `errors.size === 0` because it won't need to be updated if there are future field additions.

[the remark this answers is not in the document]
JimmyMow: I think `size` is only an `underscore.js` function. Something like `Object.keys(errors).length` would be the functionality you're looking for I believe?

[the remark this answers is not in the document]
Empact: Ah yeah that would be better. 👍

app/reducers/contactsform.js
```diff
@@ -270,7 +267,7 @@ contactFormSelectors.manualFormIsValid = createSelector(manualSearchQuerySelecto
   }
   return {
     errors,
-    isValid: isEmpty(errors)
+    isValid: errors.length === 0
```
Empact: Errors is an object, has no length method. `Object.keys(errors).length === 0` would work.

mrfelton:
Cheers guys. I believe this should all be set right now.

app/components/Contacts/ContactModal.js
```diff
@@ -47,7 +46,7 @@ const ContactModal = ({
   }
 
   // the remote node for the channel
-  const node = find(channelNodes, { pub_key: channel.remote_pubkey })
+  const node = channelNodes.find(node => node.pubkey === channel.remote_pubkey)
```
Empact: You want `pub_key`

Empact:
@mrfelton just one typo

app/components/Contacts/ContactModal.js
mrfelton: doh! Fixed ;)

Empact:
utACK ba34367

PR #475, opened 2018-06-22
Feature/syncing

JimmyMow:
Our syncing page is currently plain. Just a progress bar. This PR introduces a new design along with a deposit address so users can fund their wallet while they wait for their node to sync.

screen shot 2018-06-22 at 4 16 56 pm


Empact:
Tested ACK 8ac7718

PR #476, opened 2018-06-22
Do not bundle moment-timezone that we don't use 

mrfelton:
We don't use `moment-timezone` at all. This can be replaced with `moment` for a smaller memory footprint.

This PR removes `moment-timezone`

Shaves another 176.88k of final renderer.prod.js bundle size.

**before:**
4-lodash

**after:**
5-moment-timezone


JimmyMow:
🎊 utACK

Empact:
utACK da2d695

PR #477, opened 2018-06-24
fix(p2pkh): generate np2wkh instead of p2pkh

JimmyMow:
We shouldn't be generating `p2pkh` addresses. We were when a new transaction was received. This switches to generating a `np2wkh` address


Empact:
utACK afb690d

PR #478, opened 2018-06-25
fix: kill lnd process on quit

### LINKED ISSUE #470: Javascript error in main process (Production)
With a production build: if you close and open the app too soon or sometimes at all we get weird `error in main process`. The error isn't always consistent but always has something to do with LND or gRPC. Seems as if we aren't properly shutting down the application and our LND process.

cc @Empact @mrfelton 
screen shot 2018-06-21 at 10 40 20 pm

mrfelton:
Ensure that any lnd process started by us is killed when the app quits.

Fix #470

### DIFF
--- a/app/main.dev.js
+++ b/app/main.dev.js
@@ -29,6 +29,8 @@ let didFinishLoad = false
 let startedSync = false
 let sentGrpcDisconnect = false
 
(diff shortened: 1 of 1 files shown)


Empact:
utACK bd26d26

PR #479, opened 2018-06-25
Refactor and modularize main process

mrfelton:
`main.dev.js` is getting a little unwieldy. It can be hard to follow and understand the code flow. It houses code to handle quite a few distinct things.

In this PR I've done a fairly substantial refactor of `main.dev.js`. The main changes are:

## 1. Extract code for starting and monitoring the lnd process to a standalone module

A new `Neutrino` class has been created with 2 public methods. `start` and `stop`. `start` will launch a new `lnd` process whilst `stop` will stop it. This class extends the `EventEmitter` class and emits the following events based on activity detected from the lnd log output.

 - `grpc-proxy-started` - gRPC started
 - `wallet-opened` Wallet opened
 - `fully-synced` - lnd is all caught up to the blockchain
 - `got-block-height` - got updated block height

See https://github.com/mrfelton/zap-desktop/blob/refactor/main-process/app/lnd/lib/neutrino.js

## 2. Extract the code that deals with communication between the main and renderer processes to a standalone module:

A new `ZapController` class  has been created which houses all of the logic for intraprocess communication between the main and renderer processes.

See https://github.com/mrfelton/zap-desktop/blob/refactor/main-process/app/zap.js

## 3. Optimise the application loading flow so that it starts in the quickest possible time.

Previously we had several `setInterval` loops that were checking every second to see if the application status has changed and trigging the appropriate action if so. This was pretty hard to follow has been replaced here with more extensive use of promises. This enables us to act instantly to relevant changes rather than waiting up to 1 second for the next interval to fire.

The only stuff that lives in `main.dev.js` now is the top level `app` listeners, which calls out the other parts mentioned above to bootstrap the application.

See https://github.com/mrfelton/zap-desktop/blob/refactor/main-process/app/main.dev.js

The end result is some slightly more modular and easy to follow code, which should also be more easy to write tests against.


JimmyMow:
slow clap 👏 👏 👏 

utACK 9497863

both in `dev` and I built prod and have been running it. can confirm this also fixes #470

app/lnd/index.js
```diff
@@ -1,33 +1,44 @@
...
-  const lndMethods = (event, msg, data) => methods(lnd, mainLog, event, msg, data)
-
-  callback(lndSubscribe, lndMethods)
+    return {
+      lndSubscribe,
+      lndMethods
+    }
+  } catch (e) {
+    throw e
+  }
```
Empact: What's the purpose of this catch block?

app/lnd/index.js
```diff
@@ -1,33 +1,44 @@
...
 
-  const walletUnlockerObj = walletUnlocker(lndConfig.lightningRpc, lndConfig.lightningHost)
-  const walletUnlockerMethodsCallback = (event, msg, data) =>
-    walletUnlockerMethods(walletUnlockerObj, mainLog, event, msg, data)
-
-  callback(walletUnlockerMethodsCallback)
+    return walletUnlockerMethodsCallback
+  } catch (e) {
+    throw e
+  }
```
Empact: Same question

app/main.dev.js
```diff
@@ -8,348 +6,112 @@
...
+  /**
+   * Add application event listener:
+   *  - Open zap payment form when lightning url is opened
+   */
+  app.setAsDefaultProtocolClient('lightning')
+  app.on('open-url', (event, url) => {
+    mainLog.debug('open-url')
+    event.preventDefault()
+    const payreq = url.split(':')[1]
+    zap.sedMessage('lightningPaymentUri', { payreq })
```
JimmyMow: @mrfelton should be `sendMessage` here

Empact:
Now that #478 is merged, it can be dropped from this PR.
Could you also split out cf1d1291ac3 as a separate PR?
Finally, it would be helpful to split up 94978639064d038f87f3 into a few smaller commits, given it's a substantial refactor of core logic, without an easy way of reviewing en-masse.

app/lnd/index.js
mrfelton: This was a relic from earlier testing of the PR. I have since moved the error handling of these methods to the places where they are called - where we now show an error message to the user and abort the app if there is a problem.

See https://github.com/mrfelton/zap-desktop/blob/6d5163ade40805688f0fac53cc6c99eb336e8242/app/zap.js#L95-L113

This can happen if for example, you already have an lnd process running on your system which is not fully functional (eg missing tls cert in the expected location)

I've removed these unneeded try/catch blocks now.

app/main.dev.js
mrfelton: thanks - fixed

mrfelton:
@Empact I've created a new PR for the additional performance related logging here https://github.com/LN-Zap/zap-desktop/pull/488. That commit is still in this PR but can be removed once that has been merged.

Regarding this PR itself - given the nature of the PR and the fact that it touches quite a few things, I haven't been able to see a way to break it into multiple smaller PR's in such a way that doesn't create an inordinate amount of additional work. Many of the changes are somewhat interconnected and dependent on each other. If I were to extract out just one part of the change set into a standalone PR it would mean changing the way the code is integrated purely as an intermediate step, only to then have to create another PR that again changes the integration code.

For example, one logical intermediate changeset would be the creation of the `Neutrino` class and integration of that into `main.dev.js`. However to do this outside of the context the wider changeset would mean reworking the integration code so that it works with how main.dev.js works today, only to then rework it all again in the next intermediate changeset as more things are updated and moved around. I think the amount of work involved in doing that is disproportional to the gains.

The easiest way to review this PR is to review the relevant files individually:

- [app/lnd/index.js](https://github.com/mrfelton/zap-desktop/blob/6d5163ade40805688f0fac53cc6c99eb336e8242/app/lnd/index.js)
- [app/lnd/lib/neutrino.js](https://github.com/mrfelton/zap-desktop/blob/6d5163ade40805688f0fac53cc6c99eb336e8242/app/lnd/lib/neutrino.js)
- [app/lnd/lib/util.js](https://github.com/mrfelton/zap-desktop/blob/6d5163ade40805688f0fac53cc6c99eb336e8242/app/lnd/lib/util.js)
- [app/main.dev.js](https://github.com/mrfelton/zap-desktop/blob/6d5163ade40805688f0fac53cc6c99eb336e8242/app/main.dev.js)
- [app/zap.js](https://github.com/mrfelton/zap-desktop/blob/6d5163ade40805688f0fac53cc6c99eb336e8242/app/zap.js)

Empact:
@mrfelton Fair enough and I don't want to create unnecessary work. The only clean extraction I saw was `isLndRunning`, but it's a good practice to keep in mind.

app/main.dev.js
```diff
@@ -1,368 +1,116 @@
...
+// Determine wether we should start our own lnd process or connect to one that is already running.
+const zapMode = lnd
+  .isLndRunning()
+  .then(res => {
+    mainLog.debug('lnd already running: %s', res)
+    mainLog.timeEnd('Time until lnd process lookup finished')
+    return res ? 'external' : 'internal'
   })
-}
+  .catch(mainLog.error)
```
Empact: Should we explicitly return internal here?

app/zap.js
```diff
@@ -0,0 +1,217 @@
...
+        walletUnlockerMethods(event, msg, data)
+      })
+
+      this.sendMessage('walletUnlockerStarted')
+    } catch (error) {
+      dialog.showMessageBox({
+        type: 'error',
+        message: `Unable to start lnd wallet unlocker. Please check your lnd node and try again: ${error}`
+      })
+      app.quit()
```
Empact: I think this pattern of dialog + quit results in the dialog disappearing shortly after revealed.

app/main.dev.js
Empact: Nevermind, I see that the default handling below goes to internal.

Empact:
Needs rebase

app/zap.js
mrfelton: In my environment the dialog blocks the application until you hit ok, after which the application quits.

mrfelton:
rebased

PR #480, opened 2018-06-25
fix(husky hooks): drop opt --in from husky hooks

Empact:
With opt --in present, the hooks are not run, without it they are.

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,6 +1,6 @@
 {
   "hooks": {
-    "commit-msg": "opt --in commit-msg --exec 'commitlint -E HUSKY_GIT_PARAMS'",
-    "pre-commit": "opt --in pre-commit --exec 'lint-staged'"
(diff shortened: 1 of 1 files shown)


Empact:
@mrfelton I assume these were working for you with opt --in, can you point me to what might be wrong with my system if we should be keeping it?

mrfelton:
Did you add the .opt-in file? See git hooks section of https://github.com/LN-Zap/zap-desktop/blob/master/CONTRIBUTING.md

Empact:
Ah no I did not - how do you feel about always having these enabled? It would free us from needing to police formatting or commit messages on incoming PRs.

mrfelton:
Personally, I would prefer that it was always on. I don't think we should ever commit things that don’t follow the standards.

I made it opt in because it was quite a change to the previous workflow and I didn’t want to be too imposing from the onset. If you guys are happy to integrate the flow by default then I’m all for it.

Empact:
Closing in favor of #487

PR #481, opened 2018-06-25
enhance(mainnet conf): add an lnd.conf entry for bitcoin.mainnet

Empact:
Commented out as testnet is still our default.


JimmyMow:
👏 ayyy

utACK

PR #482, opened 2018-06-25
fix(channel-list): change the logic to look at the currentChannels

JimmyMow:
Before our conditionals in `Network.js` were looking at `channels` which does not account for `pendingChannels`. So the bug was that a user with no channels would open their first one and not see it pending. This fixes our conditionals by looking at the `currentChannels` list

### DIFF
--- a/app/components/Contacts/Network.js
+++ b/app/components/Contacts/Network.js
@@ -32,8 +32,7 @@ class Network extends Component {
         filter,
         selectedChannel,
         loadingChannelPubkeys,
-        closingChannelIds,
-        channels
+        closingChannelIds
       },
       currentChannels,
       balance,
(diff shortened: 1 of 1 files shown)


Empact:
utACK 1c9a23b

Empact:
@JimmyMow Should we rename `channels.channels`?

JimmyMow:
@Empact eh, I think it makes sense. I am simply following the `lncli` language. `channels.channels` is `lncli listchannels`, etc.

Maybe `currentChannels` needs to be renamed. `currentChannels` is a selector that takes `channels`, `pending` and `closing` and returns them based on the current search and/or filter. 

The naming makes sense to me personally but open to discussing if others feel it's not descriptive enough

Empact:
Cool - I'm fine with leaving as-is. We can adjust if it causes more issues down the road.

PR #485, opened 2018-06-25
fix(waiting_close_channels): account for waiting_close_channels

JimmyMow:
`waiting_close_channels` is a fairly new property of the `pendingChannels` object returned by `LND`. any channels that were returned in this list were:

1) displayed as `offline`
2) not included in the `Closing` filter in the channels list

This PR just accounts for `waiting_close_channels` and fixes both issues above


Empact:
utACK 5c30107

PR #486, opened 2018-06-25
fix(channel-details): use currencyName

JimmyMow:
We weren't using `currencyName` in the channel details. This fixes that


JimmyMow:
screen shot 2018-06-25 at 6 28 07 pm

Empact:
utACK f84fb97

PR #488, opened 2018-06-26
Add logging to show boot time

mrfelton:
Add some logging so that we can see how long it takes from starting up the app until the user is first able to interact with it.


Empact:
Tested ACK ce201eb

PR #489, opened 2018-06-26
fix(grpc): handle gRPC startup issues

mrfelton:
If there is a problem initializing the lnd gRPC connection, show an error message to the user and quit the app.

This is an extraction of some of the changes from https://github.com/LN-Zap/zap-desktop/pull/479


Empact:
utACK 72691f4

PR #494, opened 2018-06-26
Update concurrently to v3.6.0

mrfelton:
Concurrently v3.6.0 includes a number of improvements and fixes. Of particular interest is the new npm command shortcut feature which we can use to shorten our package scripts a little more.

See https://github.com/kimmobrunfeldt/concurrently/releases/tag/3.6.0


Empact:
utACK 2b82edb

PR #491, opened 2018-06-26
fix(devtools): patch webrequest to fix devtools

mrfelton:
Fix an incompatibility between electron 2.x and devtools by patching the electron webRequest handler.

See https://github.com/electron/electron/issues/13008#issuecomment-400261941

Fix #490

PR #495, opened 2018-06-26
Fix devtools electron compatibility issues

mrfelton:
It looks like there was a typo in the fix for https://github.com/LN-Zap/zap-desktop/pull/491. This PR corrects it.


Empact:
Ha well I can fault myself for not calling out that `hack`. I saw it but thought it was some kind of url flag you were passing to optionally disable the warning.

utACK c022937

PR #496, opened 2018-06-26
Update electron and electron builder to latest versions

mrfelton:
Electron v2.0.3 was released a few days ago which fixes a number of minor bugs which are likely to affect us. There is also one fix which is marked as a security fix

See https://github.com/electron/electron/releases/tag/v2.0.3

Also, update electron-builder which has had quite a few nice updates since the version we had installed.

Have tested with `npm run dev`, `npm start`, and via a package built with `npm run package` and everything appears to work as expected.


Empact:
utACK cea699b

PR #492, opened 2018-06-26
Remove unused dependencies

mrfelton:
Remove the following project dependencies, which are not being used:

- express
- rimraf
- sinon
- bitcore-lib
- d3-force
- d3-selection
- d3-zoom
- xtend

PR #497, opened 2018-06-26
Add missing rimraf dependency

mrfelton:
It looks like we were a little hasty in #492 as rimraf is used directly within package.json in the `flow-typed` script and so it needs to remain as a dev dependency.

See https://github.com/LN-Zap/zap-desktop/pull/492


JimmyMow:
utACK 1856b86

PR #501, opened 2018-06-27
feat(grpc): lnd connection error recovery

mrfelton:
If you try to connect to a custom LND instance with some credentials that are incorrect, we currently show a native error dialog and then quit the app.

This PR improves this so that we instead show the relevant errors to the user and allow the user to correct and try again.

eg:
screenshot 2018-06-27 11 39 08

screenshot 2018-06-27 11 39 23

screenshot 2018-06-27 11 39 36

### DIFF
--- a/app/lnd/lib/lightning.js
+++ b/app/lnd/lib/lightning.js
@@ -1,6 +1,6 @@
-import fs from 'fs'
 import grpc from 'grpc'
 import config from '../config'
+import { getDeadline, validateHost, createSslCreds, createMacaroonCreds } from './util'
 
 // Default is ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384
 // https://github.com/grpc/grpc/blob/master/doc/environment_variables.md
@@ -19,22 +19,39 @@ process.env.GRPC_SSL_CIPHER_SUITES =
     'ECDHE-ECDSA-CHACHA20-POLY1305'
   ].join(':')
 
-const lightning = (rpcpath, host) => {
+/**
+ * Creates an LND grpc client lightning service.
+ * @returns {rpc.lnrpc.Lightning}
+ */
+const lightning = async () => {
   const lndConfig = config.lnd()
-  const lndCert = fs.readFileSync(lndConfig.cert)
-  const sslCreds = grpc.credentials.createSsl(lndCert)
-  const rpc = grpc.load(lndConfig.lightningRpc)
+  const { host, rpcProtoPath, cert, macaroon } = lndConfig
 
-  const metadata = new grpc.Metadata()
-  const macaroonHex = fs.readFileSync(lndConfig.macaroon).toString('hex')
-  metadata.add('macaroon', macaroonHex)
+  // Verify that the host is valid before creating a gRPC client that is connected to it.
+  return await validateHost(host).then(async () => {
+    // Load the gRPC proto file.
+    const rpc = grpc.load(rpcProtoPath)
 
-  const macaroonCreds = grpc.credentials.createFromMetadataGenerator((params, callback) =>
-    callback(null, metadata)
-  )
-  const credentials = grpc.credentials.combineChannelCredentials(sslCreds, macaroonCreds)
+    // Create ssl and macaroon credentials to use with the gRPC client.
+    const [sslCreds, macaroonCreds] = await Promise.all([
+      createSslCreds(cert),
+      createMacaroonCreds(macaroon)
+    ])
+    const credentials = grpc.credentials.combineChannelCredentials(sslCreds, macaroonCreds)
 
-  return new rpc.lnrpc.Lightning(host, credentials)
+    // Create a new gRPC client instance.
+    const lnd = new rpc.lnrpc.Lightning(host, credentials)
+
+    // Call the getInfo method to ensure that we can make successful calls to the gRPC interface.
+    return new Promise((resolve, reject) => {
+      lnd.getInfo({}, { deadline: getDeadline(2) }, err => {
+        if (err) {
+          return reject(err)
+        }
+        return resolve(lnd)
+      })
+    })
+  })
 }
 
 export default lightning
--- a/app/lnd/lib/util.js
+++ b/app/lnd/lib/util.js
@@ -1,11 +1,100 @@
+import dns from 'dns'
+import fs from 'fs'
+import { promisify } from 'util'
 import { lookup } from 'ps-node'
+import grpc from 'grpc'
+import isIP from 'validator/lib/isIP'
+import isPort from 'validator/lib/isPort'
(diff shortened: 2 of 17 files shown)


app/lnd/lib/lightning.js
```diff
@@ -19,22 +19,64 @@ process.env.GRPC_SSL_CIPHER_SUITES =
...
-  const lndCert = fs.readFileSync(lndConfig.cert)
-  const sslCreds = grpc.credentials.createSsl(lndCert)
+
+  const host = lndConfig.lightningHost
   const rpc = grpc.load(lndConfig.lightningRpc)
 
-  const metadata = new grpc.Metadata()
-  const macaroonHex = fs.readFileSync(lndConfig.macaroon).toString('hex')
-  metadata.add('macaroon', macaroonHex)
+  const [sslCreds, macaroonCreds] = await Promise.all([
```
JimmyMow: nice!

JimmyMow:
Tested ACK 37d97b7

Not just a feature improvement. Lots of improvements in refactoring as well. Nice @mrfelton!

app/lnd/lib/util.js
```diff
@@ -0,0 +1,24 @@
...
+  validateFileAccess,
+  readFile: fsReadFile
+}
+
+/**
+ * Validates that the file at the specified path can be read
+ * @param {String} path
+ * @returns {Promise<Boolean>}
+ */
+async function validateFileAccess(path) {
```
Empact: Why `async`? I don't think promise gives us anything here.

app/lnd/lib/lightning.js
```diff
@@ -19,22 +19,64 @@ process.env.GRPC_SSL_CIPHER_SUITES =
...
+ */
+const createSslCreds = async certPath => {
+  if (!(await validateFileAccess(certPath))) {
+    const error = new Error('SSL cert path could not be accessed')
+    error.code = 'LND_GRPC_SSL_CERT_NOT_ACCESSIBLE'
+    throw error
+  }
+
+  let lndCert = await readFile(certPath)
+  let sslCreds = grpc.credentials.createSsl(lndCert)
```
Empact: nit: these could be `const`

app/lnd/lib/lightning.js
```diff
@@ -19,22 +19,64 @@ process.env.GRPC_SSL_CIPHER_SUITES =
...
   return new rpc.lnrpc.Lightning(host, credentials)
 }
 
+/**
+ * Validates and creates the ssl channel credentials from the specified file path
+ * @param {String} certPath
+ * @returns {grpc.ChanelCredentials}
+ */
+const createSslCreds = async certPath => {
+  if (!(await validateFileAccess(certPath))) {
```
Empact: According to https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback

> Using fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

app/main.dev.js
```diff
@@ -361,7 +349,28 @@ app.on('ready', async () => {
       startLnd(options.alias, options.autopilot)
     } else {
       mainLog.info('CONNECTING TO CUSTOM LND INSTANCE')
-      startGrpc()
+      try {
+        await startGrpc()
```
Empact: nit: Given you're catching the errors, it would be cleaner to use `.then`/`.catch` rather than immediately catch a `throw`.

app/lnd/lib/lightning.js
mrfelton: Makes sense. I can rework this.

app/lnd/lib/util.js
mrfelton: > Why async?

Well, the method is asynchronous, so it either needs to operate with a callback, a promise, or async. It's probably a case of personal preference but now that async is available natively in node I prefer to use this over promises or callbacks as I just find the code to be cleaner, simpler, and more readable without `.then` and `.catch` all over the place and instead just use good old `try/catch` syntax to achieve the same thing

app/main.dev.js
mrfelton: I think thats pretty subjective. In this codebase we are currently very inconsistent with regards to wether we use callbacks, promises, or async. IMO async generally allows for a cleaner and more universally standard syntax. But again, I think that is a little subjective.

I don't have too string an opinion either way so if the consensus is to stick with promises I can update it. Feels like a bit of a shame not to leverage nodes latest and greatest features though!

mrfelton:
@Empact After playing around with some different options for the error handling, I've pushed a revised version that I think looks pretty decent - using  `await` in conjunction with `catch` for the error handling.

There are some nice examples of some different possible approaches here: https://medium.freecodecamp.org/here-are-examples-of-everything-new-in-ecmascript-2016-2017-and-2018-d52fa3b5a70e

JimmyMow:
@mrfelton @Empact I just tested 74e4ce3

Everything works great except if I pass valid `cert` and `macaroon` paths but an bad `host`. I passed valid macaroon and cert paths but entered 123456789 for the host. Hitting `next` after those form submissions crashed the app. After doing a little bit of debugging the issue is `new rpc.lnrpc.Lightning(host, credentials)` doesn't complain if you hand it a bunch of random numbers as the host. It returns the object and when we try and call `getInfo` on it the application crashes immediately. Wondering how we can verify the `host` or `catch` an error somehow given a bad `host` trying to generate the `rpc.lnrpc.Lightning` object

Empact:
Something like: https://validatejs.org/#validators-url ?

mrfelton:
Interesting. After a little more testing I have found that as long as your hostname starts with a letter and not a number then you get a `Name resolution failure` instantly. If you start the hostname with a number then you either get a `Connect Failed` error but only after about 20 seconds of trying to connect with the lightning bold screen showing whilst you wait, or the app crashes

mrfelton:
Ok so this is a pretty interesting issue. The crux of the issue is https://github.com/grpc/grpc/issues/13163.

In addition to that, I have found that node-grpc behaves differently depending on wether you are setting the host as an IP address, a string, or a number. In some cases it crashes outright. In others it will eventually reject with an error.

In this updated PR I have set a deadline on the grpc connection, which mitigates part of the issue - ensuring that grpc times out quickly rather than consuming all available resources and crashing the application in cases where it tries to connect to an unreachable host.

In addition to that, I have added more checks to ensure that the host that we pass to the grpc client is valid. This includes:

 - if the user enters something that looks like an IP address, validate that it is a valid IP address
 - If the user enters a hostname, validate that it resolves using `dns.lookup`
 - if the user includes a port, validate that it is a valid port number
 - trim user supplied strings so that we have clean strings to work with

Empact:
Tested ACK dd134aa

PR #502, opened 2018-06-27
Enable electron security checks on travis/appveyor

mrfelton:
Unset the `ELECTRON_DISABLE_SECURITY_WARNINGS` environment variable which was being used to disable the security checks when testing the app on ci. This is no longer needed now that we have a more strict Content Security Policy when the app runs in production mode.

See https://github.com/LN-Zap/zap-desktop/pull/456

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ language: node_js
 env:
   global:
     - DEBUG=electron-builder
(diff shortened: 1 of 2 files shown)


JimmyMow:
utACK 9cd7a79

Empact:
utACK 9cd7a79

PR #505, opened 2018-06-28
fix(onboarding-empty-pass): add empty password validation

odb366:
Concerning issue #415. Now users can't proceed when password is empty.


Empact:
utACK a712381

PR #506, opened 2018-06-28
fix(onboarding-password-bug): add more password validation

odb366:
This PR fixes:

- It was possible to proceed with:
 Password: `1234`
 Confirm Password: `12`

- It was also possible to proceed with filled in Password but Confirm Password empty.

I hope that's the last PR related to #415 :)


Empact:
Tested ACK 31a8084

PR #508, opened 2018-06-29
Adding Windows 10 info

elkimek:
Git for Windows is needed to download and run Zap from master.


odb366:
@elkimek I think people reading in the advanced section should know that they need git.

Empact:
Yeah I'm NACK on this. There are also several ways to run git on a windows machine, including https://git-scm.com/downloads, so plugging gitforwindows.com seems unwarranted.

PR #509, opened 2018-06-29
Refactor file structure

mrfelton:
I have been finding that the file structure of the project has become a little confusing in places. This PR moves all of the application level styles to a new `app/styles` directory and consolidates the supporting code in `app/lib`.

This screenshot highlights the changes

zap-file-structure

EDIT: breaking this up into 3 separate PRs:

- [ ] https://github.com/LN-Zap/zap-desktop/pull/510
- [ ] https://github.com/LN-Zap/zap-desktop/pull/511
- [ ] https://github.com/LN-Zap/zap-desktop/pull/512


JimmyMow:
Love it. Concept ACK.

Will pull it down and give it a spin a bit later

Empact:
Concept ACK. `lib/utils` sounds a bit redundant to me. How about moving those to `lib`? E.g. `lib/log`, `lib/bech32`.

mrfelton:
Personally, I find it cleaner that everything is categorised to at least one level within the `lib` dir. Feels more consistent that everything lives within a primary category than if some do and some don't. If everything has to live within a subdirectory of `lib` as a rule then there is less chance of a "should it or shouldn't it" question that can lead to subjective discussions about what should live where.

mrfelton:
... for a smaller app then maybe letting some things dangle in the top level `lib` dir would be ok, but as time goes on the size and scope of this app will only continue to increase so it'd be better to keep things as organised as possible from the onset.

Empact:
Could you split this into 3 PRs? The commits are independent and splitting makes it easy to merge the things which are easily agreed upon.

mrfelton:
sure thing:

 - https://github.com/LN-Zap/zap-desktop/pull/510
 - https://github.com/LN-Zap/zap-desktop/pull/511
 - https://github.com/LN-Zap/zap-desktop/pull/512

Empact:
Thanks!

Empact:
Closing in favor of #510, #511, #512

PR #511, opened 2018-06-29
refactor: move styles to styles directory

mrfelton:
Spin off from https://github.com/LN-Zap/zap-desktop/pull/509


Empact:
Could you also update the lint-styles paths in package.json?

mrfelton:
I've extracted those path definitions out into an npm config variable and updated the scripts to use that.

mrfelton:
I have actually noticed that none of the stylesheets in app/components are being linted. I will fix in a follow up PR as after fixing, many of the stylesheets will need to be re-linted and fixed where they are broken

mrfelton:
rebased

Empact:
utACK bca5761

PR #514, opened 2018-06-29
Ensure all component stylesheets are properly linted

### LINKED ISSUE #513: Stylesheets under app/components do not pass lint rules.
There is a bug in our configuration for stylesheet linting causing none of the the stylesheets in app/components to be linted.

We lint using the glob `app/components/*.scss` It should be `app/components/**/*.scss` in order to get things in all sub directories.

Many of our stylesheets will need relinting and fixing after fixing the lint script config

mrfelton:
Update glob pattern in stylesheet lint scripts to ensure all stylesheets are linted.

Fix #513


mrfelton:
After fixing the lint script. Even after running the `lint-styles-fix` script there are a lot of issues that will need to be manually fixed:

The list of issues is pretty long...

screenshot 2018-06-29 21 39 08

mrfelton:
All stylesheets are now passing the lint tests. I have disabled 2 lint rules in the process.

1. `no-descending-specificity` - this rule creates a lot of errors which would require fairly substantial reworking of some of our stylesheets in order to resolve. This would create a large and hard to review PR. This rule also reports some false positives and trying to follow this rule can result in much less DRY code. From what I can see of reading around the web a little, a lot of people are disabling this rule for similar reasons. There are open bug reports about the issues against the stylelint project.
2. `no-empty-source` - the current config of our build process expects there to be an scss file to go with each component. If we remove the empty files (there are only 2 of them) then the build process breaks.

Empact:
utACK f3bf899

PR #515, opened 2018-06-29
Update docs to suggest building before running tests

asoltys:
I tried running "npm run test" as suggested at the end of these instructions and got a bunch of errors saying "The main process is not built yet. Build it by running "npm run build-main" so just making a note of that in the docs


Empact:
utACK 36b1481

PR #516, opened 2018-06-29
Fix anchor links in docs

asoltys:
The anchor links on the Advanced Usage page weren't working in Chrome on Ubuntu. Making them lowercase and removing the parentheses seems to solve the problem.


Empact:
Tested ACK 0e003a0 https://github.com/asoltys/zap-desktop/blob/master/ADVANCED.md

PR #526, opened 2018-07-02
fix: ensure onboarding completes after syncing

mrfelton:
Fix a race condition that could cause the app to think that it is still syncing even after it has completed and as a result being stuck reporting 100% syncing during the onboarding process.

`startGrpc` can take some time to fully establish and verify that the connection is set up as we wait for a successful result from a call to `getInfo`.

Previously, we would send the `lndSyncing` message to the app only after `startGrpc` was complete. However, if the syncing process had actually completed before `startGrpc` then we would first notify that syncing had completed, only to then notify that it had started after it had already completed.

In this fix, we notify the app of `lndSyncing` immediately after the wallet has been unlocked.

### DIFF
--- a/app/zap.js
+++ b/app/zap.js
@@ -45,7 +45,7 @@ class ZapController {
     this._registerIpcListeners()
 
     // Show the window as soon as the application has finished loading.
-    this.mainWindow.webContents.on('did-finish-load', async () => {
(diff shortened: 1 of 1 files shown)


michaelWuensch:
Tested ACK e06fc94

Empact:
utACK e06fc94

PR #527, opened 2018-07-02
Remember last unused wallet address

### LINKED ISSUE #519: The funding address displayed during syncing changes every time I restart zap, although I never funded it
> Shouldn't the address stay the same until it is actually used?
 -- @michaelWuensch

mrfelton:
Only generate a new wallet address after the last one has been used.

Fix #519

### DIFF
--- a/app/components/Onboarding/Syncing.js
+++ b/app/components/Onboarding/Syncing.js
@@ -9,21 +9,24 @@ import styles from './Syncing.scss'
 
 class Syncing extends Component {
   componentWillMount() {
-    const { fetchBlockHeight, blockHeight, newAddress } = this.props
+    const { fetchBlockHeight, blockHeight } = this.props
 
     // If we don't already know the target block height, fetch it now.
     if (!blockHeight) {
       fetchBlockHeight()
     }
-
-    newAddress('np2wkh')
   }
 
   render() {
-    const { syncPercentage, address, blockHeight, lndBlockHeight } = this.props
+    const { hasSynced, syncPercentage, address, blockHeight, lndBlockHeight } = this.props
+
     const copyClicked = () => {
       copy(address)
       showNotification('Noice', 'Successfully copied to clipboard')
+    }
+
+    if (typeof hasSynced === 'undefined') {
+      return null
     }
 
     return (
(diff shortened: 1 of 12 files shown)


michaelWuensch:
Tested ACK 2de1686

app/reducers/address.js
```diff
@@ -36,15 +38,46 @@ export function closeWalletModal() {
   }
 }
 
+// Get our existing address if there is one, otherwise generate a new one.
+export const walletAddress = type => async (dispatch, getState) => {
```
Empact: No need for `async`

mrfelton:
I've removed the unneeded `async` definition @Empact .

I think it is worth noting that with this approach it is possible that a user could be presented with an address that they have in fact already used.

The only way that we know wether an address has already been used is by checking the past transactions and comparing with the currently stored wallet address. We generate and store locally a new wallet address only in the following circumstances:

1. the first time the wallet has been opened
2. after a new onchain transaction has been received
3. after pulling the list of all past transactions from lnd and finding one that matches the one we have stored.

If you open up an existing wallet after having not used it for some time and there has been an on-chain transaction to your wallet address since you last synchronised it then we will show the used address until we have synced the data up until the point of your new transaction, at which point we detect it has already been used and generate a fresh one.

This could mean that is possible for the syncing screen to show an already used wallet address and then update it to show a new one mid-way through the syncing process. I'm not sure there is a way round this or wether it is a scenario that is likely to be encountered.

JimmyMow:
@mrfelton A little hack around is to only show the deposit address UI on the sync screen if the user has a wallet balance and/or channel balance of 0. 

The only real user story supporting the deposit UI there is a brand new user syncing from scratch. Greatly improves overall UX if they can deposit funds while syncing. Especially if they enable `autopilot`. Then it's -> create wallet -> deposit while sync -> once fully synced channels start to open via autopilot

Once you've used Zap before and have on-chain funds the sync screen deposit address is _rarely_ helpful. 

So my suggestion is we have a dynamic sync screen design based on `walletbalance` and/or `channelbalance`.

The only small edge case here that could present reused address is a user that synced Zap, didn't deposit funds, closes Zap, then someone sends to that on-chain address while their node isn't online, then they launch Zap again. I think that's an acceptable edge case for the beta stage we are in. I'll continue to think about solutions to that

app/components/Onboarding/Syncing.js
```diff
@@ -9,10 +9,9 @@ import styles from './Syncing.scss'
 
 class Syncing extends Component {
   componentWillMount() {
-    const { fetchBlockHeight, newAddress } = this.props
+    const { fetchBlockHeight } = this.props
 
     fetchBlockHeight()
-    newAddress('np2wkh')
```
JimmyMow: For brand new nodes this PR doesn't fetch an address and I get a forever spinner while syncing. This is because `fetchInfo` (which then calls `walletAddress`) isn't called until the node is synced. The PR relies on there already being an address created. 

We should call `fetchInfo` here instead so that we go grab the `pubKey` and generate an address.

JimmyMow: Ah that actually won't work as it can get called before `gRPC` is connected. Let me think some more

JimmyMow: @mrfelton removing this is great. on current master this is called (`componentWillMount`) _before_ gRPC is connected so the call to get a new address hangs and we never get one. 

Instead of calling for a new address on the `componentWillMount` callback I suggest we `dispatch(fetchInfo())` (which also gets a new address according to this PR) in the `grpcConnected` callback here: https://github.com/LN-Zap/zap-desktop/blob/master/app/reducers/lnd.js#L46

JimmyMow: I patched the above suggestion locally and it works great ^

mrfelton:
@jackmallers the change that you suggested re calling `dispatch(fetchInfo())` in the `grpcConnected` callback has actually been made in https://github.com/LN-Zap/zap-desktop/pull/529 already.

I have updated this PR to also store a boolean named `hasSynced` in the wallets electron-store which get's set after the first time the wallet has fully synced and I use this to determine wether or not to show the funding address on the syncing page.

As this PR now depends on the changes from https://github.com/LN-Zap/zap-desktop/pull/529 I have rebased it from that. Once that PR gets merged, this can one can be more easily reviewed and merged.

mrfelton:
You can review the changes for this PR only without seeing the rest of the changes from #529 here https://github.com/mrfelton/zap-desktop/compare/fix/sync-issues...mrfelton:feat/remember-wallet-address

JimmyMow:
Cool. I'll go test #529 again

mrfelton:
rebased

JimmyMow:
Tested ACK a989c03

Empact:
utACk a989c03

PR #529, opened 2018-07-04
Improve feedback to users during sync process

mrfelton:
This patch fixes some issues with the sync process. The parsing of lnd log entries where we calculate the current sync progress has been moved from the renderer back to the main process and updated to improve it's accuracy.

We now call dispatch a call to `fetchInfo` as soon as the gRPC connection has been established so that we get the node info as early as possible, and we defer showing the syncing screen until we are sure that a grpc connection has been established and syncing is in progress.

These enable us to provide better feedback to the user and prevent further race conditions that I have observed.

As a bonus, we show the user some more information about the current sync progress.

screenshot 2018-07-04 02 15 12

### DIFF
--- a/app/components/Onboarding/Syncing.js
+++ b/app/components/Onboarding/Syncing.js
@@ -16,8 +16,7 @@ class Syncing extends Component {
   }
 
   render() {
-    const { syncPercentage, address } = this.props
-
+    const { syncPercentage, address, blockHeight, lndBlockHeight } = this.props
     const copyClicked = () => {
       copy(address)
       showNotification('Noice', 'Successfully copied to clipboard')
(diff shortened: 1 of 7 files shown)


JimmyMow:
Tested ACK `ce0bc3d`

michaelWuensch:
Haven't tested it yet, though I would suggest to change
"... syncing block 1,060,000 of 1,349,469"
to an actual time users can refer to:
"5 years 26 Weeks behind" or "2,010 days behind"
I have the feeling showing the block height is to technical for ordinary people. I think If they see, ah, I have to catch up what happended during the last 20 days it gives them more useful feedback than I am at "1,323,212 of 1,349,469"

mrfelton:
@michaelWuensch I've just pushed several updates for this which I think improves the end user experience a little

JimmyMow:
@mrfelton wanna make the change to a new blockexplorer in the PR? `testnet-api.smartbit.com.au` is down again, which destroys the UX here. We shouldn't be so reliant on an API endpoint we don't control but a short term fix is switching to a more stable one. Maybe `blockcypher`?

mrfelton:
Yeah, thats annoying @jackmallers . I've added a second commit to this which adds some redundancy there. We now check 3 different block explorers

mrfelton:
I've pulled the changes for the block explorer fallback into a separate PR https://github.com/LN-Zap/zap-desktop/pull/535.

Additionally, I have fine tuned the lnd log parsing somewhat in order to improve efficiency. We can actually find out the current block height directly from the lnd logs so there is no need to call out to external block explorers at all if we have already determined the current block height from the logs.

JimmyMow:
Wow @mrfelton `neutrino.js` and this entire refactor is a massive improvement to the codebase, this is wonderful.

Tested ACK 1c8a2e9

Empact:
Tested ACK 1c8a2e9

PR #531, opened 2018-07-04
Reduce default log level to info to reduce terminal noise

mrfelton:
Reduce our loggers to `info` level by default to reduce the amount of noise that gets output on the console.


Empact:
Tested ACK f7172d2

PR #532, opened 2018-07-04
Add more logging for lnd service

mrfelton:
Log all calls to lnd service methods so that we can get a better picture of our intraprocess communication.


JimmyMow:
utACK 9a6b886

Empact:
This seems like a lower-level message to me, e.g. debug. Thoughts?

mrfelton:
Debug is very very noisy. In https://github.com/LN-Zap/zap-desktop/pull/531 I have specifically reduced our default log level to `info`. These messages are very useful high-ish level info about what the app is doing and are very useful when included in the info level.

Empact:
utACK 9a6b886

PR #535, opened 2018-07-04
Fetch block height from multiple block explorers incase one is unavailable

### LINKED ISSUE #463: Redundant block explorer api backends

mrfelton:
Call out to 3 different block explorers to get the current block height and use the result from whichever responds quickest.

Fix #463

### DIFF
--- a/app/api/index.js
+++ b/app/api/index.js
@@ -29,11 +29,33 @@ export function requestTickers(ids) {
 }
 
 export function requestBlockHeight() {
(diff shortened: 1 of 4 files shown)


JimmyMow:
slow clap 👏... 👏... 👏

utACK d548915

Empact:
utACK d548915 👏

PR #538, opened 2018-07-06
Handle case where BTCd node is not fully synced with the blockchain

mrfelton:
Handle the case were the Neutrino backend BTCd node is still synchronising the blockchain as LND is not able to start syncing until the BTCd node that it is connected to is fully synced.

In this case, we display a message to the user indicating that we are waiting for peers. If it takes more than 10 seconds that generally indicates that the btcd node is quite far behind (eg, it has been restarted and is doing a full/fresh sync) and so after 10 seconds of waiting we advise the user that it could take some time.

This PR corrects a bug from https://github.com/LN-Zap/zap-desktop/pull/529 where we calculate the current block height incorrectly. In that PR, we incorrectly assumed that the message `Syncing to block height X` from lnd would tell us the current block height however it actually only tells us the highest block that BTCd is aware of (ie, the highest block that it has already synced). To resolve that and ensure that we know the actual current block height, we fetch the block height from the block explorers and we now do this as early as possible from the main process rather than waiting for the react `Syncing` component to mount.

### DIFF
--- a/app/components/Onboarding/Syncing.js
+++ b/app/components/Onboarding/Syncing.js
@@ -8,14 +8,59 @@ import { showNotification } from 'notifications'
 import styles from './Syncing.scss'
 
 class Syncing extends Component {
-  componentWillMount() {}
+  state = {
+    timer: null,
+    syncMessageDetail: null
+  }
+
+  componentWillMount() {
+    const { syncStatus } = this.props
+
+    // If we are still waiting for peers after some time, advise te user it could take a wile.
+    let timer = setTimeout(() => {
+      if (syncStatus === 'waiting') {
+        this.setState({
+          syncMessageDetail:
+            'It looks like this could take some time - you might want to grab a coffee or try again later!'
+        })
+      }
+    }, 10000)
+
+    this.setState({ timer })
+  }
+
+  componentWillUnmount() {
+    const { timer } = this.state
+    clearInterval(timer)
+  }
 
   render() {
-    const { hasSynced, syncPercentage, address, blockHeight, lndBlockHeight } = this.props
+    const {
+      hasSynced,
+      syncStatus,
+      syncPercentage,
+      address,
+      blockHeight,
+      lndBlockHeight
+    } = this.props
+    let { syncMessageDetail } = this.state
 
     const copyClicked = () => {
       copy(address)
       showNotification('Noice', 'Successfully copied to clipboard')
+    }
+    let syncMessage
+    if (syncStatus === 'waiting') {
+      syncMessage = 'Waiting for peers...'
+    } else if (typeof syncPercentage === 'undefined' || syncPercentage <= 0) {
+      syncMessage = 'Preparing...'
+      syncMessageDetail = null
+    } else if (syncPercentage > 0 && syncPercentage < 99) {
+      syncMessage = `${syncPercentage}%`
+      syncMessageDetail = `${lndBlockHeight.toLocaleString()} of ${blockHeight.toLocaleString()}`
+    } else if (syncPercentage >= 99) {
+      syncMessage = 'Finalizing...'
+      syncMessageDetail = null
     }
 
     if (typeof hasSynced === 'undefined') {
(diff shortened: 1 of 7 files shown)


mrfelton:
rebased

mrfelton:
rebased

Empact:
What are the repro steps for this? I don't see the problem locally.

mrfelton:
Lnd can not sync from a BTCd node that is itself not fully synced from the blockchain. It needs to wait until the BTCd node has been synced.

so, start up a BTCd node locally that is not fully synced, and then connect Zap to it (eg, simulate the case where the Zap BTCd node is in the middle of resyncing).

You can also try connecting to a BTCd node that doesn't exist to simulate the case where our BTCd node is offline.

And you can shut down your BTCd node mid-way through syncing, and then restart it again to simulate an intermittent connection issue with BTCd.
 
With PR it will tell the user that we are waiting for peers until BTCd is actually ready to start syncing from, and if this takes more than 10 seconds then we inform them that they should try again later - rather than showing a stuck progress indicator with no indication of whats going on as is the case now in master.

Once the btc backend is ready to sync from and lnd actually starts syncing, only then do we show the progress indicator to the user.

mrfelton:
The other issue is that in master, if you connect to a BTCd node that is mid-way through syncing, we calculate the current block height incorrectly as we assume that one of the numbers given to us in the lnd log output is the current block height, where as in actual fact it is only the current highest block that BTCd has actually synced.

mrfelton:
I've rebased this branch off https://github.com/LN-Zap/zap-desktop/pull/548. The changes are all in this commit https://github.com/LN-Zap/zap-desktop/pull/538/commits/449ec5a9fc561dd6fc7ad1ca3d961044c1584e64

app/components/Onboarding/Syncing.js
```diff
@@ -8,15 +8,60 @@ import { showNotification } from 'notifications'
...
+  state = {
+    timer: null,
+    syncMessageDetail: null
+  }
+
+  componentWillMount() {
+    const { syncStatus } = this.props
+
+    // If we are still waiting for peers after some time, advise te user it could take a wile.
+    let timer = setTimeout(() => {
```
Empact: nit: const

app/lnd/lib/neutrino.js
```diff
@@ -5,20 +5,37 @@ import config from '../config'
...
+const NEUTRINO_SYNC_STATUS_PENDING = 'chain-sync-pending'
+
+// Waiting for chain backend to finish synchronizing.
+const NEUTRINO_SYNC_STATUS_WAITING = 'chain-sync-waiting'
+
+// Initial sync is currently in progress.
+const NEUTRINO_SYNC_STATUS_IN_PROGRESS = 'chain-sync-started'
+
+// Initial sync has completed.
+const NEUTRINO_SYNC_STATUS_COMPLETE = 'chain-sync-finished'
```
Empact: `chain-sync-complete`?

app/lnd/lib/neutrino.js
```diff
@@ -5,20 +5,37 @@ import config from '../config'
...
 import { fetchBlockHeight } from './util'
 
+// Sync status is currenty pending.
+const NEUTRINO_SYNC_STATUS_PENDING = 'chain-sync-pending'
+
+// Waiting for chain backend to finish synchronizing.
+const NEUTRINO_SYNC_STATUS_WAITING = 'chain-sync-waiting'
+
+// Initial sync is currently in progress.
+const NEUTRINO_SYNC_STATUS_IN_PROGRESS = 'chain-sync-started'
```
Empact: `chain-sync-in-progress`

app/lnd/lib/neutrino.js
```diff
@@ -59,28 +76,43 @@ class Neutrino extends EventEmitter {
...
-      // LND syncing has started.
-      if (!this.state.chainSyncStarted) {
+      // If the sync has already completed then we don't need to do anythibng else.
+      if (this.is(NEUTRINO_SYNC_STATUS_COMPLETE)) {
+        return
+      }
+
+      // Lnd waiting for backend to finish syncing.
+      if (this.is(NEUTRINO_SYNC_STATUS_PENDING) || this.is(NEUTRINO_SYNC_STATUS_IN_PROGRESS)) {
+        if (
```
Empact: nit: could flatten these

app/lnd/lib/neutrino.js
```diff
@@ -98,38 +130,63 @@ class Neutrino extends EventEmitter {
...
+
+        if ((match = line.match(/Rescanned through block.+\(height (\d+)/))) {
+          height = match[1]
         } else if ((match = line.match(/Caught up to height (\d+)/))) {
-          this.emit('got-lnd-block-height', match[1])
+          height = match[1]
         } else if ((match = line.match(/Processed \d* blocks? in the last.+\(height (\d+)/))) {
-          this.emit('got-lnd-block-height', match[1])
+          height = match[1]
+        }
```
Empact: nit: could do or fallback here since the matches are consistent

Empact:
Seems good to me, could be a bit more consistent across naming the same concept.

Empact:
Tested ACK 637b27d

Issue #544, opened 2018-07-10
error in main javascript proccess


mntetai:
Steps to reproduce  -
I pulled latest version from master branch.
Got an error while zap was syncing.
Screenshot of error attached.

![error_in_main_10_07_18](https://user-images.githubusercontent.com/1884392/42527113-7590957a-8480-11e8-9d8e-37d1ceb826f0.png)

Debug log entries -
`Tue, 10 Jul 2018 17:29:39 GMT zap:main [INF]  Calling lnd method: { msg: 'channels', data: undefined }
Tue, 10 Jul 2018 17:29:39 GMT zap:lnd  [INF]  2018-07-10 20:29:39.860 [INF] RPCS: [listchannels] fetched 25 channels from DB
Tue, 10 Jul 2018 17:29:39 GMT zap:main [INF]  Calling lnd method: { msg: 'channels', data: undefined }
Tue, 10 Jul 2018 17:29:39 GMT zap:lnd  [INF]  2018-07-10 20:29:39.963 [INF] RPCS: [listchannels] fetched 25 channels from DB
Tue, 10 Jul 2018 17:29:41 GMT zap:lnd  [INF]  2018-07-10 20:29:41.018 [INF] HSWC: Sent 108 satoshis and received 0 satoshis in the last 10 seconds (0.400000 tx/sec)`

Os : win 10 pro
zap commit : 3c316d8d4d14ee538bad1a7a332d62e320630210

PR #546, opened 2018-07-11
fix: improve lnd log line matching

Empact:
When matching log lines, switch on the regex match rather than an
excerpt, so that there are never any false positives.

Note this allows conditional assignment within parens as well.


Empact:
Fixes #544

Issue #520, opened 2018-06-30
I should not have to go through the onboarding process every time I start the app


mrfelton:
> Once you are onboarded, after shutting down and running Zap again, connection type, node name and autopilot have to be chosen again. This is something that could be remembered. Maybe we could add an additional "change connection" button on the wallet unlock page and start right there instead?
 -- @michaelWuensch

mrfelton:
> I think the reason onboarded users see the process again is because they might want to turn on/off the autopilot option or choose custom connection. Wouldn't it be better these to be put in ZAP as an options away from the onboarding process?
 -- @odb366

mrfelton:
This has been addressed and an updated flow will be available in the next release, so closing this issue.

PR #547, opened 2018-07-12
feat(ux): keep app running when main window closed

mrfelton:
The standard approach on Darwin is to keen applications running when the window is closed. Clicking the apps icon again in the doc should bring back the application window.

This changeset refactors things to properly handle this. It will keep Zap and lnd running in the background when the main application window is closed. The gRPC connection is closed out when the window is closed and a new gRPC connection is established again once it has been reopened. Quitting the app will stop both the app and lnd.

I think there are further improvements that can be made to the user experience here as part of https://github.com/LN-Zap/zap-desktop/issues/520

### DIFF
--- a/app/lnd/lib/lightning.js
+++ b/app/lnd/lib/lightning.js
@@ -2,52 +2,112 @@ import grpc from 'grpc'
 import { loadSync } from '@grpc/proto-loader'
 import config from '../config'
 import { getDeadline, validateHost, createSslCreds, createMacaroonCreds } from './util'
+import methods from '../methods'
+import { mainLog } from '../../utils/log'
+import subscribeToTransactions from '../subscribe/transactions'
+import subscribeToInvoices from '../subscribe/invoices'
+import subscribeToChannelGraph from '../subscribe/channelgraph'
 
 /**
  * Creates an LND grpc client lightning service.
- * @returns {rpc.lnrpc.Lightning}
+ * @returns {Lightning}
  */
-const lightning = async () => {
-  const lndConfig = config.lnd()
-  const { host, rpcProtoPath, cert, macaroon } = lndConfig
-
-  // Verify that the host is valid before creating a gRPC client that is connected to it.
-  return await validateHost(host).then(async () => {
-    // Load the gRPC proto file.
-    // The following options object closely approximates the existing behavior of grpc.load.
-    // See https://github.com/grpc/grpc-node/blob/master/packages/grpc-protobufjs/README.md
-    const options = {
-      keepCase: true,
-      longs: String,
-      enums: String,
-      defaults: true,
-      oneofs: true
+class Lightning {
+  constructor() {
+    this.mainWindow = null
+    this.lnd = null
+    this.subscriptions = {
+      channelGraph: null,
+      invoices: null,
+      transactions: null
     }
-    const packageDefinition = loadSync(rpcProtoPath, options)
+  }
 
-    // Load gRPC package definition as a gRPC object hierarchy.
-    const rpc = grpc.loadPackageDefinition(packageDefinition)
+  /**
+   * Connect to the gRPC interface and verify it is functional.
+   * @return {Promise<rpc.lnrpc.Lightning>}
+   */
+  async connect() {
+    const lndConfig = config.lnd()
+    const { host, rpcProtoPath, cert, macaroon } = lndConfig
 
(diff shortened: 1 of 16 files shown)


app/lnd/lib/lightning.js
```diff
@@ -21,37 +26,111 @@ process.env.GRPC_SSL_CIPHER_SUITES =
...
+  subscribeToInvoices() {
+    this.subscriptions.invoices = subscribeToInvoices(this.mainWindow, this.lnd, mainLog)
+  }
+
+  /**
+   * Subscribe to the transaction stream.
+   */
+  subscribeToTransactions() {
+    this.subscriptions.transactions = subscribeToTransactions(this.mainWindow, this.lnd, mainLog)
+  }
```
Empact: How about inlining these into `subscribe` to minimize the interface. `subscriptions` could then be an array produced by a map.

mrfelton: I'm not quite sure what you mean about making subscriptions an array produced by a map, but I have refactored this to reduce the interface so that there is only the one `subscribe` method

app/main.dev.js
```diff
@@ -132,4 +126,16 @@ app.on('ready', () => {
       app.quit()
     }
   })
+
+  /**
+   * Add application event listener:
+   *  - reopen window when dock icon is clicked.
+   */
+  app.on('activate', () => {
```
JimmyMow: nice!

JimmyMow:
utACK 54b7cb1

@mrfelton can you rebase this so I can test more accurately?

mrfelton:
Hmm. It seems this isn't working as well as it did before for some reason. Leave this one with me for a while @JimmyMow

JimmyMow:
👍

mrfelton:
@JimmyMow I've updated this PR to improve the way that this works somewhat.

I have implemented [electron-redux](https://www.npmjs.com/package/electron-redux) to enable us to share application state between windows. This gives us a single state that is initialised from the main process and syncronised with every browser window subproces. This allows us to open a new browser window with the previous application state in place for a more seamless user experience on the mac.

Now, if you close a window everything keeps running in the background (except the gRPC connection which we terminate). When you click on the app icon in the doc it will open a new window, reconnect gRPC, load the most recent data, and return the user to their already logged in wallet view.

app/lnd/subscribe/channelgraph.js
```diff
@@ -1,8 +1,14 @@
-export default function subscribeToChannelGraph(mainWindow, lnd) {
+import { status } from 'grpc'
+
+export default function subscribeToChannelGraph(mainWindow, lnd, log) {
   const call = lnd.subscribeChannelGraph({})
 
   call.on('data', channelGraphData => mainWindow.send('channelGraphData', { channelGraphData }))
+  call.on('end', () => log.info('end'))
+  call.on('error', error => (error.code === status.CANCELLED ? null : log.error(error)))
```
JimmyMow: this can be simplified a bit because you're not utilizing the front half of the conditional:

`call.on('error', error => (error.code !== status.CANCELLED && log.error(error)))`

app/lnd/subscribe/invoices.js
```diff
@@ -1,8 +1,15 @@
...
 
-  call.on('data', invoice => mainWindow.send('invoiceUpdate', { invoice }))
+  call.on('data', invoice => {
+    log.info('INVOICE:', invoice)
+    mainWindow.send('invoiceUpdate', { invoice })
+  })
   call.on('end', () => log.info('end'))
-  call.on('error', error => log.error(error))
-  call.on('status', status => log.info('status:', status))
+  call.on('error', error => (error.code === status.CANCELLED ? null : log.error(error)))
```
JimmyMow: Same as the other comment, this can be simplified

app/lnd/subscribe/transactions.js
```diff
@@ -1,10 +1,15 @@
...
 export default function subscribeToTransactions(mainWindow, lnd, log) {
   const call = lnd.subscribeTransactions({})
+
   call.on('data', transaction => {
     log.info('TRANSACTION:', transaction)
     mainWindow.send('newTransaction', { transaction })
   })
   call.on('end', () => log.info('end'))
-  call.on('error', error => log.error('error: ', error))
+  call.on('error', error => (error.code === status.CANCELLED ? null : log.error(error)))
```
JimmyMow: Can be simplified

PR #548, opened 2018-07-12
Ensure that we fetch the correct block height in the case where BTCd is mid way through syncing

mrfelton:
Fetch the current block height from multiple block explorers early on in the sync process. This ensures that we get the correct block height in the case where our BTCd node is still mid way through syncing. Do this in the main process rather than in the render process.

### DIFF
--- a/app/components/Onboarding/Syncing.js
+++ b/app/components/Onboarding/Syncing.js
@@ -8,14 +8,7 @@ import { showNotification } from 'notifications'
 import styles from './Syncing.scss'
 
 class Syncing extends Component {
-  componentWillMount() {
-    const { fetchBlockHeight, blockHeight } = this.props
-
-    // If we don't already know the target block height, fetch it now.
-    if (!blockHeight) {
-      fetchBlockHeight()
-    }
-  }
+  componentWillMount() {}
 
   render() {
     const { hasSynced, syncPercentage, address, blockHeight, lndBlockHeight } = this.props
@@ -112,7 +105,6 @@ class Syncing extends Component {
 }
 
 Syncing.propTypes = {
-  fetchBlockHeight: PropTypes.func.isRequired,
   address: PropTypes.string.isRequired,
   hasSynced: PropTypes.bool,
   syncPercentage: PropTypes.number,
--- a/app/lnd/lib/neutrino.js
+++ b/app/lnd/lib/neutrino.js
@@ -3,6 +3,7 @@ import { spawn } from 'child_process'
 import EventEmitter from 'events'
 import config from '../config'
 import { mainLog, lndLog, lndLogGetLevel } from '../../utils/log'
+import { fetchBlockHeight } from './util'
 
 class Neutrino extends EventEmitter {
   constructor(alias, autopilot) {
@@ -14,9 +15,7 @@ class Neutrino extends EventEmitter {
       grpcProxyStarted: false,
       walletOpened: false,
       chainSyncStarted: false,
-      chainSyncFinished: false,
-      currentBlockHeight: null,
-      targetBlockHeight: null
+      chainSyncFinished: false
     }
   }
 
@@ -79,11 +78,23 @@ class Neutrino extends EventEmitter {
       if (!this.state.chainSyncStarted) {
         const match = line.match(/Syncing to block height (\d+)/)
         if (match) {
-          const height = match[1]
+          // Notify that chhain syncronisation has now started.
           this.state.chainSyncStarted = true
-          this.state.targetBlockHeight = height
           this.emit('chain-sync-started')
-          this.emit('got-final-block-height', height)
+
+          // This is the latest block that BTCd is aware of.
+          const btcdHeight = Number(match[1])
+          this.emit('got-current-block-height', btcdHeight)
+
+          // The height returned from the LND log output may not be the actual current block height (this is the case
+          // when BTCD is still in the middle of syncing the blockchain) so try to fetch thhe current height from from
+          // some block explorers just incase.
+          fetchBlockHeight()
+            .then(
+              height => (height > btcdHeight ? this.emit('got-current-block-height', height) : null)
+            )
+            // If we were unable to fetch from bock explorers at least we already have what BTCd gave us so just warn.
+            .catch(err => mainLog.warn(`Unable to fetch block height: ${err.message}`))
         }
       }
 
@@ -99,13 +110,13 @@ class Neutrino extends EventEmitter {
       if (this.state.chainSyncStarted) {
         let match
         if ((match = line.match(/Downloading headers for blocks (\d+) to \d+/))) {
-          this.emit('got-current-block-height', match[1])
+          this.emit('got-lnd-block-height', match[1])
         } else if ((match = line.match(/Rescanned through block.+\(height (\d+)/))) {
-          this.emit('got-current-block-height', match[1])
+          this.emit('got-lnd-block-height', match[1])
         } else if ((match = line.match(/Caught up to height (\d+)/))) {
-          this.emit('got-current-block-height', match[1])
+          this.emit('got-lnd-block-height', match[1])
         } else if ((match = line.match(/Processed \d* blocks? in the last.+\(height (\d+)/))) {
-          this.emit('got-current-block-height', match[1])
+          this.emit('got-lnd-block-height', match[1])
         }
       }
     })
--- a/app/lnd/lib/util.js
+++ b/app/lnd/lib/util.js
@@ -1,14 +1,55 @@
 import dns from 'dns'
 import fs from 'fs'
+import axios from 'axios'
 import { promisify } from 'util'
 import { lookup } from 'ps-node'
 import grpc from 'grpc'
 import isIP from 'validator/lib/isIP'
 import isPort from 'validator/lib/isPort'
+import get from 'lodash.get'
 import { mainLog } from '../../utils/log'
 
 const fsReadFile = promisify(fs.readFile)
 const dnsLookup = promisify(dns.lookup)
+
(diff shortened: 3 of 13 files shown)


app/components/Onboarding/Syncing.js
```diff
@@ -8,14 +8,7 @@ import { showNotification } from 'notifications'
...
 class Syncing extends Component {
-  componentWillMount() {
-    const { fetchBlockHeight, blockHeight } = this.props
-
-    // If we don't already know the target block height, fetch it now.
-    if (!blockHeight) {
-      fetchBlockHeight()
-    }
-  }
+  componentWillMount() {}
```
Empact: I think we can drop this method, right?

app/lnd/lib/neutrino.js
```diff
@@ -79,11 +80,23 @@ class Neutrino extends EventEmitter {
       if (!this.state.chainSyncStarted) {
         const match = line.match(/Syncing to block height (\d+)/)
         if (match) {
-          const height = match[1]
-          this.state.chainSyncStarted = true
-          this.state.targetBlockHeight = height
-          this.emit('chain-sync-started')
-          this.emit('got-final-block-height', height)
+          // Thhis is the latest block that BTCd is aware of.
+          let height = match[1]
```
Empact: Shadowed, and I think this can stay `const`, right?

app/components/Onboarding/Syncing.js
mrfelton: The reason why I left is because I know we still need it in the the next PR with part of https://github.com/LN-Zap/zap-desktop/pull/538 that is about to follow... Due to our lint rules, removing this method requires us to refactor this class into a pure function, which would then have to be refactored back in the next PR!

I know its not quite nice and clean, but can we just leave this empty method in for this one commit - it won\t be empty by the next commit!

app/lnd/lib/neutrino.js
```diff
@@ -79,11 +80,23 @@ class Neutrino extends EventEmitter {
...
+            .catch(err => {
+              mainLog.warn(`Unable to fetch block height: ${err.message}`)
+              return Promise.resolve(height)
+            })
+            .then(height => {
+              this.state.chainSyncStarted = true
+              this.state.targetBlockHeight = height
+              this.emit('chain-sync-started')
+              this.emit('got-final-block-height', height)
+              return
```
Empact: unnecessary return

app/lnd/lib/neutrino.js
```diff
@@ -79,11 +80,23 @@ class Neutrino extends EventEmitter {
...
+          fetchBlockHeight()
+            // If we were unable to fetchh from bock explorers, then use what BTCd gave us.
+            .catch(err => {
+              mainLog.warn(`Unable to fetch block height: ${err.message}`)
+              return Promise.resolve(height)
+            })
+            .then(height => {
+              this.state.chainSyncStarted = true
+              this.state.targetBlockHeight = height
+              this.emit('chain-sync-started')
```
Empact: We can emit `chain-sync-started` earlier, right?

app/lnd/lib/util.js
```diff
@@ -10,6 +11,45 @@ import { mainLog } from '../../utils/log'
...
+      .then(response => {
+        const height = Number(path.split('.').reduce((a, b) => a[b], response.data))
+        mainLog.info(`Fetched block height as ${height} from: ${response.config.url}`)
+        return height
+      })
+      .catch(() => null)
+  }
+
+  const promises = []
+  sources.forEach(source => promises.push(fetchData(source.baseUrl, source.path)))
```
Empact: nit: `map`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

app/lnd/lib/neutrino.js
```diff
@@ -79,11 +80,23 @@ class Neutrino extends EventEmitter {
...
+          // Which may not be the actualy current block heighht, so try to fetch that from some block explorers.
+          fetchBlockHeight()
+            // If we were unable to fetchh from bock explorers, then use what BTCd gave us.
+            .catch(err => {
+              mainLog.warn(`Unable to fetch block height: ${err.message}`)
+              return Promise.resolve(btcdHeight)
+            })
+            .then(height => {
+              this.state.targetBlockHeight = height
+              this.emit('got-final-block-height', height)
```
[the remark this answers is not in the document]
mrfelton: They way I have it we will emit the height fetched from the explorer or if none was fetched then the height reported buy btcd will be emitted as a fallback.

This `then` block should have `height` set as `btcdHeight` from `Promise.resolve(btcdHeight)` in the `catch` block above.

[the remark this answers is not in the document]
Empact: Yeah sorry, I'm referring to the 2nd `catch`, thoughts on removing that?

app/lnd/lib/util.js
mrfelton: lint requires it

app/lnd/lib/neutrino.js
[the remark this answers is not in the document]
mrfelton: lint requires it

app/lnd/lib/util.js
mrfelton: updated

mrfelton:
@Empact I found a bug in `fetchBlockHeight` and have added `lodash.get` as a dependency in order to get round it as this enables me to select elements from json using a path, which I was trying to do manually before but it was failing

app/lnd/lib/neutrino.js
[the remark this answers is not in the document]
Empact: Ah yeah, so removing it would make failure silent, but we should want noisy error so how about just throwing the error there? Or log + throw?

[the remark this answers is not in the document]
mrfelton: We can do, although if you look at the code in the block above, I don't think it is possible that there could be an error since the only thing we are doing is setting a variable and emitting an event. The thing that could have had an  error was dealt with in the previous catch block. I'll add a log entry just for completeness sake...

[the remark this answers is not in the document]
mrfelton: added error message

app/lnd/lib/neutrino.js
```diff
@@ -79,11 +80,23 @@ class Neutrino extends EventEmitter {
       if (!this.state.chainSyncStarted) {
         const match = line.match(/Syncing to block height (\d+)/)
         if (match) {
-          const height = match[1]
           this.state.chainSyncStarted = true
-          this.state.targetBlockHeight = height
           this.emit('chain-sync-started')
-          this.emit('got-final-block-height', height)
```
Empact: Would it work to issue this then re-issue it? Would make for more immediate feedback if it works.

mrfelton: Yes, nice idea. I've updated this so that we issue the btcd height as soon as we have it, and then follow up with the highest of either the btcd height or the fetched height once we have that

app/lnd/lib/neutrino.js
```diff
@@ -79,11 +80,26 @@ class Neutrino extends EventEmitter {
...
+          this.emit('got-final-block-height', btcdHeight)
+
+          // The height returned from the LND log output may not be the actual current block height (this is the case
+          // when BTCD is still in the middle of syncing the blockchain) so try to fetch thhe current height from from
+          // some block explorers just incase.
+          fetchBlockHeight()
+            .then(height => {
+              const highestHeight = Math.max(height, btcdHeight)
+              this.state.targetBlockHeight = highestHeight
+              this.emit('got-final-block-height', highestHeight)
```
Empact: How about only update if height has change?

mrfelton:
@Empact have updated this again to only emit a new current block hight what we fetched was larger than what we got from btcd.

Additionally, I have renamed two events:

`got-current-block-height' -> `got-block-height'
`got-target-block-height' -> `got-current-block-height'

These names are more consistent with what we use elsewhere, and I think more accurate terminology as the current highest is usually referred to as the 'current block height'. I have never heard of a `target block height` before!

I left these changes as an additional commit to make it easier for you to review. Can squash once ready to merge.

Empact:
I'm not crazy about the rename because target in this case is relative to the progress bar that we're displaying. Block height doesn't have a "target" except with relative measures, e.g. coinbase tx maturity, or timelock expiration. "Current block height" vs "block height" doesn't tell me which it's referring to: the height of the chain vs the height we have scanned to.

mrfelton:
I'm not sure I totally agree with that. I believe the industry definition is: 

**Current block height:** The top most block of the blockchain.

This is exactly as it is used in our updated code. We are fetching the top of the block chain and then synchronising towards that.

What I think is confusing is taking an industryt standard term like "current block height" and using it to referring to something different (the current sync progress of lnd)

I agree that "block height" isn't exactly clear what it means so I would suggest renaming that to "lnd block height" for clarity (which is actually how we refer to it within the renderer).

So we would have:

**Current block height:** The top most block of the blockchain which we are syncing towards
**Lnd block height:** The current block that lnd has syncronised.

That sound ok?

mrfelton:
Pushed an updated version like so that you can see and review.

mrfelton:
The other argument I would make is that it's the renderer's decision to use the current block height as the target for a progress indicator, but it could also be used for other things within the app that are unrelated to 'targets'. We are just telling the renderer what the current block height is. So I don't think it makes sense to tie some generic data to one specifically use case by naming it specifically for that.

Empact:
utACK fb628cb squash please?

mrfelton:
squashed!

Empact:
utACK f7ba4f2

PR #549, opened 2018-07-13
Show channel filter when there is an empty active filter

### LINKED ISSUE #507: GUI: when an empty channel list is selected the channel menu disappears.
Under My Network column there is the possibility to filter out the channels to show. By default the option is All but it is possible to select also Online, Offline, Pending and Closing.
If I select one option that has no entries, for example Pending while I do not have any pending channel, the result is "Hmmm, looks like you don't have any channels yet. Here are some suggested nodes to open a channel with to get started"
This looks fine but it disappear the drop down menu, so I cannot return to the All or any other channel selection. The only option to see the full list of channels is to shutdown and restart the application.

mrfelton:
Ensure that the channel filter shows instead of the suggested nodes list when an empty channel filter is selected.

Fix #507

### DIFF
--- a/app/components/Contacts/Network.js
+++ b/app/components/Contacts/Network.js
@@ -31,7 +31,8 @@ class Network extends Component {
         filter,
         selectedChannel,
         loadingChannelPubkeys,
-        closingChannelIds
+        closingChannelIds,
+        channels
       },
       currentChannels,
       balance,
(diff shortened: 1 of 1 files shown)


app/components/Contacts/Network.js
```diff
@@ -175,9 +176,13 @@ class Network extends Component {
         <div className={styles.channels}>
           {!loadingChannelPubkeys.length &&
             !currentChannels.length &&
+            !channels.length &&
             !searchQuery.length && <SuggestedNodes {...suggestedNodesProps} />}
 
-          {(loadingChannelPubkeys.length || currentChannels.length) && (
+          {(loadingChannelPubkeys.length ||
+            currentChannels.length ||
+            channels.length ||
```
JimmyMow: nice

JimmyMow:
utACK 8e195c4

Empact:
Should we also add the `pendingOpenChannelsSelector` to this?

Empact:
utACK 8e195c4 we can revisit the pending open case

PR #550, opened 2018-07-13
Don't log successful results from lnd as errors

mrfelton:
Use `info` log level when logging successful results.

Previously we were logging successful results as errors.


JimmyMow:
utACK d16b6e4

Empact:
utACK d16b6e4

Issue #539, opened 2018-07-06
Received on-chain transaction shows as a sent transaction in the transaction detail modal


mrfelton:
Click on a received on-chain transaction from the transaction list and you see something like this:

<img width="880" alt="screenshot 2018-07-06 18 12 49" src="https://user-images.githubusercontent.com/200251/42389196-351bdf62-8148-11e8-9be9-b451d789d51b.png">

In the top left is shows 'Sent', yet tis was an incoming transaction.

SteliosRammos:
The Zap notification on Mac also mentions a send transaction instead of receiving funds.

bolatovumar:
Thanks for the catch @SteliosRammos.

PR #551, opened 2018-07-13
Don't show sent for received tx

mrfelton:
It seems that the fix in https://github.com/LN-Zap/zap-desktop/pull/542 for #539 was the wrong way round (sent showing as received, received showing as sent). This reverses it.

Fix for #539


JimmyMow:
utACK 48681ba

Empact:
utACK 48681ba

PR #552, opened 2018-07-13
Ensure there are sensible default values for onboarding screens

mrfelton:
- Set the default connection type to `local`
 - Set the default for autopilot to `true`


JimmyMow:
utACK 28e5310

Empact:
utACK 28e5310

PR #554, opened 2018-07-13
style(syncing): update syncing address style

JimmyMow:
Update our sync address style to be consistent within the app

screen shot 2018-07-13 at 6 11 58 pm


Empact:
utACK e131651

PR #557, opened 2018-07-15
fix(ui): increase currency selector z-index

mrfelton:
Ensure that the currency selector dropdown on the pay/request form shows ontop of other elements on the page to ensure that the full area of the dropdown is clickable.

(currently it is partially hidden behind the errors message divs.

screenshot 2018-07-15 19 34 24


JimmyMow:
utACK 5fac7b1

PR #558, opened 2018-07-15
feat(ui): show activity type icon in activity list

mrfelton:
Currently, it's not immediately obvious which transactions are on-chain vs off-chain when scanning the activity list.

Add an icon that shows the activity type in the activity list to help users clearly distinguish between on-chain and off-chain transactions.

**Before:**
screenshot 2018-07-15 23 01 49

**After:**
screenshot 2018-07-15 23 02 19

### DIFF
--- a/app/routes/activity/components/components/Activity.scss
+++ b/app/routes/activity/components/components/Activity.scss
@@ -16,14 +16,15 @@
     opacity: 0.5;
   }
 
-  .pendingIcon {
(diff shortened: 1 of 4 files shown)


mrfelton:
Have updated to add tooltips. eg:

screenshot 2018-07-19 20 04 52

JimmyMow:
utACK 3d2e4dd 🎊

PR #562, opened 2018-07-17
feat(btcpayserver): add support for BTCPayServer

mrfelton:
Allow connecting to an LND instance that is behind BTCPayServer.

### DIFF
--- a/app/components/Onboarding/BtcPayServer.js
+++ b/app/components/Onboarding/BtcPayServer.js
@@ -1,0 +1,58 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styles from './BtcPayServer.scss'
+
+const BtcPayServer = ({
+  connectionString,
+  connectionStringIsValid,
+  setConnectionString,
+  startLndHostError
+}) => (
+  <div className={styles.container}>
+    <section className={styles.input}>
+      <label htmlFor="connectionString">Connection String:</label>
+      <textarea
+        type="text"
+        id="connectionString"
+        rows="10"
+        placeholder="BTCPay Server Connection String"
+        className={
+          connectionString && (startLndHostError || !connectionStringIsValid)
+            ? styles.error
+            : undefined
+        }
+        ref={input => input}
+        value={connectionString}
+        onChange={event => setConnectionString(event.target.value)}
+      />
+      <p className={styles.description}>
+        Paste the full content of your BTCPay Server connection config file. This can be found by
+        clicking the link entitled &quot;Click here to open the configuration file.&quot; in your
+        BTCPay Server gRPC settings.
+      </p>
+      <p
+        className={`${styles.errorMessage} ${
+          connectionString && !connectionStringIsValid ? styles.visible : undefined
+        }`}
+      >
+        Invalid connection string.
+      </p>
+      <p
+        className={`${styles.errorMessage} ${
+          connectionString && startLndHostError ? styles.visible : undefined
+        }`}
+      >
+        {startLndHostError}
+      </p>
+    </section>
+  </div>
+)
+
+BtcPayServer.propTypes = {
+  connectionString: PropTypes.string.isRequired,
+  connectionStringIsValid: PropTypes.bool.isRequired,
+  setConnectionString: PropTypes.func.isRequired,
+  startLndHostError: PropTypes.string
+}
+
+export default BtcPayServer
--- a/app/lnd/config/index.js
+++ b/app/lnd/config/index.js
@@ -58,12 +58,18 @@ if (process.env.NODE_ENV === 'development') {
 }
 
 export default {
-  lnd: () => ({
-    lndPath,
-    configPath: join(appRootPath, 'resources', 'lnd.conf'),
-    rpcProtoPath: join(appRootPath, 'resources', 'rpc.proto'),
-    host: store.get('host') || 'localhost:10009',
-    cert: store.get('cert') || join(userInfo().homedir, loc),
-    macaroon: store.get('macaroon') || join(userInfo().homedir, macaroonPath)
-  })
+  lnd: () => {
+    const cert = store.get('cert')
+    const host = store.get('host')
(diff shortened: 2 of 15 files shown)


app/components/Onboarding/BtcPayServer.js
```diff
@@ -0,0 +1,54 @@
...
+}) => (
+  <div className={styles.container}>
+    <section className={styles.input}>
+      <label htmlFor="connectionString">Connection String:</label>
+      <textarea
+        type="text"
+        id="connectionString"
+        rows="9"
+        placeholder="BTCPayServer Connection String"
+        className={`${
```
Empact: Quoting looks unnecessary here

mrfelton: Could you elaborate please @Empact . I'm not sure I follow

Empact: You should be able to remove `\`${` bit, and the closing, because then it would resolve down to `undefined` or `styles.error` directly, which don't need quoting afaict.

mrfelton: Gotcha. You are right - I've altered that. Thanks.

app/lnd/lib/lightning.js
```diff
@@ -16,7 +16,9 @@ process.env.GRPC_SSL_CIPHER_SUITES =
     'ECDHE-ECDSA-AES128-GCM-SHA256',
     'ECDHE-ECDSA-AES256-GCM-SHA384',
     'ECDHE-ECDSA-AES128-CBC-SHA256',
-    'ECDHE-ECDSA-CHACHA20-POLY1305'
+    'ECDHE-ECDSA-CHACHA20-POLY1305',
+    'ECDHE-RSA-AES256-GCM-SHA384',
+    'ECDHE-RSA-AES128-GCM-SHA256'
```
Empact: Could you add a comment as to the need for these?

mrfelton: I've added a comment to that effect.

app/components/Onboarding/BtcPayServer.js
```diff
@@ -0,0 +1,53 @@
...
+            ? styles.error
+            : undefined
+        }
+        ref={input => input}
+        value={connectionString}
+        onChange={event => setConnectionString(event.target.value)}
+      />
+      <p
+        className={`${styles.errorMessage} ${
+          connectionString && !connectionStringIsValid ? styles.visible : undefined
```
Empact: nit: explicit precedence

app/components/Onboarding/BtcPayServer.js
```diff
@@ -0,0 +1,53 @@
...
+        }`}
+      >
+        Invalid connection string.
+      </p>
+      <p
+        className={`${styles.errorMessage} ${
+          connectionString && startLndHostError ? styles.visible : undefined
```
Empact: nit: explicit precedence

app/components/Onboarding/BtcPayServer.js
```diff
@@ -0,0 +1,53 @@
...
+  <div className={styles.container}>
+    <section className={styles.input}>
+      <label htmlFor="connectionString">Connection String:</label>
+      <textarea
+        type="text"
+        id="connectionString"
+        rows="9"
+        placeholder="BTCPayServer Connection String"
+        className={
+          connectionString && (startLndHostError || !connectionStringIsValid)
```
Empact: nit: explicit precedence

app/lnd/config/index.js
```diff
@@ -58,12 +58,18 @@ if (process.env.NODE_ENV === 'development') {
...
+    const host = store.get('host')
+    const macaroon = store.get('macaroon')
+
+    return {
+      lndPath,
+      configPath: join(appRootPath, 'resources', 'lnd.conf'),
+      rpcProtoPath: join(appRootPath, 'resources', 'rpc.proto'),
+      host: typeof host === 'undefined' ? 'localhost:10009' : host,
+      cert: typeof cert === 'undefined' ? join(userInfo().homedir, loc) : cert,
+      macaroon: typeof macaroon === 'undefined' ? join(userInfo().homedir, macaroonPath) : macaroon
```
Empact: Why this change? `false` does not seem valid or meaningful for these.

mrfelton: The purpose is that no value is a valid value. For example, if an empty string has been set we should not assume the default values.

app/components/Onboarding/BtcPayServer.js
mrfelton: Can you explain what you mean by this please @Empact ? Thanks.

Empact: I prefer making precedence explicit anywhere there is potential ambiguity, in this case between `&&` & `?`. e.g.

`(connectionString && !connectionStringIsValid) ? styles.visible : undefined`

mrfelton:
This has been rebased and updated to work with the latest BTCPayServer connection config file format.

app/components/Onboarding/BtcPayServer.js
mrfelton: Our lint rules do not allow for this. Prettier automatically removes the brackets.

JimmyMow:
Tested ACK c247e7a 🎊

Empact:
"BTCPay Server" rather than "BTCPayServer"
https://github.com/btcpayserver/btcpayserver

mrfelton:
I've updated "BTCPayServer" to read "BTCPay Server" everywhere where it is printed in the UI

Empact:
nit: Could reduce the font size on the connection string input - it's significcantly heavier than most other text and limits what can be seen:
screen shot 2018-07-22 at 14 40 52

mrfelton:
Update font style:

**without error:**
screenshot 2018-07-22 20 46 17

**with error:**
screenshot 2018-07-22 20 46 26

Empact:
Tested ACK fd7f447 ⚡️

PR #563, opened 2018-07-18
feat(refresh): ability to refresh activity list

### LINKED ISSUE #419: Ability to manually refresh activity list
Because we are now setting up to allow remote devices to control our Zap node we can have instances where I create an invoice on my phone and it will not update on my Desktop interface. Easiest and simplest way to resolve this is just have a refresh icon that fetches invoices, payments and transactions on click. Exactly how we do refresh of the channels list

mrfelton:
Add a refresh button in the activity list header which - when clicked - fetches the latest payments, invoices, transactions, and balance from lnd.

Fix #419

### DIFF
--- a/app/routes/activity/components/Activity.js
+++ b/app/routes/activity/components/Activity.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import Isvg from 'react-inlinesvg'
 import searchIcon from 'icons/search.svg'
 import xIcon from 'icons/x.svg'
+import FaRepeat from 'react-icons/lib/fa/repeat'
 
 import Wallet from 'components/Wallet'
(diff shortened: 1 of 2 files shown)


JimmyMow:
utACK 31b0fcb - from a code PoV

Can you post a screenshot of what it looks like @mrfelton? This will change the UI slightly, wanna see what you went with

mrfelton:
The look and feel is based on the refresh button in the network list. It says 'refresh' and when you click on it it turns in to a circle spinner whilst the refresh happens. See below:

**Normal:**
screenshot 2018-07-19 08 12 06

**Active:**
screenshot 2018-07-19 08 12 32 2

JimmyMow:
Nice. utACK 31b0fcb

PR #565, opened 2018-07-18
fix: aezeedPasswordConfirmation propagation

Empact:
I saw a warning that a required argument to NewAezeedPassword was not
present and tracked it back to here.

updateAezeedPasswordConfirmation is a function and should be sourced
via the dispatchProps, while the aezeedPasswordConfirmation holds the
value.


JimmyMow:
utACK 3f75445

PR #566, opened 2018-07-18
updated advanced documentation with details about custom lnd

nettijoe96:
added details on symlinking lnd, filling in fields that the gui asks for (macaroon, hostname, cert), and unlocking wallet (lncli unlock)


mrfelton:
This section of the documentation was recently updated. We have also added inline help text within the app.

PR #571, opened 2018-07-19
fix(onboarding): remove seed password

mrfelton:
Remove seed password support from the onboarding process until we can find a better/less confusing way to present it to users.


JimmyMow:
👏 👏 👏 

utACK d187082

Empact:
Tested ACK d187082

PR #572, opened 2018-07-19
fix(onboarding): default to create new wallet

mrfelton:
It is more common for new users to create a new wallet than to import an existing one. Make the "create new wallet" option the default selected value in the onboarding process.


JimmyMow:
utACK 08e516a

Empact:
utACK 08e516a

PR #573, opened 2018-07-19
fix: ensure conditional display uses boolean

mrfelton:
Prevents the display of an unwanted "0" by ensuring that conditional uses an actual boolean value.

screenshot 2018-07-20 00 27 17


Empact:
utACK 828be27

PR #574, opened 2018-07-19
fix: ensure received transactions show as such

mrfelton:
Inject `received` property into transactions received from the IPC receiveTransactions handler. This ensures that past transactions can correctly be displayed as either sent or received.

To test, open up a wallet with some existing on-chain transactions and verify that received on-chain transactions show as such in the wallet

### DIFF
--- a/app/reducers/transaction.js
+++ b/app/reducers/transaction.js
@@ -24,6 +24,16 @@ export const SHOW_SUCCESS_TRANSACTION_SCREEN = 'SHOW_SUCCESS_TRANSACTION_SCREEN'
 export const HIDE_SUCCESS_TRANSACTION_SCREEN = 'HIDE_SUCCESS_TRANSACTION_SCREEN'
 
 // ------------------------------------
+// Helpers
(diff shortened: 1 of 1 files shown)


Empact:
How about extracting a method so that future additions have a natural place? E.g. “decorateTransaction”

mrfelton:
Good idea @Empact - I've pushed an updated version of this which extracts the transaction decorating.

JimmyMow:
utACK 9a8ff4d

app/reducers/transaction.js
```diff
@@ -65,7 +75,8 @@ export const receiveTransactions = (event, { transactions }) => (dispatch, getSt
   const state = getState()
   const currentAddress = state.address.address
   let usedAddresses = []
-  transactions.forEach(transaction => {
+  transactions.map(transaction => {
```
Empact: No need for `map` here as you're not using the returned array.

mrfelton: right, fixed. Thanks.

PR #575, opened 2018-07-19
fix: only notify if we have not seen the transaction before

Empact:
LND errs on the side of providing redundant transaction reports, so
we should filter before issuing notifications.

### DIFF
--- a/app/reducers/transaction.js
+++ b/app/reducers/transaction.js
@@ -121,26 +121,29 @@ export const transactionError = (event, { error }) => dispatch => {
 }
 
 // Listener for when a new transaction is pushed from the subscriber
-export const newTransaction = (event, { transaction }) => dispatch => {
-  // Fetch new balance
-  dispatch(fetchBalance())
+export const newTransaction = (event, { transaction }) => (dispatch, getState) => {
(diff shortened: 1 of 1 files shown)


app/reducers/transaction.js
```diff
@@ -110,26 +110,29 @@ export const transactionError = (event, { error }) => dispatch => {
...
+      ? "Lucky you, you just received a new on-chain transaction. I'm jealous."
+      : "Hate to see 'em go but love to watch 'em leave. Your on-chain transaction successfully sent."
 
-  showNotification(notifTitle, notifBody)
+    showNotification(notifTitle, notifBody)
 
-  // Generate a new address
-  dispatch(newAddress('np2wkh'))
+    // Generate a new address
+    dispatch(newAddress('np2wkh'))
```
JimmyMow: Not sure if it should be part of this PR but we shouldn't be generating a new address every time right? We should only generate if the tx was a receiving tx. If it was the user sending BTC then they don't need a new address

Empact: Great idea - fixed in bcc86db

JimmyMow:
utACK b2c9f98

PR #576, opened 2018-07-20
feat(onboarding): require explicit external connection configuration

mrfelton:
Currently when the app starts up we detect wether the user is running lnd locally or not, and if we detect that they are we connect zap to that automatically.

This is confusing from a user perspective as they have no idea what happened, or what zap has just connected itself to or why the behaviour of zap is different from another time when they ran it when they didn't have lnd already running on their localhost already.

This feature predates the custom connection screen, and I think conflicts with it. The user should just have two very easy to understand options

1. `default` mode runs lnd from `resources` dir with our own configs (neutrino mode, connected to the zap fullnode backend)
2. `custom` mode allows you to choose your connection params explicitly. It should provide the ability top use sensible default values for a localhost connection, and default to the last used settings.

In this PR I remove the lnd process lookup.


Empact:
I worry this would represent a regression in user experience by requiring knowledge / interaction with values which can be determined automatically. This is more of an advanced user interface imo, than one that should be presented to everyone.

mrfelton:
The custom connection screen is for advanced users only - it specifically says that on the screen where users are given the option to use `default` or `custom` mode.

My thought process is that at least this way the user has a chance to understand what zap is going to do. It's unpredictable at the moment, and I think the number of people asking for help recently in slack is testament to that fact.

Just because a user might have lnd running on their local machine doesn't mean that they want to connect zap to it.

mrfelton:
@JimmyMow there are 2 aspects to this PR.

I think that the first aspect is a must - removing thing thing where we detect if the user is already running lnd on their host and try to connect to it automatically if so.

The second part is to add the `suggested value` feature into the custom connection onboarding screen, which I'm in two minds about. One the one hand I sort of question why we are suggesting values for localhost over anything else. On the other hand it is possible that some people may want to connect to lnd on their localhost and this potentially makes it easier for them.

mrfelton:
I have pulled the second part of this out into a separate PR: https://github.com/LN-Zap/zap-desktop/pull/640

This PR now only deals with removing the automajical behaviour of connecting to an existing lnd instance if we happen to find one.

JimmyMow:
Tested ACK b54fd26

JimmyMow:
Lol `automajical` 😆

PR #578, opened 2018-07-22
feat(help): add help link to tutorial

JimmyMow:
Add a help link to our onboarding forms that links to our Desktop "Getting Started" tutorial in case anyone gets stuck

screen shot 2018-07-21 at 8 53 00 pm

### DIFF
--- a/app/components/Onboarding/FormContainer.js
+++ b/app/components/Onboarding/FormContainer.js
@@ -1,3 +1,4 @@
+import { shell } from 'electron'
 import React from 'react'
 import PropTypes from 'prop-types'
 import Isvg from 'react-inlinesvg'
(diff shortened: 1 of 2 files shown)


Empact:
Concept ACK but I think we should link to a branch other than master, as this would expose documentation updates that aren't necessarily relevant to the released version. A `0.2` series branch would be good.

mrfelton:
Could we make a Github Pages site from this repo so that it looks a little more user friendly and in keeping with our branding?

https://blog.github.com/2016-08-22-publish-your-project-documentation-with-github-pages/

Empact:
utACK 81c9d45 Site looks good - and I guess we can modify the path as-needed if we need to target information at specific versions. Just should be mindful of this as we go forward with releases.

PR #583, opened 2018-07-22
fix(onboarding): improve custom connection form field help text

mrfelton:
Don't use field placeholders for form field help text as this is only visible when the form field is empty. Instead.

This PR moves the help text to just below the fields so that it is always visible to the user in order to help them ensure they are entering the correct values. Also, I have added example values so that the user knows what format we are expecting the values in (we get a lot of people asking what exactly they should enter in slack)

Here is decent article that highlights why placeholders are bad in many cases https://uxdesign.cc/alternatives-to-placeholder-text-13f430abc56f. Using for help text presents a usability issue.

**before:**
(no indication of what the user should be entering is visible once the user has started typing)
screenshot 2018-07-22 17 25 20

**after:**
(help text is always visible)
screenshot 2018-07-22 17 21 24

screenshot 2018-07-22 17 21 44

### DIFF
--- a/app/components/Onboarding/ConnectionDetails.js
+++ b/app/components/Onboarding/ConnectionDetails.js
@@ -19,12 +19,14 @@ const ConnectionDetails = ({
       <input
         type="text"
         id="connectionHost"
-        placeholder="Hostname / Port of the Lnd gRPC interface"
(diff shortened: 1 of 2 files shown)


Empact:
utACK 6b13d9a

PR #584, opened 2018-07-22
chore(security): tailor GRPC_SSL_CIPHER_SUITES to the connection type

Empact:
This avoids expanding the acceptable cipher suites in general lnd connections
relative to the possibility of BTCPay server being involved.


JimmyMow:
Tested ACK 08cb136

PR #585, opened 2018-07-23
fix(transaction notification): fix that state access was off the mark

Empact:
And make the test more robust against unexpected state.


JimmyMow:
Tested ACK bfd1383

PR #587, opened 2018-07-23
fix(version): bump version

JimmyMow:
Bump to 0.2.0-beta

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "zap-desktop",
   "productName": "ZapDesktop",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "description": "desktop application for the lightning network",
   "main": "./main.prod.js",
   "author": {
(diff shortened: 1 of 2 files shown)


mrfelton:
The current release is `v0.1.1-beta`, therefore the next release should either be:

- `v0.2.0-beta` - We keep it in beta, but increment the `minor` number since this release contains new features.
- `v0.2.0` - If we want to bump it and take it out of beta then the `minor` number should be incremented since there are new features (as above), and also the pre-release flag (`beta`) would be removed
- `v0.1.1` - If we simply wanted to take the current version out of beta then we should only remove the prerelease flag only.

https://semver.org/

I don't see how moving from `v0.1.1-beta` to `0.1.2` makes sense. That would indicate that this is just a patch release (which it is not), and that we are coming out of beta (which we are not)

I would think `v0.2.0-beta` makes the most sense, semantically speaking.

Empact:
Yeah I'm also in favor of `v0.2.0-beta`

Empact:
utACK 7072ded confirmed there are no other zap version strings in the codebase

Issue #604, opened 2018-07-25
UX improvements


SDWouters:
Keep up the great work! Here's some feedback from taking a quick look:

1. When I try to set up a channel denominated in bits, I get an error that tells me the maximum channel capacity is 0.16 btc, this should be displayed in bits (or satoshis if I select those). This may also be a bug, because I only selected 500 bits and it wouldn't go through.

2. After trying to send a testnet lightning payment, I receive a visual confirmation in a red bar at the top which states "unable to find a path to destination". It would help the user to get instructions of what the next step is to try to find a path. e.g. "Unable to find a path to this destination. Try to _open more payment channels_ first!"

3. It would help users to see a small hint of what needs to happen for a "pending" channel to go through, e.g. "Pending until 3 confirmations on the blockchain (~30 minutes)."

4. Finally, before setting up channels I receive 3 suggestions. After setting up a channel with one of them, the other suggestions disappear. Perhaps those should still be displayed to encourage people to set up several channels? That or a list of other possible peers to set up a channel with.

skcheffy:
Great suggestions!  I have been compiling some items and you touched on some.  I will be putting some down soon.

PR #606, opened 2018-07-25
fix: honour currency when opening channel

### LINKED ISSUE #523: Limits on channel balances when opening a new channel should respect the selected default currency
Some feedback from Slack:

> Opening a channel throws an error if the amount is to big. It seems the value is just compared with the number typed in. If you set your default currency unit to satoshi, you won't be able to open the channel as everything larger than 0.17 sathoshi is considered too large
 -- @michaelWuensch

mrfelton:
Ensure that the currently active unit of account is honoured when opening a new channel.

Fix #523
Also see #604


Empact:
Tested ACK 8220ed0

PR #608, opened 2018-07-25
fix(copy): fix small typo in login copy

JimmyMow:


Empact:
Could you drop the package-lock?

PR #610, opened 2018-07-26
fix(copy): fix small typo re adding funds to network

terencechow:


mrfelton:
@Empact this is now failing ci on eslint:

screenshot 2018-07-26 15 40 54

PR #612, opened 2018-07-26
feat(onboarding): allow login with Enter key press

pajasevi:
Quicker and easier login with Enter key.


Empact:
Tested ACK 2fa918c

PR #631, opened 2018-07-30
fix(config): use `app.getPath()` to get data dir

mrfelton:
Use electron's `app.getPath()` method to get the path to our data directory. This removes redundant custom code from our app and supports the XDG Base Directory Specification standard on linux.

See:
 - https://github.com/electron/electron/blob/master/docs/api/app.md#appgetpathname
 - https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html


mrfelton:
This was a mistake. Ignore.

Issue #620, opened 2018-07-28
Gracefully handle gRPC connection issues


mrfelton:
## Problem / Motivation

When a gRPC connection is lost for one reason or another the exception is currently unhandled, the user sees an error message, and the app exits. 

## Proposed Solution

We need to take steps to gracefully handle these issues, indicating to the user that there is an issue whilst we try to reestablish gRPC connectivity for them.

## Related Issues

- [ ] https://github.com/LN-Zap/zap-desktop/issues/611
- [ ] https://github.com/LN-Zap/zap-desktop/issues/601
- [ ] https://github.com/LN-Zap/zap-desktop/issues/207

PR #633, opened 2018-07-30
refactor(neutrino): more consistent naming in Neutrino class

mrfelton:
This PR cleans up some of the naming of methods and events in relation to the Neutrino client to make them more consistent and descriptive.

I'd like to try to get this in before starting work on https://github.com/LN-Zap/zap-desktop/issues/620 as this will make it easier to do a cleaner solution for that.


JimmyMow:
Tested ACK 8b4c863

PR #639, opened 2018-07-31
Feature/multiple fiat currency support

JimmyMow:
This PR fetches BTC to fiat data from blockchain.info and allows Zap to support the following fiat conversions:

`USD, EUR, JPY, GBP, CAD, KRW, AUD, BRL, CHF, CLP, CNY, DKK, HKD, INR, ISK, NZD, PLN, RUB, SEK, SGD, THB, TWB`

This also introduces the `settings` reducer and `settings` component. This lays the groundwork for futures settings like `autoSend`, `language`, etc.

The `fiatTicker` is stored using `electron-store` and persists for the user.


app/components/Settings/Fiat.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import FaAngleLeft from 'react-icons/lib/fa/angle-left'
+import Isvg from 'react-inlinesvg'
+import checkIcon from 'icons/check.svg'
+import styles from './Fiat.scss'
+
+const Fiat = ({ fiatTicker, fiatTickers, disableSubMenu, setFiatTicker }) => (
+  <div>
+    <header className={styles.submenuHeader} onClick={disableSubMenu}>
```
mrfelton: Could we make the menu close when you click outside of it too, not just on the alias? The UI feels strange when the menu stays open even as you navigate around other parts of the app.

JimmyMow: Added in latest commit

app/components/Form/Request.js
```diff
@@ -72,7 +72,9 @@ const Request = ({
             </div>
           </div>
 
-          <div className={styles.usdAmount}>{`≈ ${requestUsdAmount || 0} USD`}</div>
+          <div className={styles.usdAmount}>{`≈ ${requestUsdAmount || 0} ${
```
mrfelton: style name `usdAmount` should be `fiatAmount`

app/components/Wallet/Wallet.js
```diff
@@ -27,11 +29,12 @@ const Wallet = ({
   currencyName,
   setCurrency,
   setWalletCurrencyFilters,
-  network
+  network,
+  settingsProps
 }) => {
   const usdAmount = btc.satoshisToUsd(
```
mrfelton: `usdAmount` -> `fiatAmount`
`btc.satoshisToUsd` -> `btc.satoshisToFiat`

app/components/Wallet/Wallet.js
```diff
@@ -86,7 +94,8 @@ const Wallet = ({
                 </span>
               </h1>
               <span className={styles.usdValue}>
-                ≈ ${usdAmount ? usdAmount.toLocaleString() : ''}
+                ≈ {currentTicker[ticker.fiatTicker].symbol}
+                {usdAmount ? usdAmount.toLocaleString() : ''}
```
mrfelton: `usdAmount` -> `fiatAmount`

mrfelton: In `btcToUsd` we use `numberWithCommas(amount)` to format fiat amount. We should probably use this everywhere. Maybe we need a general `formatFiatAmount` method that we call when we want to format fiat amounts rather than calling `toLocaleString` 

Alternatively, it looks like the `toLocaleString` method actually has built in support for currencies.

See

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#Parameters

app/lib/utils/btc.js
```diff
@@ -18,7 +25,7 @@ export function btcToBits(btc) {
 
 export function btcToUsd(btc, price) {
```
mrfelton: This method should probably be renamed `btcToFiat`.

app/lib/utils/btc.js
```diff
@@ -18,7 +25,7 @@ export function btcToBits(btc) {
...
   const amount = parseFloat(btc * price).toFixed(2)
-  return (btc > 0 && amount <= 0) ? '< 0.01' : amount.toLocaleString('en')
+  return (btc > 0 && amount <= 0) ? '< 0.01' : numberWithCommas(amount)
```
mrfelton: Rather than calling this, could we use the built in currency support of  `Number.prototype.toLocaleString()` to get us an appropriately formatted amount.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#Parameters

test/unit/reducers/invoice.spec.js
```diff
@@ -11,6 +13,17 @@ import invoiceReducer, {
   INVOICE_FAILED
 } from 'reducers/invoice'
 
+jest.mock('electron', () => {
```
mrfelton: This is probably no longer needed as we have electron mocked by default via `__mocks__` dir

test/unit/reducers/payment.spec.js
```diff
@@ -7,6 +9,17 @@ import paymentReducer, {
   PAYMENT_FAILED
 } from 'reducers/payment'
 
+jest.mock('electron', () => {
```
mrfelton: same here

test/unit/reducers/ticker.spec.js
```diff
@@ -1,10 +1,27 @@
...
+
+import Store from 'electron-store'
 import tickerReducer, {
   SET_CURRENCY,
   SET_CRYPTO,
   GET_TICKERS,
   RECIEVE_TICKERS
 } from 'reducers/ticker'
 
+jest.mock('electron', () => {
```
mrfelton: and here

JimmyMow:
Updated based on feedback @mrfelton

mrfelton:
@JimmyMow Can you please squash these commits into one or more local commits. After that I think we're ready to merge. Thanks!

JimmyMow:
@mrfelton rebased and squashed all un-needed commits

PR #647, opened 2018-08-02
Update flow typed and ensure our code is processed properly

mrfelton:
Update flow packages to latest versions and add `name_mapper` definitions to map our resolved module imports - this ensures that flow runs properly on all of the code under `app`


JimmyMow:
utACK 75e6669

PR #653, opened 2018-08-02
Multilingual support

JimmyMow:
This PR implements `i18n` support in our repo using https://github.com/yahoo/react-intl


mrfelton:
@JimmyMow this is ready to be tested. You can switch languages in two different ways:

1. Using the `Language` menu at the top of the screen
2. using the `Settings` menu within the app.

Default local should be detected by electron. We don't currently store the selected language in the settings file - i'll ad this enhancement in a follow up PR.

I identified quite a lot of strings that were not included in the original extraction so none of the translations are 100% now. Additionally, the structure of the translation files has changed from what we originally uploaded to Crowdin and whilst I have tried where possible to carry over the translations that were submitted to us, it is possible that some did not make it over.

**Note: This builds on #754 #755 and #756 - so those should be merged prior to this one.**

PR #656, opened 2018-08-03
test(runner): reorganise test suites

mrfelton:
Reorganise test suite so that unit tests are cleanly separate from e2e tests. Update tests to make better use of our custom module resolves.


mrfelton:
@JimmyMow the coverage test has failed on this PR because the coverage number has dropped. Builds prior to these fixes were reporting incorrect coverage number.

Here is the latest build on master:
https://coveralls.io/builds/18297991

screenshot 2018-08-04 00 46 19

And here is the build from this PR:
https://coveralls.io/builds/18317001

screenshot 2018-08-04 00 46 23

JimmyMow:
Tested ACK 9962faf

PR #658, opened 2018-08-06
feat(lnd): update BTCD and LND to latest versions

mrfelton:
In an ideal world, lnd would make nightly snapshots available that we could download. However currently they don't do this so I have created our own solution that we can use to compile and publish lnd binaries.

I have created a fork of the lnd repo at https://github.com/LN-Zap/lnd on which I have added support for [goreleaser](https://goreleaser.com/).

You can see the changes that enable this here:
https://github.com/lightningnetwork/lnd/compare/master...LN-Zap:zap/canary-releases?expand=1

With this, we can easily publish a new release with a simple call to `./canary.sh`. This will build lnd for all of the platforms and targets that we support and then create a github release and publish the binaries to it.

For example, here is the release that I have just created from the current master HEAD commit.
https://github.com/LN-Zap/lnd/releases

Then we can use these binaries in conjunction with lnd-binary which is what this PR does.

- [x] Publish updated lnd pre-compiled binary
- [x] Update btcd backend
- [x] Update sync process
- [x] Update grpc proto and api usage
- [ ] Set up DNS name for btcd cluster

**NOTE: This PR should only be merged as part of an update that adds support for the latest lnd grpc interface to zap (update proto + usage)**

**NOTE 2: This upgrade will require users to delete their local lnd data and do a full re-sync**

### DIFF
--- a/app/lib/lnd/neutrino.js
+++ b/app/lib/lnd/neutrino.js
@@ -55,6 +55,14 @@ class Neutrino extends EventEmitter {
       `${this.lndConfig.alias ? `--alias=${this.lndConfig.alias}` : ''}`
     ]
 
+    if (this.lndConfig.network === 'mainnet') {
+      neutrinoArgs.push('--neutrino.connect=mainnet1-btcd.zaphq.io')
+      neutrinoArgs.push('--neutrino.connect=mainnet2-btcd.zaphq.io')
+    } else {
+      neutrinoArgs.push('--neutrino.connect=testnet1-btcd.zaphq.io')
+      neutrinoArgs.push('--neutrino.connect=testnet2-btcd.zaphq.io')
+    }
+
     this.process = spawn(this.lndConfig.binaryPath, neutrinoArgs)
       .on('error', error => this.emit(ERROR, error))
       .on('close', code => {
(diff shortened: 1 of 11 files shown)


app/lib/lnd/neutrino.js
```diff
@@ -55,6 +80,17 @@ class Neutrino extends EventEmitter {
       `${this.lndConfig.alias ? `--alias=${this.lndConfig.alias}` : ''}`
     ]
 
+    switch (this.lndConfig.network) {
```
JimmyMow: Why did you opt for a `switch` statement with only 2 options? Are we expecting this conditional to grow? There is also no default clause here

JimmyMow:
@mrfelton 1 small comment and also don't we wanna change `app/components/Activity/InvoiceModal.js` to display `invoice.value` if the invoice has been settled?

app/lib/lnd/neutrino.js
mrfelton: Good question. Actually at the time I had in my mind about supporting multiple different currencies but even then each one would probably only have testnet and mannet support. So I'll update this to use if/else.

mrfelton:
@JimmyMow I figured it might be better to leave the change to the invoices for a separate PR as that change will actually break the invoice display for people connecting to lnd 0.4.2.

PR #667, opened 2018-08-14
fix(activity): correct tooltips in activity log

### LINKED ISSUE #662: Mouse hover shows the text "Payment Fee" instead of "Fiat X value"
screen shot 2018-08-12 at 1 31 16

mrfelton:
Fix a bug where we were showing transaction amounts as transaction fee.

Fix #662


JimmyMow:
utACK a6439f1

PR #668, opened 2018-08-14
fix(app): quit when window closed on non-darwin

mrfelton:
On darwin, we keep the app running and just hide it when the window is closed as that is the standard convention. On linux and windows the app should be quit when the window is closed.

This resolves an issue where on linux/windows the app would remain open when the window is closed, with no way to properly quit it short of doing a force quit.


JimmyMow:
utACK ddbdf01

Issue #654, opened 2018-08-03
Invoice shows value instead of amt_paid on settled invoice


wraithm:
If I made a payment request with `value = 0` and then the invoice gets paid with a different amount (the value of `amt_paid`), Zap displays the paid amount as `0` ie. `value`, when I would intuitively expect it to display the `amt_paid`.

On line 38 of `app/routes/activity/components/components/Invoice/Invoice.js`: 
```
<Value value={invoice.value} currency={ticker.currency} currentTicker={currentTicker} />
```
I also found a similar value in `app/components/Activity/InvoiceModal.js`.

I'd expect the code to do something like `value={invoice.settled ? invoice.amt_paid : invoice.value}`.

Edit: I started messing around with the code to add this, and I found that `invoice.amt_paid` isn't coming across the `ListInvoices` rpc call. I'm a little confused because that field is definitely described in the [API docs](https://api.lightning.community/#invoice). I have an invoice that shows `amt_paid` when I use `lncli` on the lnd node that I'm connecting to. It's a settled invoice.

mrfelton:
@WraithM the current version of lnd that we bundle doesn't support the new `amt_paid` property. This will be updated as part of  #658

PR #669, opened 2018-08-14
bug(invoice-amt-paid): settled invoices show amt_paid

wraithm:
Settled invoices show the amount paid by the payee instead of the requested
value from the invoice. This involves using an updated version of the lnd rpc
protocol.

This is in response to issue #654.

### DIFF
--- a/app/components/Activity/InvoiceModal.js
+++ b/app/components/Activity/InvoiceModal.js
@@ -10,6 +10,7 @@ import { showNotification } from 'lib/utils/notifications'
 import FaAngleDown from 'react-icons/lib/fa/angle-down'
 
 import Value from 'components/Value'
+import { btc } from 'lib/utils'
 import Countdown from './Countdown'
 
 import styles from './InvoiceModal.scss'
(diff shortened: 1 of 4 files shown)


mrfelton:
@wraithm there is a slight problem with this PR, which is that it will effectively break the invoice screens for users that are running anything other than the most current lnd from the master branch since they will not have access to the new `amt_paid` property. A lot of our users will be using the latest stable version 0.4.2.

Additionally, I think we should try to do this in a way that minimises code duplication. As it is now, you do this check `settled ? x : y` in several different places.

I would suggest that we decorate the invoice object with a custom property that provides the correct value based on the invoice data.

eg:
```javascript
decorateInvoice(invoice) {
  let amount = invoice.value
  if (invoice.amt_paid) {
    amount = btc.millisatoshisToSatoshis(invoice.amt_paid)
  }
  invoice.finalAmount = amount
}
```

We can them pass all retrieved invoices through this method in our invoice reducer.

Then when we need to use the value we can just refer to `invoice.finalAmount` and it will provide the most relevant value.

As an example, refer to the transaction reducer where we do a similar thing: https://github.com/LN-Zap/zap-desktop/blob/master/app/reducers/transaction.js#L30-L34

wraithm:
Sorry for the confusion. I thought I could try to rename the branch and not destroy the PR. Please see #689.

PR #672, opened 2018-08-15
fix(invoice): only notify on settled invoices

### LINKED ISSUE #670: New invoices confuse the client
New invoices confuse the client, sometimes giving the message "congratulations, your invoice was paid", even though I just made a new invoice from another client.  Is it getting confused?

It is saying "settled:  false" in the invoice list in LND.  I have the latest version of LND.

mrfelton:
Fix a bug where we were incorrectly notifying users that an invoice had been successfully paid when it had actually just been created.

Fix #670


JimmyMow:
utACK b1be3f4

Issue #674, opened 2018-08-15
When installing Zap Desktop 0.2.1 Beta Error: LND has unexpectedly quit


d4amenace:
![example2](https://user-images.githubusercontent.com/29969146/44168411-cd9da780-a096-11e8-8d5b-42c4d373d9cb.png)

Same point every time...... Happens just after I unlock wallet.  I had installed an instance of lnd a while back but it is not running..(at least according to task manager).  It shuts itself down before getting to syncing point.

d4amenace:
I am running windows 10.

mrfelton:
@d4amenace please could you try removing your old Zap userdata before starting Zap.

On windows this can be found at `C:\\Users\user\AppData\Roaming\ZapDesktop`

evertvr:
Running Windows 7, having the same issue.

Removed the userdata, but still same error

evertvr:
Had the same error, for me it was some other service was running on 127.0.0.1:8080

skcheffy:
Just installed on Mac and it looks like i am getting the error.  See picture.

<img width="1106" alt="screen shot 2018-08-17 at 7 29 43 pm" src="https://user-images.githubusercontent.com/31999006/44292710-60377580-a254-11e8-8076-154c73347f90.png">

mrfelton:
@skcheffy could you test with the PR at https://github.com/LN-Zap/zap-desktop/pull/683 in place please? This should enhance that error message with some more detail about why specifically lnd exited.

Additionally - do you have something else already running on port `8080` or `90001`?

skcheffy:
Sorry not as tech savvy as I had hoped yet.

No I checked ports and nothing opened there.  How do I test with PR #683 in place?  I went there but not sure exactly what to do.  Do I need to put a file somewherE?

skcheffy:
So I have uninstalled ZAP on my Mac and removed all files associated on my computer via OmniSweep.

Currently re-installed and waiting for sync but no error.  I assume there was something casusing the issue in a folder I found hidden.  

Keep you posted

skcheffy:
It seems like doing the above was successful and worked.

patmillertime:
@JimmyMow @mrfelton do you have any suggestions for resolving the problem I am experiencing above? Would you like me to open a new issue?

mrfelton:
@patmillertime please could you first update to 0.3.2 (released just now) and confirm if the issue persists.

[...]

PR #683, opened 2018-08-17
feat(lnd): show error message if lnd exits

mrfelton:
If lnd unexpectedly exists, show the user the last error message reported by lnd in order to help diagnose issues.

See #674


JimmyMow:
utACK 1b4de6c

PR #684, opened 2018-08-17
 fix(lnd): disable lnd REST interface on port 8080

### LINKED ISSUE #674: When installing Zap Desktop 0.2.1 Beta Error: LND has unexpectedly quit

mrfelton:
Disable the REST interface which by default runs on port 8080 so as to avoid conflicts with users that are using port 8080 for other things. We don't use the REST interface so we don't need to expose it.

Fix #674

### DIFF
--- a/resources/lnd.conf
+++ b/resources/lnd.conf
@@ -84,6 +84,7 @@
 ;   restlisten=localhost:443
 ; On an Unix socket:
 ;   restlisten=unix:///var/run/lnd-restlistener.sock
+restlisten=0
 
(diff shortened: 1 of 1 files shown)


Empact:
utACK f2d093591ebacc950382d98449af71c75ba43eda

JimmyMow:
Tested ACK f2d0935

PR #687, opened 2018-08-18
fix(lnd): update lnd to latest version

mrfelton:
Quite a lot of improvements gone into lnd this last week, including some that help to improve the initial sync process that would benefit us:

See changelog here: https://github.com/LN-Zap/lnd/releases


JimmyMow:
Tested ACK 32345f6

PR #688, opened 2018-08-18
fix(lnd): prevent full chain sync on wallet init

mrfelton:
Do not set `recovery_window: 250` in our call to `initWallet` as this causes lnd to scan the entire chain for non-existent funds which slows the lnd sync process to a crawl.

See https://github.com/lightningnetwork/lnd/issues/1744


JimmyMow:
@mrfelton We should not hardcode `recovery_window: 250` into our `initWallet` call but we do want to use it when users are recovering from seed

mrfelton:
@jackmallers I've updated this PR to ensure that we still increase the recovery window when doing a seed restore and leave it at the default values when setting up a new wallet. On a separate note lnd never completes a sync when doing a seed recovery unless you stop and restart lnd - I'll create a separate issue to look at that.

What is the reason why we want to increase the recovery window when doing a seed restore, and why is the value of `250` chosen?

app/lib/lnd/methods/walletController.js
```diff
@@ -140,15 +140,15 @@ export function unlockWallet(walletUnlocker, { wallet_password }) {
...
+  { wallet_password, cipher_seed_mnemonic, aezeed_passphrase = '', recovery_window }
 ) {
   return new Promise((resolve, reject) => {
     walletUnlocker.initWallet(
       {
         wallet_password: Buffer.from(wallet_password),
         cipher_seed_mnemonic,
         aezeed_passphrase: Buffer.from(aezeed_passphrase, 'hex'),
-        recovery_window: 250
+        recovery_window
```
JimmyMow: no need for this right? since we are not allowing the user to configure this and are hardcoding the value in the `recoverOldWallet ` func

mrfelton: This is needed because we want to allow the caller of `initWallet` to be able to specify what values gets set as `recovery_window`. 

In `createNewWallet` we don't set it. In `recoverOldWallet` we set it to `250`

JimmyMow:
Tested ACK 9ab16e3

PR #689, opened 2018-08-18
fix(invoice-amt-paid): settled invoices show amt_paid

wraithm:
Settled invoices show the amount paid by the payee instead of the requested
value from the invoice. This involves using an updated version of the lnd rpc
protocol.

This is in response to issue #654.

amt_paid is exposed to users via invoice.finalAmount, which is constructed
via a new invoice decorator.

### DIFF
--- a/app/components/Activity/InvoiceModal.js
+++ b/app/components/Activity/InvoiceModal.js
@@ -55,7 +55,7 @@ const InvoiceModal = ({
             <section className={styles.amount}>
               <h1>
                 <Value
-                  value={invoice.value}
+                  value={invoice.finalAmount}
                   currency={ticker.currency}
                   currentTicker={currentTicker}
                 />
--- a/app/routes/activity/components/components/Invoice/Invoice.js
+++ b/app/routes/activity/components/components/Invoice/Invoice.js
@@ -40,11 +40,15 @@ const Invoice = ({ invoice, ticker, currentTicker, showActivityModal, currencyNa
     >
       <span>
         <i className={styles.plus}>+</i>
-        <Value value={invoice.value} currency={ticker.currency} currentTicker={currentTicker} />
+        <Value
+          value={invoice.finalAmount}
+          currency={ticker.currency}
+          currentTicker={currentTicker}
+        />
         <i> {currencyName}</i>
       </span>
       <span>
-        <span>${btc.convert('sats', 'usd', invoice.value, currentTicker.price_usd)}</span>
+        <span>${btc.convert('sats', 'usd', invoice.finalAmount, currentTicker.price_usd)}</span>
       </span>
     </div>
   </div>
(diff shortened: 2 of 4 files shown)


mrfelton:
Tested ACK https://github.com/LN-Zap/zap-desktop/pull/689/commits/b726381a5ab38fe38e17cc3829b5af288bcef366

app/components/Activity/InvoiceModal.js
```diff
@@ -55,7 +55,7 @@ const InvoiceModal = ({
             <section className={styles.amount}>
               <h1>
                 <Value
-                  value={invoice.value}
+                  value={invoice.settled ? invoice.finalAmount : invoice.value}
```
mrfelton: I think the `settled?` check here is redundant.

Can this be simplified to `value={invoice.finalAmount}` since `amt_paid` being set implies that the invoice has been settled.

`invoice.finalAmount` will be either the amount requested, or if a payment has already been made then it will be `amt_paid`.

app/routes/activity/components/components/Invoice/Invoice.js
```diff
@@ -40,11 +40,22 @@ const Invoice = ({ invoice, ticker, currentTicker, showActivityModal, currencyNa
     >
       <span>
         <i className={styles.plus}>+</i>
-        <Value value={invoice.value} currency={ticker.currency} currentTicker={currentTicker} />
+        <Value
+          value={invoice.settled ? invoice.finalAmount : invoice.value}
```
mrfelton: Same goes for this one. The `settled?` check is redundant.

app/routes/activity/components/components/Invoice/Invoice.js
```diff
@@ -40,11 +40,22 @@ const Invoice = ({ invoice, ticker, currentTicker, showActivityModal, currencyNa
...
+        />
         <i> {currencyName}</i>
       </span>
       <span>
-        <span>${btc.convert('sats', 'usd', invoice.value, currentTicker.price_usd)}</span>
+        <span>
+          ${btc.convert(
+            'sats',
+            'usd',
+            invoice.settled ? invoice.finalAmount : invoice.value,
```
mrfelton: Here too.

wraithm:
Okay, I fixed the redundant checks. You were absolutely right about those. Good catch.

mrfelton:
Thanks @wraithm - could you squash your commits into a single one please? Then I think this will be ready to merge.

wraithm:
Oh boy...

JimmyMow:
Haven't tested this but are we sure LND returns `finalAmount` in `millisatoshis`? Last time I checked they don't expose `millisatoshis` at the API level, only internally

mrfelton:
Whilst it is not documented in the lnd docs or code comments, I have tested this and verified that it does indeed use milisatoshis

wraithm:
Yeah, I was also surprised by milisatoshis on amt_paid. It isn't
documented, but that's definitely what it is.

PR #690, opened 2018-08-19
fix(lnd): only connect to single btcd host

mrfelton:
The lnd sync process doesn't currently handle multiple btcd connections well. Remove the alternate btcd hosts for testnet and mainnet.

See https://github.com/lightningnetwork/lnd/issues/1752


JimmyMow:
utACK d13c4f2

PR #691, opened 2018-08-19
fix(lnd): don't treat filters as filter headers

mrfelton:
During the sync process there are log message that relate to syncing filters and filter headers. This fixes an issue where we were parsing messages about syncing filters as if they were filter headers, which would cause the progress indicator to jump around during the middle of the sync process due to it thinking that more filter headers had been synced than was actually the case.


JimmyMow:
utACK d53335a

PR #693, opened 2018-08-20
fix(ux): show loading bolt while seed is generated

mrfelton:
Ensure that the loading bold graphic shows whist the seed for a new wallet is being generated. This provides a smoother user experience and prevents users from being able to initiate actions through the UI when they shouldn't be able to.


JimmyMow:
Tested ACK 1e60542

PR #695, opened 2018-08-21
fix(wallet): fetch updated channels and balance after a new transaction

### LINKED ISSUE #579: Inaccurate wallet balance after autopilot opens channels
When autopilot opens channels for my node it does not refresh my wallet balance. Sometimes I try and send on-chain funds and Zap tells me I don't have sufficient balance. I then look and realize autopilot opened a few channels for me and the funds have gone to opening channels but my wallet balance doesn't reflect that

mrfelton:
Make a call to fetch the updated balance and channels list whenever a new transaction is detected.

Fix #579


JimmyMow:
Tested ACK b3044b2

PR #698, opened 2018-08-21
feat(autopilot): enable private and minconfs=0

### LINKED ISSUE #681: Don’t advertise channels
Hey guys 🙂

sorry to barge in and congrats on the recent release!

@roasbeef and the lnd team added a `--autopilot.private` flag. Since end user hosts running apps are likely not going have the same uptime/availability as routing nodes running on servers it might be a good idea to set the flag for channels openend automatically by autopilot.

Also, I haven’t looked into the code but there is also a `private` flag for opening channel via grpc: https://api.lightning.community/#openchannel

We’re planning to add the flag in our upcoming release of the Lightning App as well. If all user agents do this it will hopefully improve routing node reliability.

Thanks again and keep up the great work!

JimmyMow:
This PR adds private channels and minconfs=0 to our autopilot configuration.

We don't want our light client users advertising channels because they are not likely to be reliable nodes with high uptime. Fixes #681.

**NOTE: This PR can only be merged _after_ we update to LND master. These new flags aren't supported with the LND that is currently packaged and LND will shut down on start because of it at the moment. (Customized my binary path to test this works)**

### DIFF
--- a/resources/lnd.conf
+++ b/resources/lnd.conf
@@ -317,6 +317,11 @@ litecoin.node=ltcd
 ; amount of attempted channels will still respect the maxchannels param.
 ; autopilot.allocation=0.6
 
+; If the user toggled autopilot on let's make sure channels are private and allow
+; it to use 0 conf txs
+autopilot.private=1
+autopilot.minconfs=0
+
 [tor]
 ; The port that Tor's exposed SOCKS5 proxy is listening on. Using Tor allows
 ; outbound-only connections (listening will be disabled) -- NOTE port must be


app/lib/lnd/neutrino.js
```diff
@@ -78,10 +78,20 @@ class Neutrino extends EventEmitter {
...
       `${this.lndConfig.alias ? `--alias=${this.lndConfig.alias}` : ''}`
     ]
 
+    // If the user toggled autopilot on let's enable it,
+    // make sure channels are private and allow it to use 0 conf txs
+    if (this.lndConfig.autopilot) {
+      // Turn autopilot on
+      neutrinoArgs.push('--autopilot.active')
+      // Ensure all channels opened by autopilot will be flagged private
+      neutrinoArgs.push('--autopilot.private')
```
mrfelton: should we move these additional options into `resources/lnd.conf`? I think the only thing we need to specify dynamically here is the `--autopilot.active` switch.

app/lib/lnd/neutrino.js
```diff
@@ -90,6 +100,7 @@ class Neutrino extends EventEmitter {
       // neutrinoArgs.push('--neutrino.connect=testnet2-btcd.zaphq.io')
     }
 
+    // this.process = spawn(this.lndConfig.binaryPath, neutrinoArgs)
```
mrfelton: this commented line can go

mrfelton:
@JimmyMow this needs another update to lnd binary, over and above the one that was merged this morning?

JimmyMow:
@mrfelton no the current LND version is fine. 

Squashed commits and moved `autopilot.private=1` and `autopilot.minconfs=0` to the `proto` file

PR #699, opened 2018-08-21
fix(channels): neutrino clients manual channels flagged private

JimmyMow:
This PR we make sure that if a user of ours is using neutrino that we flag their channels as private when they manually open them. Reason for this is because light consumer clients are not expected to have high uptime. 

However we cannot hardcode channels as private every time because we do support more power users driving their remote nodes or BTCPay Server with Zap. So we only flag channels as private if the `activeConnection` is `local`.

Addresses #681

### DIFF
--- a/app/lib/lnd/methods/channelController.js
+++ b/app/lib/lnd/methods/channelController.js
@@ -19,13 +19,14 @@ function ensurePeerConnected(lnd, pubkey, host) {
  * @return {[type]}         [description]
  */
 export function connectAndOpen(lnd, event, payload) {
-  const { pubkey, host, localamt } = payload
+  const { pubkey, host, localamt, private: privateChannel } = payload
 
   return ensurePeerConnected(lnd, pubkey, host)
     .then(() => {
       const call = lnd.openChannel({
         node_pubkey: Buffer.from(pubkey, 'hex'),
-        local_funding_amount: Number(localamt)
+        local_funding_amount: Number(localamt),
+        private: privateChannel
       })
 
       call.on('data', data => event.sender.send('pushchannelupdated', { pubkey, data }))
--- a/app/reducers/channels.js
+++ b/app/reducers/channels.js
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect'
 import { ipcRenderer } from 'electron'
+import Store from 'electron-store'
 import { btc } from 'lib/utils'
 import { showNotification } from 'lib/utils/notifications'
 import { requestSuggestedNodes } from 'lib/utils/api'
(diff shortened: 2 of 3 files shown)


tanx:
👍

.flowconfig
```diff
@@ -9,6 +9,7 @@
 <PROJECT_ROOT>/dll/.*
 <PROJECT_ROOT>/release/.*
 <PROJECT_ROOT>/git/.*
+<PROJECT_ROOT>/.git/.*
```
mrfelton: Why is this needed?

app/lib/lnd/methods/channelController.js
```diff
@@ -19,13 +19,14 @@ function ensurePeerConnected(lnd, pubkey, host) {
  * @return {[type]}         [description]
  */
 export function connectAndOpen(lnd, event, payload) {
-  const { pubkey, host, localamt } = payload
+  const { pubkey, host, localamt, private: privateChannel } = payload
```
mrfelton: Why do we need to change the variable name to `privateChannel`? Can't it just remain named `private`? especially as that is what we want it to be named when we pass it into `lnd.openChannel` below.

app/reducers/channels.js
```diff
@@ -191,7 +192,18 @@ export const openChannel = ({ pubkey, host, local_amt }) => (dispatch, getState)
...
   dispatch(addLoadingPubkey(pubkey))
 
-  ipcRenderer.send('lnd', { msg: 'connectAndOpen', data: { pubkey, host, localamt } })
+  // Grab the activeConnection type from our local store.
+  // If the active connection type is local (light clients using neutrino)
+  // we will flag manually created channels as private. Other connections
+  // like remote node and BTCPay Server we will announce to the network as
+  // these users are using Zap to drive nodes that are online 24/7
+  const store = new Store({ name: 'settings' })
+  const { type } = store.get('activeConnection', null)
```
mrfelton: If `activeConnection` is not set in the store, `store.get` should return an empty object, instead of `null`
 
```javascript
const { type } = store.get('activeConnection',{})
```

Otherwise, we would be trying to destructure `null` which would cause an error.

flow-typed/module_vx.x.x.js
```diff
@@ -1,3 +1,3 @@
-declare module 'module' {
+declare module 'grpc' {
```
mrfelton: We should probably create a new file for this rather than modifying the existing flowtype for `module`

test/unit/reducers/channels.spec.js
```diff
@@ -8,6 +10,16 @@ import channelsReducer, {
   OPENING_FAILURE
 } from 'reducers/channels'
 
+jest.mock('electron', () => {
```
mrfelton: This shouldn't be needed any longer as we've added a manual mock for electron here: https://github.com/LN-Zap/zap-desktop/blob/master/test/unit/__mocks__/electron.js

app/reducers/channels.js
```diff
@@ -191,7 +192,18 @@ export const openChannel = ({ pubkey, host, local_amt }) => (dispatch, getState)
   dispatch(openingChannel())
   dispatch(addLoadingPubkey(pubkey))
 
-  ipcRenderer.send('lnd', { msg: 'connectAndOpen', data: { pubkey, host, localamt } })
+  // Grab the activeConnection type from our local store.
```
mrfelton: nit: Line length should be 120.

mrfelton:
@JimmyMow I've created a separate PR to address the issues with flow: https://github.com/LN-Zap/zap-desktop/pull/702. That can be removed from this PR.

flow-typed/module_vx.x.x.js
JimmyMow: going to remove this as this was fixed in a diff PR and merged

app/lib/lnd/methods/channelController.js
JimmyMow: Because private is a reserved word in strict mode

test/unit/reducers/channels.spec.js
JimmyMow: 👍

mrfelton:
Tested ACK https://github.com/LN-Zap/zap-desktop/pull/699/commits/9b837ba316133172389e8a338bfc6cb1055a9db1

Issue #676, opened 2018-08-15
Make it more clear to users which network they are connecting/connected to (testnet vs mainnet)


rashbit:
Installing Zap on Win10 Pro. Right after the install, created the wallet and suggested to fund it. Bitcoin address (both QR and text) are invalid bitcoin address. This: 2MtYDfkWfWvzia1oA5M8mdxprHyRi97bka1

mrfelton:
@rashbit note that when you are using the `default` mode the app is connected to the bitcoin testnet. The address you see is a valid testnet address.

You can send some testnet bitcoins to your wallet from https://testnet.manu.backend.hamburg/faucet

If you want to use the wallet on mainnet you will need to setup your own bitcoin+lnd fullnode and then use the  `custom` option when starting zap to connect to your lnd node.

rashbit:
Thanks. Perhaps a note on the screen explaining that could be helpful for new users. I guess many just give up trying...

mrfelton:
Note that this is just a temporary situation. As soon as lnd is ready for usage on the mainnet we will default our app to connecting to the mainet. Until then we are defaulting to the testnet.

But that said, I agree that this could be made clearer to users so I'm going to reopen this story and re-purpose as a feature request to add some help text within the app to make this more clear for users.

Thanks for the feedback @rashbit

ZacharyATanenbaum:
In addition to the pill I would suggest adding some text next to the default option. Stating that Zap will connect to the testnet by default.

songproducer:
This is still an issue on the latest Mac client  (Version 0.3.2-beta (0.3.2-beta.5107))

PR #700, opened 2018-08-21
feat(testnet): add pill to UI on testnet

JimmyMow:
Add a Testnet pill to the UI next to our logo when a user is on testnet. Helps address #676 

screen shot 2018-08-21 at 6 19 33 pm

### DIFF
--- a/app/components/Wallet/Wallet.js
+++ b/app/components/Wallet/Wallet.js
@@ -49,6 +49,7 @@ const Wallet = ({
         <header className={styles.header}>
           <section className={styles.logo}>
             <Isvg className={styles.bitcoinLogo} src={zapLogo} />
+            {info.data.testnet && <span className={styles.testnetPill}>Testnet</span>}
           </section>
 
           <section className={styles.user}>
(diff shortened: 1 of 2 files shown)


odb366:
The pill diverts users attention to it way too much. In my opinion, we should use text only without it. The pill could also be mistaken for a button.

mrfelton:
A agree it's possibly a little heavy. Perhaps just a green outline rather than a solid background colour?

JimmyMow:
I got the design from our mobile app.

Changed the text color to match the rest of our wallet header and only made the border green. Thoughts @odb366 @mrfelton?

screen shot 2018-08-22 at 9 18 46 am

JimmyMow:
Updated

odb366:
@JimmyMow @mrfelton Now its good that it doesn't draw so much attention. However, I've shown the picture to the gf and 1 friend and they both thought that this was a button. I think we should use text-only so users don't mistake it for a button.

JimmyMow:
@odb366 @mrfelton updated the UI:

screen shot 2018-09-04 at 10 32 13 am

mrfelton:
LGTM

PR #701, opened 2018-08-22
fix(wallet): ensure clean state after refresh

mrfelton:
## Problem / Motivation

Start the app, hit `cmd+r` twice, then try to progress through the onboarding process. You will get a `transition is invalid in current state` error.

## Proposed Solution

The `onOnboarding` state transition handler only fires if you are entering the `onboarding` state from another state. If you are already in the `onboarding` state it does not fire when the `startOnboarding`
event is initiated. By contrast, `onStartOnboarding` always fires when `startOnboarding` is run.

Ensure that IPC listeners are only registered once when transitioning to the `onboarding` state from another state.

### DIFF
--- a/app/lib/zap/controller.js
+++ b/app/lib/zap/controller.js
@@ -130,8 +130,11 @@ class ZapController {
   onOnboarding() {
     mainLog.debug('[FSM] onOnboarding...')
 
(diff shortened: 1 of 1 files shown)


JimmyMow:
Tested ACK 92e2dc5

PR #703, opened 2018-08-22
chore(deps): update jest to v23.5.0

mrfelton:
This resolves an issue with jest that was preventing tests from completing on windows environments and is the reason that lots of our builds on appveyor have been failing.

See facebook/jest#6639


JimmyMow:
Concept ACK ac3ee5f

PR #707, opened 2018-08-23
fix(lnd): ensure graceful shutdown

mrfelton:
Lnd doesn't currently do a graceful shutdown when a `SIGTERM` is issued. To ensure a graceful shutdown, call `stopDaemon` on the gRPC interface prior to terminating lnd.

See
 - https://github.com/lightningnetwork/lnd/issues/1028
 - https://api.lightning.community/#stopdaemon

### DIFF
--- a/app/lib/lnd/lightning.js
+++ b/app/lib/lnd/lightning.js
@@ -95,6 +95,21 @@ class Lightning {
   }
 
   /**
+   * Gracegfully shutdown the gRPC service.
+   */
+  shutdown() {
+    this.unsubscribe()
+    return new Promise((resolve, reject) => {
+      this.lnd.stopDaemon({}, (err, data) => {
+        if (err) {
+          reject(err)
+        }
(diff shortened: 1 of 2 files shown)


app/lib/lnd/lightning.js
```diff
@@ -94,6 +94,21 @@ class Lightning {
     }
   }
 
+  /**
+   * Gracegfully shutdown the gRPC service.
+   */
+  shutdown() {
+    this.unsubscribe()
+    return new Promise((resolve, reject) => {
+      this.lnd.stopDaemon({}, (err, data) => {
```
JimmyMow: can we get a `log` here or somewhere so it's easier to see this gets called? Tough to test/debug. Only relevant log is `Shutting down lightning Wallet...`. Would be nice to see a `Calling lnd method: { msg: 'stopDaemon', data: undefined }` or something like that

PR #710, opened 2018-08-25
feat(errors): create a util for making error messages more user-friendly

kaloudis:
screen shot 2018-08-24 at 11 45 45 pm

I know the lnd team is working on improving their error messages but I think zap would really benefit by having its own set of user-friendly error messages. You may understand what `Error: 11 OUT_OF_RANGE: EOF` means if you've been hands on with `lnd`, but the average user is left confused by seeing that after the failure of a channel connection - I sure was.

### DIFF
--- a/app/components/GlobalError/GlobalError.js
+++ b/app/components/GlobalError/GlobalError.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import MdClose from 'react-icons/lib/md/close'
+import errorToUserFriendly from 'lib/utils/userFriendlyErrors'
 import styles from './GlobalError.scss'
 
 class GlobalError extends React.Component {
@@ -20,7 +21,7 @@ class GlobalError extends React.Component {
           <div className={styles.close} onClick={clearError}>
             <MdClose />
           </div>
-          <h2>{error}</h2>
+          <h2>{errorToUserFriendly(error)}</h2>
         </div>
       </div>
     )
--- a/app/lib/utils/userFriendlyErrors.js
+++ b/app/lib/utils/userFriendlyErrors.js
@@ -1,0 +1,9 @@
+const userFriendlyErrors = {
+  'Error: 11 OUT_OF_RANGE: EOF':
+    "The person you're trying to connect to isn't available or rejected the connection.\
+ Their public key may have changed or the server may no longer be responding."
+}
+
+const errorToUserFriendly = error => userFriendlyErrors[error] || error
+
+export default errorToUserFriendly
--- a/app/components/GlobalError/GlobalError.scss
+++ b/app/components/GlobalError/GlobalError.scss
@@ -35,7 +35,8 @@
       max-width: 75%;
       margin-left: auto;
       margin-right: auto;
-      font-size: 20px;
+      font-size: 15px;
+      line-height: 20px;
       letter-spacing: 1.5px;
       font-weight: bold;
     }
--- a/test/unit/utils/userFriendlyErrors.spec.js
+++ b/test/unit/utils/userFriendlyErrors.spec.js
@@ -1,0 +1,17 @@
+import errorToUserFriendly from 'lib/utils/userFriendlyErrors'
+
+describe('userFriendlyErrors', () => {
+  describe('errorToUserFriendly', () => {
+    it('should handle defined user-friendly errors', () => {
+      expect(errorToUserFriendly('Error: 11 OUT_OF_RANGE: EOF')).toBe(
+        "The person you're trying to connect to isn't available or rejected the connection.\
+ Their public key may have changed or the server may no longer be responding."
+      )
+    })
+
+    it('should return the original error when there is no user-friendly error conversion', () => {
+      expect(errorToUserFriendly('Error 12')).toBe('Error 12')
+      expect(errorToUserFriendly('???')).toBe('???')
+    })
+  })
+})


app/components/GlobalError/GlobalError.js
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import MdClose from 'react-icons/lib/md/close'
 import styles from './GlobalError.scss'
+import errorToUserFriendly from '../../lib/utils/userFriendlyErrors'
```
JimmyMow: change to `import errorToUserFriendly from 'lib/utils/userFriendlyErrors'`

app/components/GlobalError/GlobalError.js
```diff
@@ -13,14 +14,15 @@ class GlobalError extends React.Component {
 
   render() {
     const { error, clearError } = this.props
+    const errorMsg = errorToUserFriendly(error)
```
JimmyMow: why do we need to declare this const here? we can probably just render `errorToUserFriendly(error)` right?

app/lib/utils/userFriendlyErrors.js
```diff
@@ -0,0 +1,15 @@
+const userFriendlyErrors = {
+  'Error: 11 OUT_OF_RANGE: EOF':
+    "The person you're trying to connect to isn't available or rejected the connection.\
+ Their public key may have changed or the server may no longer be responding."
+}
+
+const errorToUserFriendly = error => {
```
JimmyMow: this func can be a 1 liner and much simpler:

`const errorToUserFriendly = error => userFriendlyErrors[error] || error`

JimmyMow:
Nice, thanks for the PR @kaloudis. Added some comments. Also I think we should improve the UI of the error message in this PR if we are going to merge this. Smaller font size, space between lines, etc

app/components/GlobalError/GlobalError.js
kaloudis: I'm not sure that'll work but I'll try it

JimmyMow:
Also can you squash your commit and follow our contributing guide found here: https://github.com/LN-Zap/zap-desktop/blob/master/CONTRIBUTING.md#committing-and-pushing-changes

kaloudis:
@JimmyMow Thanks for the comments. Totally agree on the styling -- I'll work on it

Perhaps squash merging should be enabled on the repo? Check out https://blog.github.com/2016-04-01-squash-your-commits/ -- regardless I'll amend the commits myself this time around

kaloudis:
New GlobalError styling:
screen shot 2018-08-25 at 4 17 31 pm
screen shot 2018-08-25 at 4 18 16 pm

mrfelton:
@kaloudis conceptually I really like the idea behind this PR. However, I think the current implementation may be a little problematic because we a relying on a stringified error message to do our matching where as we should really be using `error.code`. Lnd's error messages are subject to change, where as their error codes should not be.

To address this we would need to refactor slightly how we capture and pass lnd error messages through to the app.

```javascript
.catch(error => {
  log.error('addInvoice:', error)
  event.sender.send('invoiceFailed', { error: error.toString() })
})
```
https://github.com/LN-Zap/zap-desktop/blob/master/app/lib/lnd/methods/index.js#L169-L172

Code like the above should be changed to pass the full error object through So that we have access to all of the original error properties, including `error.code`:
```javascript
event.sender.send('invoiceFailed', { error })
```
This would need to be done throughout all of the code in `/lib/lnd` and then the we would need to adjust the app accordingly.

mrfelton:
So I did a little more digging, and unfortunately it seems that Lnd's error handling is not great. They don't seem to have well defined error codes or error type information that gets propagated through the gRPC interface.

As an example, here is the error object returned from lnd when you try to create an invoice that is too large

```javascript
{ Error: 2 UNKNOWN: payment of 1000 BTC is too large, max payment allowed is 0.04294967 BTC
    at Object.exports.createStatusError (/Users/tom/workspace/zap-desktop/app/node_modules/grpc/src/common.js:87:15)
    at Object.onReceiveStatus (/Users/tom/workspace/zap-desktop/app/node_modules/grpc/src/client_interceptors.js:1188:28)
    at InterceptingListener._callNext (/Users/tom/workspace/zap-desktop/app/node_modules/grpc/src/client_interceptors.js:564:42)
    at InterceptingListener.onReceiveStatus (/Users/tom/workspace/zap-desktop/app/node_modules/grpc/src/client_interceptors.js:614:8)
    at callback (/Users/tom/workspace/zap-desktop/app/node_modules/grpc/src/client_interceptors.js:841:24)
  code: 2,
  metadata: Metadata { _internal_repr: {} }
```

`error.code` is `2` (the same as used on many other unrelated errors) and there here is nothing on this error object that makes easy distinguishable from any number of other errors with a code of `2` (other than the message itself).

It seems that the `UNKNOWN` error code `2` actually comes from `node-grpc` and is there precisely because lnd does not return enough error information:
```javascript
/**
 * Unknown error.  An example of where this error may be returned is
 * if a status value received from another address space belongs to
 * an error-space that is not known in this address space.  Also
 * errors raised by APIs that do not return enough error information
 * may be converted to this error.
 */
UNKNOWN: 2,
```
https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/src/constants.js#L35-L42

So with that in mind, matching on the error message is perhaps the best we can do for the time being :/

kaloudis:
@mrfelton thanks for investigating. I agree, matching on error code is optimal. Perhaps we can refactor down the line as error handling in lnd improves

PR #714, opened 2018-08-26
feat(lnd): update lnd to v0.4.2-beta-913-g26f68da5

mrfelton:
This updates LND to the latest commit from master. This was incompatible with the version of BTCD that we were running so I have also updated `testnet1-btcd.zaphq.io` and `testnet2-btcd.zaphq.io` to the latest version of BTCD (which got several updates in the last week). I verified first that this new version of BTCD was still compatible with the old version of lnd that we were previously running so as not to impact users of 0.2.1.

### DIFF
--- a/app/lib/lnd/config.js
+++ b/app/lib/lnd/config.js
@@ -113,7 +113,7 @@ class LndConfig {
   // Read only data properties.
   +key: string
   +binaryPath: string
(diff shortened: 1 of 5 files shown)


JimmyMow:
niiiice 🎊. updated the macaroon path too.

Tested ACK 853c694

Issue #666, opened 2018-08-14
Text too close to the bottom of the window on onboarding syncing screen (linux)


mrfelton:
As you can see here, the text is too close to the bottom of the window on the onboarding screen on linux:

<img width="1010" alt="screenshot 2018-08-14 11 45 27" src="https://user-images.githubusercontent.com/200251/44084808-0797e7d2-9fb8-11e8-9a2c-9521d2dd9962.png">

PR #716, opened 2018-08-26
fix(syncing): add space to bottom of sync page

JimmyMow:
Fix for #666


JimmyMow:
@mrfelton ok this can be closed. i was just scrolling through the open issues looking for things to fix. let's close #666 then?

mrfelton:
Yes, I think so. I like this approach because it ensures that we have the exact same pixel height to work with on both platforms rather than leaving it up to guess work

mrfelton:
... I mean to merge the fix from #708 and close this one

PR #719, opened 2018-08-27
fix(lnd): ensure graceful shutdown

mrfelton:
Lnd doesn't currently do a graceful shutdown when a `SIGTERM` is issued. To ensure a graceful shutdown, call `stopDaemon` on the gRPC interface prior to terminating lnd. If a graceful shutdown is unsuccessful then force kill the daemon.


JimmyMow:
Tested ACK 133e34d

Issue #715, opened 2018-08-26
payInvoice Timeout


JimmyMow:
Payment routing isn't optimized yet and error handling hasn't gotten too much attention. This results in stale channels and nodes disrupting paying an invoice and `payInvoice` hanging forever. To patch the UX before our next release we should add some timeout to `payInvoice` and a ticker next to the sending animation so the user knows after _x_ amount of time we'll cal it quits

JimmyMow:
closing for #721

Issue #637, opened 2018-07-31
Stuck in "Sending your transaction"


SDWouters:
I tried to buy some coffees from the Acinq store through the Lightning testnet, with a funded channel to Yalls.org (02212d3ec8) and another to Endurance (03933884aa). When I send a transaction, it infinitely remains stuck in the loading animation "Sending your transaction".

I'm running 0.2.0-beta on Mac.

Displaying the problem in the GIF below, plus another visual bug where refreshing leads to old transactions changing from Received to sent and + to -

![2018-07-31 16 23 34](https://user-images.githubusercontent.com/6563736/43465734-3d5eae12-94de-11e8-8d6c-235e7508cf32.gif)

JimmyMow:
closing for #721

PR #721, opened 2018-08-27
feat(payment): add 60 second payment timeout

JimmyMow:
In this commit we add a 60 second timeout when paying a LN invoice. If the payment doesn't successfully route after 60 seconds we display an error. 

Addresses #715, #637 

New payment success UX:

ezgif com-video-to-gif 2

New payment fail UX:

ezgif com-video-to-gif 3

### DIFF
--- a/app/components/Wallet/Wallet.js
+++ b/app/components/Wallet/Wallet.js
@@ -30,7 +30,8 @@ const Wallet = ({
   setCurrency,
   setWalletCurrencyFilters,
   network,
-  settingsProps
+  settingsProps,
+  paymentTimeout
 }) => {
   const fiatAmount = btc.satoshisToFiat(
     parseInt(balance.walletBalance, 10) + parseInt(balance.channelBalance, 10),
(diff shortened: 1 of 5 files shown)


mrfelton:
I was interested to know how this would handle making 2 payments in quick succession... This screenshot shows the result:

screenshot 2018-08-27 22 22 53

In my testing I had to be pretty quick on the keyboard to get a second payment in whilst the first was still processing (or shortly after it processed in my case I think), but I can imagine that if a user could be waiting for up to 60 seconds for a first payment to go through or timeout it's quite feasible that they might try to pay another invoice within that 60 second window.

In my example, you can see that the "Successfully sent payment" message from the 1st payment is still on screen and kind of in the way of the "Sending your transaction" message from the 2nd payment.

Possibly a bit of an edge case though, and I think what this PR delivers is a good improvement on the current situation. To solve fort that we might want to think about implementing a global task log/queue and some nice way to display that but that would be outside of the scope of this PR.

If you can find a quick fix fot that display quirk that would be good but I don't see it as a blocker for this.

mrfelton:
Tested https://github.com/LN-Zap/zap-desktop/pull/721/commits/948d44809796a4da80f56b16eed108918ee1754d ACK

Tests failing though

JimmyMow:
I'll fix the tests.

Agree that a robust solution is out of the scope for this PR but I'll try and see if I have any CSS tricks to make things a bit better for the time being

Issue #725, opened 2018-08-29
Light client invoices don't include private routing hints


JimmyMow:
## Description

Anyone using Zap + Neutrino right now cannot currently accept payments because all channels for them are private and our invoices don't include private routing hints

## Expected Behavior

New invoices include private routing hints for light clients

## Actual Behavior

No private routing hints for light client invoices

## Possible Fix

Utilize the private bool in https://api.lightning.community/#addinvoice

PR #726, opened 2018-08-29
feat(invoice): add priv routing hints for light clients

JimmyMow:
We currently do not flag for private channel routing hints when we create invoices. We need to if the user is a light client because all channels will be private. This commit checks which type of client is using zap and flags invoices to add private hints if it's a light client.

Fix for #725


mrfelton:
Tested ACK https://github.com/LN-Zap/zap-desktop/pull/726/commits/545285f0c3130af7c850096c7af9f13c6a149e82

PR #728, opened 2018-08-29
fix(grpc): use Numbers to represent long values

### LINKED ISSUE #727: Invoice amount always shows as 0 on invoice detail page
## Description

Invoices are showing an incorrect amount in the invoice detail view. The show the correct amount in the activity list.

## Expected Behavior

The invoice amount should be correct when you click on an invoice to view the detail.

## Actual Behavior

Unpaid invoices always show `0` for the amount.

## Steps to Reproduce

- 1. Create an invoice
- 2. From the activity list, click on the invoice
- 3. Notice that the amount shows as 0

## Your Environment

- Zap Desktop version: master
- Operating System and version: OSX

mrfelton:
Configure gRPC to use Numbers to represent long values instead of Strings. This ensures that we can do accurate handling of numerical values.

## Description:

This ensures that numbers are actually numbers and not strings so a check like `if (val) { ... }` would will give expected results.

For example:
```
if (0) { echo "something" }
   => undefined
```
but 
```
if ("0") { echo "something" }
    => "something"
```

(`"0"` is a truthy value, whilst `0` is not)

## Motivation and Context:

Fix #727

## Types of changes:

Bux fix

## Checklist:

### DIFF
--- a/app/lib/lnd/lightning.js
+++ b/app/lib/lnd/lightning.js
@@ -71,7 +71,7 @@ class Lightning {
       // See https://github.com/grpc/grpc-node/blob/master/packages/grpc-protobufjs/README.md
       const options = {
         keepCase: true,
-        longs: String,
+        longs: Number,
         enums: String,
         defaults: true,
         oneofs: true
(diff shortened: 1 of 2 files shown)


JimmyMow:
utACK 47b95a0

PR #733, opened 2018-08-31
feat(channels): warn when opening multiple channels to the same node

### LINKED ISSUE #252: Needs a warning / notification before allowing you to open a double channel
When opening a new channel, the app should check the peer or alias of the new node against the list of nodes you already have opened channels with and notify you before opening a double channel (idiot-proof is good, right?)

Also maybe check if the channel is still funded -> so the notification window can remind the user of the funds they already have available in that channel or alternatively, check to see if there's a path available for a payment to that node from another already funded lightning channel -> so the notification window could show all this payment options before allowing the user to open a new channel (especially a double channel).

Could be very useful for users who might think they need to open a new channel to pay each merchant, more so as the network is still growing and paths will be harder to find in the beginning.

jamaljsr:
## Description:

Display a warning when the user attempts to open a channel to a node that already has active channels open. Also display the node alias and total capacity of the existing channels in the warning message. The total capacity units automatically change when the user changes the unit in the dropdown menu on this screen.

## Motivation and Context:

Fixes #252 

## How Has This Been Tested?

Attempted to open channels to a node in each of these scenarios:
- node with no existing channels open. the warning is not displayed
- node with multiple open channels. the warning shows with the number of open channels, the alias and total capacity
- node with an open channel but no alias: the warning shows with the total capacity but no alias

image

image

## Types of changes:

Minor UI improvement

## Checklist:

### DIFF
--- a/app/components/Contacts/SubmitChannelForm.js
+++ b/app/components/Contacts/SubmitChannelForm.js
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 
 import FaAngleDown from 'react-icons/lib/fa/angle-down'
+import FaExclamationCircle from 'react-icons/lib/fa/exclamation-circle'
 
 import AmountInput from 'components/AmountInput'
 import styles from './SubmitChannelForm.scss'
@@ -17,6 +18,7 @@ class SubmitChannelForm extends React.Component {
       updateContactCapacity,
       openChannel,
       fiatTicker,
+      dupeChanInfo,
 
       ticker,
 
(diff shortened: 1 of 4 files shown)


mrfelton:
@jamaljsr I'm seeing some errors in the console when accessing the add channel form for the first time after starting up the wallet. On subsequent attempts I don't see the error.

```
warning.js:33 Warning: Failed prop type: The prop `dupeChanInfo` is marked as required in `SubmitChannelForm`, but its value is `null`.
    in SubmitChannelForm (created by ChannelForm)
    in ChannelForm (created by App)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App) (created by Route)
    in Route (created by withRouter(Connect(App)))
    in withRouter(Connect(App)) (created by routes)
    in routes (created by Root)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by Root)
    in Provider (created by Root)
    in Root (created by Connect(Root))
    in Connect(Root)
    in e
```

jamaljsr:
@mrfelton Ah, good catch. I'll get it fixed today.

jamaljsr:
@mrfelton I fixed the console error and retested all scenarios. This should be good to go now.

PR #734, opened 2018-09-02
fix(ui): show wallet once balances are loaded

mrfelton:
## Description:

Due to a recent change in the way that we handle numeric data from gRPC we need to be explicit when checking for unset/null values.

## Motivation and Context:

Wallet GUI missing for new wallets with a zero balance.

## How Has This Been Tested?

Create and sync new wallet and verify that the UI shows correctly.

## Types of changes:

Bug fix

## Checklist:


JimmyMow:
utACK a63bbac

PR #748, opened 2018-09-06
feat(autopilot): use new routing flag

JimmyMow:
This PR adds the new experimental routing flag that will improve the efficiency of autopilot and improve the overall on-boarding UX as autopilot will now start searching for channels before the headers are synced

## Description:

Added the new routing flag

## Motivation and Context:

Autopilot will open channels quicker and users will be on-boarded more seamlessly 

## How Has This Been Tested?

Use the new routing flag with autopilot

## Types of changes:

Push a new flag into the neutrino args
## Checklist:

### DIFF
--- a/resources/lnd.conf
+++ b/resources/lnd.conf
@@ -338,3 +338,8 @@ autopilot.minconfs=0
 ; This means that multiple applications (other than lnd) using Tor won't be mixed
 ; in with lnd's traffic.
 ; tor.streamisolation=1
+
+[routing]
+; Use the experimental routing config assumechanvalid
+
+routing.assumechanvalid=1


app/lib/lnd/neutrino.js
```diff
@@ -90,6 +90,10 @@ class Neutrino extends EventEmitter {
       // neutrinoArgs.push('--neutrino.connect=testnet2-btcd.zaphq.io')
     }
 
+    // Use the experimental routing config assumechanvalid
+    //https://github.com/lightningnetwork/lnd/blob/309e656a979f22bf53fa2c85544fc2ee9faa10fd/routing/conf_experimental.go
+    neutrinoArgs.push('--routing.assumechanvalid')
```
mrfelton: Can we move this to `resources/rpc.proto` please @JimmyMow since it doesn't need to be dynamic.

mrfelton:
@JimmyMow the '--routing.assumechanvalid' switch is only available after building lnd with the `experimental` build tag, like:

```
make install tags=experimental
```

I'm tempted to say that we should leave this out until it is included in an official lnd release. Otherwise, the only way that we can have it included in a release of our own is to bundle zap with one of our own custom lnd builds. I would rather see us include an official signed lnd binary in our official releases.

mrfelton:
Can you squash these commits plz @JimmyMow

PR #752, opened 2018-09-06
fix(lnd): do not require a single lnd process

mrfelton:
## Description:

There is no need to check wether an lnd process exists and prevent Zap from starting if so. We don't care if there is another process running. We only care that we can successfully start up our own process. If there is a conflict with another lnd process such as a port that we need already being in use we already handle this elsewhere in our code and we pass more detail to the error message the other way.

## Motivation and Context:

The lnd process lookup is slow, unreliable, and not needed. Remove redundant code that adds unneeded dependencies and improve user experience.

## How Has This Been Tested?

1. Start up an lnd process:
```
lnd --bitcoin.active --bitcoin.testnet --bitcoin.node=neutrino --neutrino.connect=mainnet1-btcd.zaphq.io
```
2. Then start up zap and verify that it gives you an error message, as before (actually we get more specific detail now)

**Before:*
screenshot 2018-09-07 01 14 36

**After:**
screenshot 2018-09-07 01 14 14

## Types of changes:

Enhancement

## Checklist:

### DIFF
--- a/app/lib/lnd/util.js
+++ b/app/lib/lnd/util.js
@@ -4,7 +4,6 @@ import axios from 'axios'
 import { promisify } from 'util'
 import { basename, dirname, join, normalize } from 'path'
 import { platform } from 'os'
-import { lookup } from 'ps-node'
 import { app } from 'electron'
 import isDev from 'electron-is-dev'
 import grpc from 'grpc'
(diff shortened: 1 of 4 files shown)


JimmyMow:
Tested ACK fa81760

PR #758, opened 2018-09-13
feat(lnd): update lnd to latest master build

mrfelton:
Update lnd to v0.5-beta-rc2-60-g1c4bd04c

https://github.com/LN-Zap/lnd/releases/tag/v0.5-beta-rc2-60-g1c4bd04c

This build enables the experimental features which are needed for #748

### DIFF
--- a/app/lib/lnd/neutrino.js
+++ b/app/lib/lnd/neutrino.js
@@ -15,7 +15,7 @@ const CHAIN_SYNC_COMPLETE = 'chain-sync-finished'
 
 // Events
 const ERROR = 'error'
-const CLOSE = 'close'
+const EXIT = 'exit'
 const WALLET_UNLOCKER_GRPC_ACTIVE = 'wallet-unlocker-grpc-active'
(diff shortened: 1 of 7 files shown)


JimmyMow:
@mrfelton is this working for you? Getting `Lnd has unexpectedly quit: null`

ZapUser77:
panic: runtime error: invalid memory address or nil pointer dereference
Is what LND 0.5 reports when I start ZAP.  I'm using an older version of ZAP, I'll update to current and see if I still get the same error.

mrfelton:
@JimmyMow I've updated this again this morning to use the latest lnd version (they released 0.5-beta this morning) and have tested this and confirmed it to be working.

I also made a couple of updates here to handle the termination process slightly differently which seems to be more reliable now (sending `SIGINT` to the process directly instead of using the grpc `stop` method).

I would suggest that you do `yarn clean && yarn` on this branch before retesting.

JimmyMow:
Tested ACK 3326db1

PR #759, opened 2018-09-13
feat(lnd): add lnd release scripts

mrfelton:
## Description:
Add the scripts that we have been using to create lnd releases and document how to use them.
## Motivation and Context:

I created these scripts a while ago and have been using them to publish lnd binaries. I'm adding them to the repo so that they don't get lost and so that they are available for general use!


JimmyMow:
Concept ACK

PR #768, opened 2018-09-16
fix(log): log lnd stderr output as errors

mrfelton:
## Description:

This PR includes three enhancements to our error logging:

1. Ensure that lnd output that goes to the stderr stream gets logged using the error log level and that the cause of the error is captured.

2. Limit the log namespaces that are enabled by default to only include the top level zap namespace and not also all of it's sub-namespaces.

3. Ensure that logs are output when the app is run in production mode. not only development mode so that its easier to diagnose an issue with a production build.

## Motivation and Context:

If lnd has a panic attack, the log messages are not using the proper error level and the cause of the error is not being captured and displayed to the user.

## How Has This Been Tested?

Manual testing with lnd instance that is crashing.

**Before:**

screenshot 2018-09-16 23 05 28

screenshot 2018-09-16 23 10 44

**After:**

screenshot 2018-09-16 23 04 50

screenshot 2018-09-16 23 03 16

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/lib/lnd/neutrino.js
+++ b/app/lib/lnd/neutrino.js
@@ -114,10 +114,9 @@ class Neutrino extends EventEmitter {
     // Listen for when neutrino prints data to stderr.
     this.process.stderr.pipe(split2()).on('data', line => {
       if (process.env.NODE_ENV === 'development') {
-        const level = lndLogGetLevel(line)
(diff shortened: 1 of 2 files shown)


JimmyMow:
Concept ACK be2482e

PR #796, opened 2018-09-20
fix(channels): fetch in activity mount not app

JimmyMow:
## Description:

Call `fetchChannels` from the `Activity` `componentWillMount` not in the `App.js` callback 

## Motivation and Context:

`App.js` `componentWillMount` callback gets called before we signal to the state machine that we are fully synced. `Activity.js` callback is where we call for all other app related user node data and only renders afterwards. In this PR we move the `fetchChannels` call to `Activity.js` so that it can be received by `ipcMain` and return us channel data.

## How Has This Been Tested?
Run the app on master and see the issue, run this PR and see it's fixed

## Checklist:

### DIFF
--- a/app/components/Activity/Activity.js
+++ b/app/components/Activity/Activity.js
@@ -28,12 +28,19 @@ class Activity extends Component {
   }
 
   componentWillMount() {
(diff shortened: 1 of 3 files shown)


JimmyMow:
Updated @mrfelton

PR #808, opened 2018-09-21
feat(lnd): use free port for lnd

mrfelton:
## Description:

Select free ports for lnd from a pool of options. By default we will use the standard lnd ports (10009 and 9735) but if those are not available then we will search for free ports in a nearby range.

## Motivation and Context:

This allows us to have multiple instances of the wallet running and avoids possible conflicts with other instances of lnd that the user may be running.

## How Has This Been Tested?

Manually

## Types of changes:

Enhancement

## Checklist:


JimmyMow:
Tested ACK 52cefd7

PR #814, opened 2018-09-21
fix(loading): fix glitch on loading sequence

mrfelton:
## Description:

Refactor such that the LoadingBolt component is only mounted once, in the root component.

## Motivation and Context:

The loading bolt was being shown in several distinct places, which was resulting in a glitch between completing onboarding and showing the wallet ui.

## How Has This Been Tested?

Manually - complete onboarding in various ways and see how the app transitions from the loading bolt to the wallet ui.

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/containers/Root.js
+++ b/app/containers/Root.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import { ConnectedRouter } from 'react-router-redux'
 import { Switch, Route } from 'react-router'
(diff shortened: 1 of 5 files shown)


app/containers/Root.js
```diff
@@ -208,47 +211,64 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
...
-          theme={onboardingProps.theme}
-          visible={!onboardingProps.onboarding.onboarding}
-        />
-        <Onboarding {...onboardingProps} />
-        <Syncing {...syncingProps} />
-      </div>
-    )
+class Root extends Component {
+  componentWillMount() {
+    const { fetchTicker } = this.props
```
JimmyMow: don't need to declare a const here if this is the only var we use. can just call `this.props.fetchTicker ()`

JimmyMow:
Tested ACK 05f30c4

@mrfelton one tiny comment

app/containers/Root.js
mrfelton: Our lint rules don't allow for that and not sure that want to change them just for this. Leaving as is for now.

Issue #818, opened 2018-09-25
Connecting Manually - Adding funds defaults dollar amt. to previous dollar amt. input


mattrolightning:
When adding trying to add funds to a channel you're opening manually, the dollar amount below the box where you input the number of Satoshis defaults to the previous amount input in the box while the box itself lists 000000 Satoshis. When you attempt to open a channel manually, the dollar amt. should default to 0 until you input the number of Satoshis in the box above.

![screen shot 2018-09-24 at 11 17 23 pm](https://user-images.githubusercontent.com/43560996/45990841-224a2000-c050-11e8-9b62-a7fbea065a75.png)

- Zap Desktop version: 2.2
- Operating System and version: Mac OS Sierra 10.12.6

JimmyMow:
Ok thanks @mattrolightning. PR #823 fixes this

mrfelton:
Fixed via #823

PR #823, opened 2018-09-26
fix(add-channel): dont remove input text

JimmyMow:
## Description:

Remove `clearTextInput` on `componentDidMount` in the `SubmitChannelForm` component. We aren't clearing it in the reducer so we shouldn't be in the UI

## Motivation and Context:
Fix for #818 

## How Has This Been Tested?
Manually

## Checklist:


mrfelton:
@JimmyMow shouldn't the form default to zero when opening it rather than using values from a previous view of the form?

JimmyMow:
@mrfelton We've always persisted the form input in the reducer and never clear it to 0. mostly because that's how I personally like to use it :p

I'm open to clearing the forms but I think that should be in a different PR that addresses it across the app? For now this will at least fix a temp bug

PR #824, opened 2018-09-29
fix(ticker): use Coinbase API - fixes #815

### LINKED ISSUE #815: Litecoin support broken in recent releases
Lnd w/ Litecoin no longer works.

## Description

After configuring wallet to point at custom lnd instance (w/ litecoind) the app used to finish onboarding and load up the main view (with transactions/pay/request/channel management). With recent releases this is no longer the case. After pressing the connect button, the app is stuck in a perpetual loop which says Loading.

## Expected Behavior

It should work.

## Actual Behavior

It doesn't work 🤣 

## Possible Fix

[Change ticker fetch logic to include Litecoin.](https://github.com/LN-Zap/zap-desktop/commit/f5afb7f085745a44eacd759982bffd3f6d78d346#diff-569b8f2c76eb653b49b4e215b9a2c151R15) As this logic now queries blockchain.info exclusively for rates (and blockchain.info does not support Litecoin), the app doesn't work properly. Reversing some of the changes in PR #639 will likely fix this issue. Yet to test this though.

## Steps to Reproduce

Commit `f9f1a53` is the last litecoin supporting build of zap. This issue is caused by commit `f5afb7f ` (aka PR #639).

## Context

There is a lot of interest in Litecoin support in Zap. The Litecoin developers have been quite busy recently, however we will be contributing code in the future. Not only to ensure Litecoin support, but also to work on improving zap for everyone.

## Your Environment

- Zap Desktop version: Zap v0.2.2-dev #f5afb7f
- Operating System and version: macOS 10.13.6

losh11:
Fixes bug where Litecoin users are stuck in infinite loading screen during on-boarding. Increases selection of fiat currencies selectable by user.

## Description:

PR #639 changed ticker fetching logic to use blockchain.info's ticker API instead of coinmarketcap. This broke Litecoin support, and caused the user to get stuck in an infinite loading screen during on-boarding. 

This PR uses the Coinbase price API to fetch bitcoin & litecoin rates and retains this data in the redux store.

## Motivation and Context:

[It makes litecoin actually work. See issue #815 for greater detail.](https://github.com/LN-Zap/zap-desktop/issues/815)

## How Has This Been Tested?

Tested on bitcoin network in neutrino mode (custom lnd node). Tested on litecoin network in litecoind and ltcd mode. Looks good in redux-devtools too. macOS 10.13.6 w/ node v8.12.0 & yarn latest.

## Types of changes:

Non-breaking bug-fix which fixes issue #815.
zap-desktop no longer uses the Coinbase API over Blockchain.info's ticker.

## Checklist:

- [x ] My code follows the code style of this project.
- [x ] I have reviewed and updated the documentation accordingly.
- [x ] I have read the _CONTRIBUTING_ document.
- [x ] I have added tests to cover my changes where needed.
- [x ] All new and existing tests passed.
- [x ] My commits have been squashed into a concise set of changes.


app/lib/utils/i18n.js
```diff
@@ -93,30 +93,172 @@ export const locales = [
   'zh-TW'
 ]
 
-// Defaine list of currencies that we will support.
+// Define list of currencies that we will support.
```
mrfelton: If we are going to define a complete list like this, could/should we pull the list from the `country-data-lookup` package, which is already installed and imported into this library?

See https://github.com/Zooz/country-data

You can fetch a list of all currencies buy calling `currencies.all`

app/lib/utils/i18n.js
```diff
@@ -93,30 +93,172 @@ export const locales = [
...
 export const currencies = [
-  'USD',
-  'EUR',
-  'JPY',
-  'GBP',
-  'CAD',
-  'KRW',
+  'AED',
```
mrfelton: If we are going to define a complete list like this, could we pull the list from the `country-data-lookup` package instead, which is already installed and being used in this library?

See https://github.com/Zooz/country-data

You can fetch a list of all currencies buy calling `currencies.all`

losh11: Sure can do. I'll take a look. This is just a list of all the currencies that Coinbase `https://api.coinbase.com/v2/exchange-rates?currency=btc` spits out.

mrfelton:
This work was merged via #921

PR #828, opened 2018-10-01
fix(lnd): show error on problem connecting to lnd

mrfelton:
## Description:

If there is a problem connecting to the Lightning interface on a local lnd instance, return the user to the onboarding screen and show the user the error message.

## Motivation and Context:

Previously, the app would hang on an infinite loading screen.

## How Has This Been Tested?

Simulate a problem connecting to the Lightning interface when using the local lnd by changing the connection deadline from 20 seconds to 0 seconds here https://github.com/LN-Zap/zap-desktop/blob/master/app/lib/lnd/lightning.js#L102 and trying to start up zap in the default mode.

You should be returned to the onboarding screen with the error message shown as a global error.

screen shot 2018-10-01 at 10 30 21

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/containers/Root.js
+++ b/app/containers/Root.js
@@ -4,6 +4,9 @@ import { ConnectedRouter } from 'react-router-redux'
 import { Switch, Route } from 'react-router'
 import PropTypes from 'prop-types'
 import { hot } from 'react-hot-loader'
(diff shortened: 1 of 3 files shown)


JimmyMow:
Tested ACK 0660801

PR #829, opened 2018-10-01
fix(channels): handle suggested nodes load failure

### LINKED ISSUE #392: Zap should handle failures to load the suggested nodes list gracefully
When the app starts up, it tries to connect to http://zap.jackmallers.com/suggested-peers. Currently, this URL is returning a 404. This is causing errors in the console:

```
GET http://zap.jackmallers.com/suggested-peers 404 (Not Found)

warning.js:33 Warning: Failed prop type: The prop `suggestedNodes` is marked as required in `SuggestedNodes`, but its value is `undefined`.
```

This seems to prevent the app from getting past the onboarding phase on startup with the user being stuck on the final 'loading' screen.

What is http://zap.jackmallers.com/suggested-peers supposed to provide?

Is it necessary?

mrfelton:
## Description:

If we are unable to load the suggest nodes list, display a message prompting the user to open channels.

## Motivation and Context:

Previously we would show an infinite spinner.

Fix #392

## How Has This Been Tested?

Manually - edit the suggested nodes URL here https://github.com/LN-Zap/zap-desktop/blob/master/app/lib/utils/api.js#L32 and start zap in order to simulate fetching a broken url.

**Normal:**
screenshot 2018-10-01 21 43 30

**With error - before:**
screenshot 2018-10-01 21 45 17

**With error - after:*
screenshot 2018-10-01 21 44 00

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/components/Contacts/SuggestedNodes/SuggestedNodes.js
+++ b/app/components/Contacts/SuggestedNodes/SuggestedNodes.js
@@ -31,24 +31,31 @@ const SuggestedNodes = ({
   return (
     <div className={styles.container}>
       <header>
-        <FormattedMessage {...messages.empty_description} />
+        {suggestedNodes.length > 0 ? (
+          <div>
+            <FormattedMessage {...messages.empty_description} />
+            <ul className={styles.suggestedNodes}>
+              {suggestedNodes.map(node => (
+                <li key={node.pubkey}>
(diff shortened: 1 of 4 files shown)


JimmyMow:
Tested ACK 5622952

PR #836, opened 2018-10-03
Change autopilot strings

michaelWuensch:
## Description:

Changes for the Autopilot selection strings. 

## Motivation and Context:

In some languages like german, the order of the words was wrong and could not be fixed within Crowdin.

## How Has This Been Tested?

it passed the lint test, zap runs and the correct strings are shown.

## Types of changes:

It changes two string references in en.json. 
This is my first commit and I am not sure, if replacing the references was the right way to go.
If that will cause problems for other languages, please let me know.
Also if further action is necessary to for Crowdin to recognize the changes, I am happy to update my PR.

## Checklist:

- [ x] My code follows the code style of this project.
- [x ] I have read the _CONTRIBUTING_ document.


michaelWuensch:
Thanks for your comment mrfelton!
Well I see a few possibilities. Which do you think is the best way to go?

1. revert the message key to the old one (e.g: "disable")
   In this case all existing translations already know the message key. Crowdin shows it as translated but 
   actually the word "autopilot" is missing.

2. change the message key for all translations and make the translation entry empty ("") so Crowdin shows it as untranslated

3. change the message key for all translations and add "Autopilot" at the end of the entry which should result in the same wording we had before and no translation gets lost.

4. revert the message key to the old one (e.g: "disable") and change the translation like in 2 or 3

mrfelton:
@michaelWuensch I would suggest we go with option 4.

I think it would be better for us to not break this string for the other languages as I don't believe the change of the message content is semantically different enough from the original content to cause issues within the UI. This is an optimisation for some languages.

michaelWuensch:
@mrfelton 
I updated the PR. 
Message keys are reverted and all translations were adapted. So no translation got lost.

mrfelton:
@michaelWuensch could you rebase this from master please?

Thanks.

michaelWuensch:
@mrfelton I just rebased the PR to master.

michaelWuensch:
@mrfelton 
Now its squashed and has a hopefully correct commit message...
A few more change requests and I'll become a real git pro! ;) 
Thanks for the patience!

mrfelton:
Thanks @michaelWuensch - I've gone ahead and merged that 👍

Issue #522, opened 2018-06-30
Users should be advised of the importance of keeping their seed safe and secure during the onboarding process


mrfelton:
Some feedback from Slack:

> When creating the seed it is explained that it is used for recovery. I think for the next five years or so we have to also show a big warning, that anyone who gets the seed will be able to steal your money. New users might not know this.
 -- @michaelWuensch

seanyesmunt:
Working on this

mrfelton:
Fixed via #1029

PR #908, opened 2018-11-19
style(ui): add warning about keeping seed private

seanyesmunt:
## Description:
Added a warning label in the `NewWalletSeed` component. This uses a new ui component called `WarningLabel`

## Motivation and Context:

For #522 

## How Has This Been Tested?

Manually tested on light and dark mode

### Before
screen shot 2018-11-19 at 1 38 30 am

### After
screen shot 2018-11-19 at 1 21 56 am
screen shot 2018-11-19 at 1 23 20 am

## Types of changes:

Style change

## Checklist:


app/components/Onboarding/NewWalletSeed/NewWalletSeed.scss
```diff
@@ -18,7 +18,6 @@
 
       &:nth-child(1) {
         width: 15%;
-        text-align: center;
```
seanyesmunt: I removed this so the left row of seed words is aligned with the left edge of the warning label box. I see the reason for centering it though. I think either looks ok.

seanyesmunt:
Will add a test for this component and run the messages script.

seanyesmunt:
@mrfelton I ran `yarn extract-messages` but it seems to have removed some messages. Any ideas why?

mrfelton:
@seanyesmunt in the `next` the onboarding process has been fully refactored - so this will need to be reworked against that branch.

Also, please could you add this new element into the storybook as part of this PR? IO notived that you added a new colour as part of this. I'm not sure wether or not we want to introduce a new colour to the colour palette, but if we do then that should also be added to the theme definitions and to colour palette section of the storybook (these are both newly added in the `next` branch)

I have asked Ole to take a look from a design perspective.

mrfelton:
Regarding the message strings - it looks like some of these were updated recently in master but the extract message command had not been run so some of the translation files were a little out of sync.

I have run `yarn extract-messages` against master and created a PR to resolve this issue here https://github.com/LN-Zap/zap-desktop/pull/916

mrfelton:
@seanyesmunt I rebased this branch and updated this PR slightly to reuse our existing `Message` component for the message. This was merged in #1029

PR #975, opened 2018-11-30
feat(grpc): use @grpc/grpc-js grpc implementation

### LINKED ISSUE #951: Migrate from the `grpc` package to `@grpc/grpc-js`.
## Detailed Description
The grpc package contains native code, and causes grief and difficulty of usage for the end user. Switching to the @grpc/grpc-js package, a pure javascript implementation, would simplify greatly the installation story for the end users. See #760 for instance, but also a few more issues that got created here in the past.

## Context
Since electron 3, it became possible to use the pure javascript version of grpc, since the necessary http2 module became available there. Also, a quick glance shows that zap-desktop doesn't use any feature that's currently not implemented in this version.

See https://github.com/grpc/grpc-node/blob/master/PACKAGE-COMPARISON.md for more details.

## Possible Implementation
Remove the dependency from `grpc` to `@grpc/grpc-js`, and change the import to `'@grpc/grpc-js'`. Nothing else should have to change.

mrfelton:
## Description:

Switch to the @grpc/grpc-js package, a pure javascript implementation, to simplify greatly the installation story for the end users.

## Motivation and Context:

The grpc package contains native code, and causes grief and difficulty of usage for the end user. Since electron 3, it became possible to use the pure javascript version of grpc, since the necessary http2 module became available there.

Fix #951

## How Has This Been Tested?

Manually

## Types of changes:

Dependency update

## Checklist:


JimmyMow:
Tested ACK be7ed5b

PR #984, opened 2018-11-30
fix(channels): refresh channel details on update

### LINKED ISSUE #724: Channel request limit does not update without a manual refresh
After making a payment to https://testnet.yalls.org/ with a direct channel open to to their node and visible in my channels list, the Pay limit & Request Balance indicators do not update. The only way to get them to update is to click the `refresh` link at the top of the channels list.

## Expected Behavior

Pay limit & Request balance should update in realtime after making the payment. Should not require a refresh.

## Actual Behavior

Values do not update without a manual refresh.

## Your Environment

- Zap Desktop version: master
- Operating System and version: OSX

mrfelton:
## Description:

Fetch an updated channel list if we receive a channel update to one of our active channels.

## Motivation and Context:

Previously, we would only update if the advertising_node or connecting_node was our own node. This fix expands the criteria so that we also trigger the channels list re-fetch if there is a match against one of our channels' remote_pubkey.

Fix #724

## How Has This Been Tested?

Manually - make a payment and verify that the channels balances get updated correctly once the payment has gone through successfully.

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/reducers/channels.js
+++ b/app/reducers/channels.js
@@ -276,32 +276,35 @@ export const pushclosechannelstatus = () => dispatch => {
 
 // IPC event for channel graph data
 export const channelGraphData = (event, data) => (dispatch, getState) => {
-  const { info } = getState()
+  const { info, channels } = getState()
   const {
     channelGraphData: { channel_updates }
   } = data
 
   // if there are any new channel updates
+  let hasUpdates = false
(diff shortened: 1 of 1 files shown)


JimmyMow:
Tested ACK 91a5532

PR #1004, opened 2018-12-03
feat(migration): import wallets from fs

mrfelton:
## Description:

When starting up the app, scan the local filesystem for wallet directories. If we find a directory without a corresponding entry in indexeddb, create a new database entry so that the wallet becomes available within the app.

## Motivation and Context:

- Ensure that users upgrading from an old version of Zap (pre indexeddb) have their wallets continue to be available.
- Make it possible to import wallets simply by moving them to an expected location on the filesystem and then starting the app

## How Has This Been Tested?

Manually -

1. Check out an old version of Zap (prior to the implementation of  indexeddb) and create a local wallet (check out the `0.2.2-beta` tag)
2. Quit Zap, Update the codebase to the `next` branch and restart the wallet.

You should find that your old wallet is now visible in the new home screen.

You can also test by deleting one of your db entries from indexeddb and restarting zap - the wallet entry should get recreated for you in indexeddb automatically and become available from the home screen.

## Types of changes:

Enhancement

## Checklist:

### DIFF
--- a/app/containers/Initializer.js
+++ b/app/containers/Initializer.js
@@ -48,8 +48,10 @@ class Initializer extends React.Component {
         } else {
           history.push(`/home/wallet/${activeWallet}`)
         }
-      } else {
-        history.push('/onboarding')
+      }
+      // If we have an active wallet set, but can't find it's settings then send the user to the homepage.
+      else {
+        history.push('/home')
       }
     }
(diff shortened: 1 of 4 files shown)


JimmyMow:
ACK f215f85

PR #1031, opened 2018-12-05
fix(wallet): display wallet name in settings menu

mrfelton:
## Description:

Display the wallet name in the settings menu using the `WalletName` component.

As part of this I also moved the `WalletName` component from `components/Home` into `components/Util` since it doesn't specifically relate to the Home component. Also moved the `Truncate` component since this is really as utility and not a UI component.

## Motivation and Context:

Be consistent with other places where we display the wallet name.

## How Has This Been Tested?

Manually - start remote and local nodes with and without a wallet name set and verify that the name displays correctly in the settings menu.

## Types of changes:

Enhancement

## Checklist:


lieteau2:
Tested ACK 39aea1f

PR #1091, opened 2018-12-14
feat(lndconnect): basic support for lndconnect links

mrfelton:
## Description:

Add basic support for `lndconnect` links.

`lndconnect` is an updated version of https://github.com/LN-Zap/zap-iOS/blob/master/docs/zap_connect_uri.md

## Motivation and Context:

Provide ability to populate custom connection fields (host, cert, macaroon) from a url.

## How Has This Been Tested?

1. Open the `Create new wallet` onboarding process
2. Select `Connect your own node`
3. Open your browser and access an `lndconnect` url

The form fields should be filled out for you automatically

example `lndconnect` link
```
lndconnect:?cert=~/.lnd/thedeathnode/tls.cert&macaroon=~/.lnd/thedeathnode/admin.macaroon&host=thedeathnode:10009
```

## Types of changes:

New feature

## Checklist:

### DIFF
--- a/app/main.dev.js
+++ b/app/main.dev.js
@@ -13,6 +13,8 @@ import installExtension, {
 } from 'electron-devtools-installer'
 import get from 'lodash.get'
 import path from 'path'
+import url from 'url'
+import querystring from 'querystring'
 import { mainLog } from './lib/utils/log'
 import ZapMenuBuilder from './lib/zap/menuBuilder'
 import ZapController from './lib/zap/controller'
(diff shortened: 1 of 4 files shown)


app/main.dev.js
```diff
@@ -163,16 +165,44 @@ app.on('ready', async () => {
...
+  app.setAsDefaultProtocolClient('lightning')
+  app.setAsDefaultProtocolClient('lndconnect')
+
+  app.on('open-url', (event, input) => {
+    mainLog.debug('open-url: %s', input)
+    event.preventDefault()
+
+    const parsedUrl = url.parse(input)
+
+    switch (parsedUrl.protocol) {
```
JimmyMow: should we add a `default` to this switch case and throw an error to the console? I envision this being helpful when other services build in support to connect to zap and they aren't inputting the correct uri handler, this way debugging issues will be easier.

mrfelton: We only handle `lightning` and `lndconnect` urls, so these are the only 2 things that can possibly trigger this part of the code.

When we call `sendMessage` that is already logged to the console, so it's quite easy to see both the input and output of these handlers (we log the input above).

JimmyMow:
@mrfelton also this PR assumes you are on the `Connection Details` form. I think if a user were to click this link from the default home screen the app should push the user to the `Connection Details` form and auto-populate. Simplifies the UX

app/main.dev.js
JimmyMow: ah yes of course you have to register the handlers to fire those callbacks anyways 👍

Issue #1108, opened 2018-12-17
Application displaying blank screen when restarting 


lieteau2:
After quitting desktop application and reopening it, the application screen is blank and I cannot proceed to use Zap 

I am testing version 0.3.0 prerelease 
<img width="1280" alt="screen shot 2018-12-17 at 1 31 53 pm" src="https://user-images.githubusercontent.com/45567325/50110566-342b0180-0200-11e9-8c66-0527949843ce.png">

JimmyMow:
@mrfelton I am getting this as well. For some more context, I found this in my logs: 

`zap:main [ERR]  Unable to start lnd: transition is invalid while previous transition is still in progress +5ms`

PR #1159, opened 2018-12-22
fix(app): reset onboarding after app refresh

mrfelton:
## Description:

Only trigger the onboarding process to start once the app is ready. This means waiting until we have loaded all the available wallet data and the onboarding process has been triggered from the main process.

## Motivation and Context:

This ensures that we do not attempt to start the onboarding process until the main process has been fully reset (disconnect / stop lnd) and prevents us from trying to start lnd when it is already started after a window refresh.

This is a more comprehensive fix for https://github.com/LN-Zap/zap-desktop/issues/1108 which doesn't have the sideeffect of breaking app reload.

## How Has This Been Tested?

Connect to a wallet, and then trigger a window reload (cmd+r). Previously this would result in a blank screen and an error about trying to start lnd in an invalid state on the terminal. With this patch it should work as expected.

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/containers/Initializer.js
+++ b/app/containers/Initializer.js
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import { withRouter } from 'react-router'
+import { appSelectors } from 'reducers/app'
 import { startActiveWallet } from 'reducers/lnd'
 import { initCurrency, initLocale } from 'reducers/locale'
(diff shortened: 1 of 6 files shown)


lieteau2:
Tested ACK 444ce11

mrfelton:
I'e pushed an update for this @JimmyMow which should get this working correctly regardless of where you are in the app when you refresh.

mrfelton:
rebased @JimmyMow

Issue #636, opened 2018-07-31
Font rendering is very bad on non-retina displays


janniks:
The font rendering looks horrible on external/non-retina displays.

e.g.
![screen shot 2018-07-31 at 11 38 13](https://user-images.githubusercontent.com/6362150/43451931-68139082-94b6-11e8-9dc6-3f4e11ed75ae.png)
![screen shot 2018-07-31 at 11 38 18](https://user-images.githubusercontent.com/6362150/43451937-69ffafe8-94b6-11e8-9bbd-837288716a41.png)
![screen shot 2018-07-31 at 11 38 21](https://user-images.githubusercontent.com/6362150/43451939-6bb06d96-94b6-11e8-83c0-96bc91a378ef.png)

bolatovumar:
I'm just going to add to this bug instead of opening a new issue.
![screen shot 2018-12-30 at 7 14 56 pm](https://user-images.githubusercontent.com/1934678/50554003-7d02a180-0c67-11e9-89b9-5e4e7647bbea.png)
![screen shot 2018-12-30 at 7 15 09 pm](https://user-images.githubusercontent.com/1934678/50554004-7d9b3800-0c67-11e9-844b-d513d4821a54.png)
Text in the above images is extremely small. Also, for some reason the parent link doesn't have a proper "hand" cursor once you click on it. You can still click on it to go back but the cursor is the "text selection" kind.

Also, this text in the sidebar is justified resulting in weird gaps between words. Other text that appears in this place is left-aligned (when you click "Open a channel" for example).
![screen shot 2018-12-30 at 7 15 30 pm](https://user-images.githubusercontent.com/1934678/50554019-c9e67800-0c67-11e9-81e4-c5b354aa7815.png)

PR #1202, opened 2018-12-31
fix(ui): left-align column text

bolatovumar:
Addresses the weird gap between words in column issue from #636. See the last image and paragraph here: [https://github.com/LN-Zap/zap-desktop/issues/636#issuecomment-450604350](https://github.com/LN-Zap/zap-desktop/issues/636#issuecomment-450604350)

### DIFF
--- a/app/components/Contacts/SuggestedNodes/SuggestedNodes.js
+++ b/app/components/Contacts/SuggestedNodes/SuggestedNodes.js
@@ -43,7 +43,7 @@ const SuggestedNodes = ({
    */
   const renderNodes = () => (
     <>
-      <Text mb={4} textAlign="justify">
+      <Text mb={4}>
         <FormattedMessage {...messages.empty_description} />
       </Text>
       {suggestedNodes.map(node => (
--- a/app/components/Contacts/SuggestedNodes/messages.js
+++ b/app/components/Contacts/SuggestedNodes/messages.js
@@ -4,7 +4,7 @@ import { defineMessages } from 'react-intl'
 export default defineMessages({
   connect: 'Connect',
   empty_description:
-    "Hmmm, looks like you don't have any channels yet. Here are some suggested nodes to open a channel with to get started",
+    "Hmmm, looks like you don't have any channels yet. Here are some suggested nodes to open a channel with to get started.",
   empty_description_alt:
     "Hmmm, looks like you don't have any channels yet. Open some channels to get started."
 })
--- a/app/components/Settings/Fiat/Fiat.scss
+++ b/app/components/Settings/Fiat/Fiat.scss
@@ -1,6 +1,7 @@
 @import 'styles/variables.scss';
 
 .submenuHeader {
+  cursor: pointer;
   padding: 20px;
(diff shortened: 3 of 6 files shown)


bolatovumar:
I tried out using 16px font size (currently set at 10px) on my MacBook Air (non-retina) and it looks much more readable. My only concern for increasing the font size here would be that it might show up too big on retina screens. I doubt it would be an issue but would like someone to confirm.

bolatovumar:
And as far as legibility for non-retina screens goes I found that using `font-weight: 400;` instead of the currently used `font-weight: 300;` works wonderfully. Not sure whether retina screens would be negatively impact by this change though.

mrfelton:
@bolatovumar regarding the font-sizes - I think the best way to do this is to take a more holistic approach and make our font-sizes (and other page elements) responsive via breakpoints.

I don't think we really want to adjust the default font-sizes or weights, but we can tailor for smaller/larger screen sizes.

Rebass (our theme system, https://github.com/rebassjs/rebass) already supports responsive theming and I actually already have a PR in my local branch that makes a start at making our font-sizes responsive using this approach. If you are interested on taking this on let me know and I can share what I have with you.

bolatovumar:
@mrfelton do you mean you would prefer to adjust font sizes/weights for NON-RETINA displays through media queries that target these displays rather than enable the change for all displays?

The issue with font sizes and weights here is not due to the viewport or resolution being too small but due to the font-sizes themselves being too small and therefore rendering less than ideally.

For example, I'm having trouble reading some of the text without glasses (especially within the menu where the fiat/theme/language are set). It's a lot better with the glasses but still quite small. I generally find that using <12px font sizes in UIs is a no-go from the accessibility perspective especially if light font weights are used.

The type scale for the Material Design system from Google shows how they use Roboto at various sizes [(https://material.io/design/typography/the-type-system.html#)](https://material.io/design/typography/the-type-system.html#). Notice that the 10px size is used in all-caps only and in the Regular font-weight.

mrfelton:
@bolatovumar Yes, I was suggesting that we adjust font sizes for specific breakpoints but leave the defaults / base sizes as is. Reason being that we've been through a design process and these were the sizes and weights that were picked by the designers, so changing this would have an impact on the designs and overall look and feel. Not saying that the current font sizes are right or perfect, just that our current look and feel has been designed with these sizes.

The reason that I was looking at adjusting the font sizes via breakpoints is that we know that when the app is displayed in full screen mode on a large hi-res retina display, the fonts look very small and there is a lot of space on the screen - so making fonts larger when there is is more screen space would be nice.  But, making the font sizes larger on smaller screens / resolutions could cause some issues in some parts of the app, with things not fitting onto the page fully without requiring the user to scroll - which is something that we want to avoid if possible.

I generally work on a mac with a retina display, at a res of 1440 x x900 and there the font size is good (for me). But I think your point about accessibility is a valid one and we certainly don't want users having to squint at the screen to read things.

@bolatovumar could you open a separate issue to carry on this discussion about font sizes please, so that it doesn't get lost in this closed PR? Thanks.

PR #1203, opened 2018-12-31
fix(ui): set global error z-index to 99999

### LINKED ISSUE #1200: Error message not shown when requesting more than maximum amount of BTC
## Description

When you enter more than a maximum amount allowed in the "Request Bitcoin" screen you get no error message and it's quite confusing. If you close this modal, however, you see an error message behind it.

screen shot 2018-12-30 at 7 24 22 pm

## Expected Behavior

I would expect to see the error message saying you can't request more than a certain maximum amount of BTC. It would probably also be nice to state the maximum amount one can request upfront and maybe even limit the user's ability to enter more than that amount in the textbox

## Actual Behavior

No error message is displayed.

## Steps to Reproduce

1. Click on "Request" button
2. Enter more than a maximum amount allowed in the textbox (let's say 1 BTC)
3. Click on the button that says "Request 1 BTC"
4. Observe that the button jitters weirdly and nothing happens.
5. Quickly close the modal
6. See the error message that was behind the modal and should have been above it.

## Your Environment

- Zap Desktop version: 0.3.2-beta.5107
- Operating System and version: MacOS X Mojave

bolatovumar:
Fixes #1200.

I set the `z-index` property on for the `GlobalError` component to `99999` from `100`. The bug was caused by the modal having a higher z-index value than the error (`999` vs `100`).

This bug is fixed for now, however, a more robust solution needs to be implemented to avoid these types of bugs in the future. See this post for some ideas: [https://css-tricks.com/handling-z-index/](https://css-tricks.com/handling-z-index/).

Also, there might be a bug somewhere in the `componentDidUpdate` because I noticed that the timeout of `10000` doesn't always apply. If you trigger an error, wait a bit, close it and then quickly trigger another error you will see that the new error disappears in much less than 10 seconds.

### DIFF
--- a/app/components/GlobalError/GlobalError.js
+++ b/app/components/GlobalError/GlobalError.js
@@ -32,7 +32,7 @@ class GlobalError extends React.Component {
         {show =>
           show &&
           (springStyles => (
-            <Box mt="22px" px={3} width={1} css={{ position: 'absolute', 'z-index': '100' }}>
+            <Box mt="22px" px={3} width={1} css={{ position: 'absolute', 'z-index': '99999' }}>
               <animated.div style={springStyles}>
                 <Notification variant="error" onClick={clearError}>
                   {errorToUserFriendly(error)}


mrfelton:
@bolatovumar we did have a more robust way of handling this with some global constants (see https://github.com/LN-Zap/zap-desktop/blob/master/app/styles/variables.scss#L32-L64) but are currently in the process of moving away from scss towards styled-components and have not yet migrated that.

I think the best way to do this in our new setup would be to define z-index groups in our base theme (see https://github.com/LN-Zap/zap-desktop/blob/master/app/themes/base.js) and then refer to those props in the places where we are applying z-index.

I'll merge this PR as is since it fixes the issue at hand, but a nice follow up to this could be to refactor as above.

mrfelton:
@bolatovumar could you create separate bug reports for the `componentDidUpdate` issue please?

bolatovumar:
> @bolatovumar could you create separate bug reports for the `componentDidUpdate` issue please?

Raised as #1206

PR #1205, opened 2018-12-31
 fix(ui): wording update

### LINKED ISSUE #1204: Weird wording in the blockchain sync screen
screen shot 2018-12-30 at 11 18 23 pm
"... while whilst ..." feels very weird here. They are two variations of the same word repeated after each other. "Please wait while we fetch..." would probably be better here.

bolatovumar:
Fixes #1204


mrfelton:
Needs a rebase @bolatovumar

bolatovumar:
Conflict resolved.

mrfelton:
@bolatovumar generally we prefer people to rebase their feature branches on top of master rather than to merge master into the branch. This keeps the commit history as clean as possible when we finally merge into master.

Please could you squash these 2 commits, and also add the reference to issue #1204 which it fixes so that we have the cross reference in our changelog?

bolatovumar:
Will do!

PR #1208, opened 2019-01-01
fix(unit tests): clean up some warning messages

### LINKED ISSUE #1201: Clean up warnings when running unit tests
When running `npm run test-unit` multiple tests finish with warnings. This sort of looks bad as it gives the impression of something being wrong even though tests don't actually fail. This should be cleaned up where possible.

Zap release version: 0.3.2-beta (0.3.2-beta.5107)

bolatovumar:
Fix #1201.

This doesn't clean up all warnings but cleans up some of them. 

I couldn't figure out how to fix one of them (incorrect ref usage `Select.js`). 

The other one that I left as is would involve changing the `size` prop passed to a `Code` component in `QRCode.js` to 112 (a number) instead of "80%" (string) since the prop type for this component require `size` to be a number. I wasn't sure if this was an acceptable change so I didn't do it.

### DIFF
--- a/test/unit/components/UI/Countdown.spec.js
+++ b/test/unit/components/UI/Countdown.spec.js
@@ -5,7 +5,7 @@ import { Countdown } from 'components/UI'
 
 describe('component.UI.Countdown', () => {
   it('should render correctly with default props', () => {
-    const wrapper = shallow(<Countdown />)
+    const wrapper = shallow(<Countdown date={new Date('2009-01-03T18:15:05+00:00')} />)
     expect(toJSON(wrapper)).toMatchSnapshot()
   })
 })
--- a/test/unit/components/UI/LightningInvoiceInput.spec.js
+++ b/test/unit/components/UI/LightningInvoiceInput.spec.js
@@ -10,7 +10,7 @@ describe('component.UI.LightningInvoiceInput', () => {
   it('should render correctly', () => {
     const tree = renderer
       .create(
-        <IntlProvider>
+        <IntlProvider locale="en">
           <ThemeProvider theme={dark}>
             <Form>
               <LightningInvoiceInput field="name" chain="bitcoin" network="mainnet" theme={dark} />
--- a/test/unit/components/UI/__snapshots__/Countdown.spec.js.snap
+++ b/test/unit/components/UI/__snapshots__/Countdown.spec.js.snap
@@ -13,7 +13,7 @@ exports[`component.UI.Countdown should render correctly with default props 1`] =
   <FormattedRelative
     style="best fit"
     updateInterval={1000}
-    value={Date { NaN }}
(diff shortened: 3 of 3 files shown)


mrfelton:
@bolatovumar  Regarding the issue with the QRCode `size` prop - I think what we really need to do here is submit a patch to https://github.com/zpao/qrcode.react that allows strings as well as numbers for this prop,. We don't really want to set an absolute size, since the QRCode container element can be displayed at different sizes and the inner qrcode needs to display displayed at a size relative to it's container.

bolatovumar:
I'm not sure if SVG actually allows for strings. It makes sense in the
context of the DOM but may not make sense in other contexts. I will try to
look into this more to see if it would make sense to raise an issue in a
different repo or just leave it as is.

test/unit/components/UI/Countdown.spec.js
```diff
@@ -5,7 +5,7 @@ import { Countdown } from 'components/UI'
 
 describe('component.UI.Countdown', () => {
   it('should render correctly with default props', () => {
-    const wrapper = shallow(<Countdown />)
+    const wrapper = shallow(<Countdown date={new Date(0)} />)
```
mrfelton: We should use a date other than the epoch for this. It would make it easier to identify date related issues with the snapshot in the future. I'd suggest simply `new Date()` rather than `new Date(0)`

bolatovumar:
Using new Date() would result in a new date getting generated every time
the test is run resulting in a broken test because the snapshot wouldn't
match anymore.

mrfelton:
> Using new Date() would result in a new date getting generated every time

Yes, that's a fair point - so lets set it to a specific timestamp that is not the Epoch. The problem with using the Epoch is that it could mask issues with invalid dates, which tend to render as the Epoch.

bolatovumar:
Sure, I can change it to some specific date instead.

bolatovumar:
Changed the date string to "January 3, 2009 18:15:05"

bolatovumar:
Ooops, the build failed. I should be more specific in my date string and specify a timezone.

mrfelton:
@bolatovumar 

Sorry to be the commit message Nazi on this...  but can you make the following 2 changes to the commit message please:

1) commit scope should be `test` rather than `fix` so that this change doesn't show up in our release notes. (only `bugs` and `fix` commits are included in the release notes, since those are the only things relevant to users of the app). We don't really want these types of internal changes being added to the release notes as it makes them far too granular and hard to parse. See https://github.com/LN-Zap/zap-desktop/blob/master/CONTRIBUTING.md#committing-and-pushing-changes for a list of the various scopes that can be used.

2. Commit message should include `Fix #1201` so that we get the cross reference.

mrfelton:
Closing - will merge in https://github.com/LN-Zap/zap-desktop/pull/1208

PR #1211, opened 2019-01-01
fix(ui): add autofocus to search input field

### LINKED ISSUE #1210: Clicking on the search icon doesn't focus the search input field
## Expected Behavior

I would expect the search input field to be focused as soon as I click on the search icon. As it is now, I have to click on the search input field itself after clicking the search icon.

## Actual Behavior

Search input field is not in focus.

## Possible Fix

Focus the search input field in JS after the search input icon has been clicked.

## Steps to Reproduce

1. Click on the magnifying glass icon by by the "Refresh" text above the transactions list.
2. Notice that the "Search" text appears on the right but the field is not focused. You need to click on it for it to gain focus.

bolatovumar:
Fixes #1210

### DIFF
--- a/app/components/Activity/Activity.js
+++ b/app/components/Activity/Activity.js
@@ -113,6 +113,7 @@ class Activity extends Component {
       <>
         <Form width={1}>
           <Input
+            autoFocus
             field="search"
             id="search"
             type="text"


bolatovumar:
Travis CI failed for mysterious reasons again. It seem to fail only on the latest version of node and only on Linux. MacOS seems to be fine for both versions. This may be worth investigating as it seems to happen very frequently.

PR #1230, opened 2019-01-03
fix(ui): improve english grammar on sync page

### LINKED ISSUE #1204: Weird wording in the blockchain sync screen

bolatovumar:
Fix #1204


bolatovumar:
Closed #1205 and opened this PR instead

Issue #1213, opened 2019-01-01
Consider using TypeScript


bolatovumar:
[TypeScript](https://www.typescriptlang.org) advertises itself as "Javascript that scales".

I have been using it at work for the past three years and couldn't recommend it enough. Quite a few large companies have adopted it for their applications (Slack, VSCode, HootSuite, a bunch of others...). I find that it helps tremendously with catching simple bugs so they don't even make it into the build in the first place. It also makes refactoring a breeze. IDE support for it is excellent. Linting ecosystem is great. I can go on and go on about how great it is really.

I see the project already uses Flow to some degree but it's becoming very clear that Flow has been losing to TS in a big way for the past several years. Very few people outside of Facebook seem to be using it and as a result the ecosystem is not as rich.

I understand that switching to TS in one go is a big undertaking but it can be done gradually. It supports even plan .js files so there is no need to jump into .ts-land right away. It also has quite a few compiler options that allow you to configure it to your liking. You can start with loose rules at first and make them stricter as you go.

You may still not want to switch to it at this stage of the project but it's something worth keeping in mind.

mrfelton:
@bolatovumar  As you noted, this project makes very limited use of Flow currently. Where it has been used, it has brought a lot of benefit and I'd be keen to see this type of thing integrated much more deeply across the project as a whole.

As for Flow vs TS - I don't have a huge amount of experience with either, but my thinking about which is a better fit moving forwards pretty much mirrors your own for many of the reasons that you have mentioned.

A nice initial patch to get the ball rolling on this might be to replace everything todo with Flow, for it's Typescript equivalent, in all of the the places where it has been used across the codebase so that we can fully replace Flow with TypeScript from the onset (Would like to avoid having both Flow and TS active at once). So the initial scope would be pretty much limited to updating the tooling itself, and the code in `lib/lnd/config` and the few places where that lib is used.

From there, we could start to incrementally built TS support into other parts of the app as and when it makes sense.

mrfelton:
My preference would be to use babel for transpiring as we currently depend on this quite heavily and there are some nuances with the setup. Keeping the changeset minimal would be ideal.

bolatovumar:
@mrfelton I started working on this in my local branch and I'm using Babel's built-in TS preset. It seems to be the easiest way to integrate TS considering the heavy usage of Babel in this project.

It's quite a hefty undertaking but a good way to learn the codebase and discover a bunch of smaller issues.

[...]

PR #1258, opened 2019-01-06
chore(devops): add bare-bones typescript configuration

bolatovumar:
Start work on #1213


bolatovumar:
@mrfelton so, I attempted to replace all of flow stuff with TS entirely. Removing the all the dependencies and plug-ins is pretty easy. Where I ran into trouble was converting the files that currently use flow to TS. There are two options here:

1. Keep using the .js extension for those files and replace all the flow type information with TS inline type information as described [here](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html). This works but the typing info takes up more space than in the case of flow. The main issue here is that you can't define types like `someProp: string` and have to put the type information above `someProp` instead so it takes up two lines. Example:
```javascript
/** @type {string} */
host
```
instead of
```typescript
host: string
```
2. Change the file extensions of the files using flow to `.ts` and update the build pipeline to recognize `.ts` files. This is fairly easy to do with the new babel typescript [preset](https://babeljs.io/docs/en/babel-preset-typescript). I'm able to use it to build the app without any issues as far as I can tell, however, when I try to use the `npm run dev` command it chokes on the first `.ts` file it encounters. Specifically, the command under `start-main-dev` in package.json is what throws the error. Looks like that command doesn't specify a webpack configuration file so the typescript files aren't recognized. I created a new webpack configuration file specifically for that task to use and added the stuff required for typescript to it and that solves the issue of it not recognizing typescript files but when the command finishes running all I get is an empty Electron window. Sigh...
I can keep fiddling with this to try and resolve whatever issue there is but my knowledge of webpack/babel/electron build configuration stuff is limited so I'm not sure if I will succeed.

I guess there is also a third option here: remove whatever flow type declarations there are completely and build up from scratch from there (similar to option 1 really). This might be an OK thing to do since there are not many files which currently have inline flow type declarations.

bolatovumar:
@mrfelton got it.

mrfelton:
We have decided against implementing typescript support for the time being, and have also removed Flow support from our codebase. Unless we are going to use it exclusively/extensively (which we are not) It's too much overhead for us with little gain.

Thank you for your work on this. I'm going to close the PR for now. We may revisit this in the future.

PR #1274, opened 2019-01-07
fix(wallet): handle invoices with msat amounts

### LINKED ISSUE #1273: Unable to pay invoices that are denoted in amount other than satoshis
## Description

Invoices generated by blockstream store are denoted in millisatoshis and we don't handle them properly., We are expecting the amount to be in satoshis and when we don't find the amount we are looking for, we assume it is a zero amount invoice and show the user the amount fields to allow them to enter a custom amount.

Ref https://twitter.com/erickuhn19/status/1081077191880826880

## Expected Behavior

Bolt 11 allows invoices to be denoted in any unit. We should support all, not just satoshis.

## Actual Behavior

Only satoshi denoted invoices are correctly handled

## Steps to Reproduce

1. Generate an invoice on blockstream store
2. Try to pay it
3. Note that the amount is not parsed from the invoice

## Your Environment

- Zap Desktop version: 0.3.2
- Operating System and version: Any

mrfelton:
## Description:

Ensure that invoices with amounts that are specified in millisatoshis are properly handled.

## Motivation and Context:

Fix #1273 

## How Has This Been Tested?

- I've added some new knobs to the knobs tab in the storybook that allow you to easily test the request form with various different types of invoices.

- Ensure currency conversions continue to function properly in the app.

- Ensure `lightning:` links work correctly

- Try paying various different types of invoices

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/lib/utils/btc.js
+++ b/app/lib/utils/btc.js
@@ -38,7 +38,7 @@ export function bitsToBtc(bits, price) {
   if (bits === undefined || bits === null || bits === '') return null
   const sats = bits * 100
 
-  return satoshisToBtc(sats, price) || 0
+  return satoshisToBtc(sats) || 0
 }
 
 export function bitsToSatoshis(bits, price) {
@@ -136,13 +136,13 @@ export function fiatToBits(fiat, price) {
 }
 
 export function fiatToSatoshis(fiat, price) {
-  return btcToSatoshis(fiatToBtc(fiat, price))
-  return btcToFiat(satoshisToBtc(satoshis), price)
+  const btcAmount = fiatToBtc(fiat, price)
+  return btcToSatoshis(btcAmount)
 }
 
 export function fiatToMilliSatoshis(fiat, price) {
-  return btcToMillisatoshis(fiatToBtc(fiat, price))
-  return btcToFiat(millisatoshisToBtc(satoshis), price)
+  const btcAmount = fiatToBtc(fiat, price)
+  return btcToMillisatoshis(btcAmount)
 }
 
 export function convert(from, to, amount, price) {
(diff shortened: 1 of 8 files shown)


app/lib/utils/btc.js
```diff
@@ -93,6 +140,11 @@ export function fiatToSatoshis(fiat, price) {
   return btcToFiat(satoshisToBtc(satoshis), price)
 }
 
+export function fiatToMilliSatoshis(fiat, price) {
+  return btcToMillisatoshis(fiatToBtc(fiat, price))
```
korhaliv: Looks like a mistake. Second return is unreachable

app/lib/utils/btc.js
```diff
@@ -31,20 +38,26 @@ export function bitsToBtc(bits, price) {
...
+}
+
+export function bitsToMillisatoshis(bits, price) {
+  if (bits === undefined || bits === null || bits === '') return null
+
+  return bits * 100 * 1000 || 0
 }
 
 export function bitsToFiat(bits, price) {
   if (bits === undefined || bits === null || bits === '') return null
```
korhaliv: This condition is used a lot in this module (>3 times). I would extract it into a separate function. Like isEmptyAmount. For readability and ease of refactoring later.

app/lib/utils/btc.js
mrfelton: Thanks. Have refactored.

app/lib/utils/btc.js
mrfelton: This was bothering me too but thought I'd leave it out of this PR. In retrospect, give that this PR makes quite a few changes across this file it makes sense to clean that up here and now. Have added an additional refactor commit that extracts this.

korhaliv: ok

PR #1275, opened 2019-01-07
fix(wallet): support uppercase payment requests

### LINKED ISSUE #1266: Upper case bech32 not recognized
According to the bech32 spec, there is no distinction between upper/lower case. some QR codes use a special mode where they only use UPPER case to encode the data more efficiently.

mrfelton:
## Description:

Add support for uppercase payment requests.

I've patched `bolt11` to support this. See https://github.com/bitcoinjs/bolt11/pull/12

## Motivation and Context:

Fix #1266

## How Has This Been Tested?

- Convert a payment request to all uppercase and verify it can be used as expected.
- Verify normal payment requests continue to function as normal

## Types of changes:

Bug fix

## Checklist:


JimmyMow:
Tested ACK 49e7b6e

PR #1293, opened 2019-01-08
fix(payment): timeout payment requests after 60 seconds

### LINKED ISSUE #1292: Payment that timed out remains in sending state until app is restarted
image

Expecting the payment to fail after 1 minutes.

Actual: Keep spinning.

mrfelton:
## Description:

Mark payments as failed if there is no response within 60 seconds.

## Motivation and Context:

Fix #1292

## How Has This Been Tested?

Try to pay a payment that times out.

I was able to do this by trying to send a payment immediately after logging into a local wallet - presumably before lnd had a chance to reestablish an up to date network picture.

image

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/components/Activity/Payment/Payment.js
+++ b/app/components/Activity/Payment/Payment.js
@@ -50,7 +50,9 @@ const Payment = ({
             )}
             {payment.status === 'failed' && (
               <Message variant="error">
-                <FormattedMessage {...messages.status_error} /> {payment.error}
+                <FormattedMessage {...messages.status_error} />
+                {` `}
+                {payment.error}
               </Message>
             )}
           </>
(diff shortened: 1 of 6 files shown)


lieteau2:
Tested ACK

mrfelton:
@JimmyMow this should be working again now.

Initially I opted to switch from `sendPaymentSync` to `sendPayment`, which according to Roasbeef is better maintained and more robust. However, on further investigation I believe that a switch to that would be more involved and not something that I want to get into for this PR.

So I've switched back to using `sendPaymentSync` and added the timeout to that, as well as a couple of more checks for specific errors that we could encounter.

JimmyMow:
Makes sense to me 👍 

Tested ACK 151cda8

PR #1298, opened 2019-01-09
chore(deps): remove unused dependency chalk

mrfelton:
## Description:

Remove unused `chalk` dependency

## Motivation and Context:

Package isn't used, so remove it.


lieteau2:
Tested ACK

PR #1299, opened 2019-01-09
perf(boot): load files asynchronously

mrfelton:
## Description:

Load files asynchronously to ensure that we don't block the event loop needlessly.

## Motivation and Context:

grpc proto file and tls.cert was being loaded using Sync versions of the fs methods which causes a bit of hanging in the boot process. Replace with non-blocking async versions.

## How Has This Been Tested?

1. Create new wallet
2. Load existing wallet

## Types of changes:

Performance enhancement.

## Checklist:

### DIFF
--- a/app/lib/lnd/util.js
+++ b/app/lib/lnd/util.js
@@ -9,7 +9,8 @@ import { credentials, Metadata } from '@grpc/grpc-js'
 import get from 'lodash.get'
 import { mainLog } from '../utils/log'
 
-const fsReadFile = promisify(fs.readFile)
+const readFile = promisify(fs.readFile)
+const stat = promisify(fs.stat)
 
 // ------------------------------------
 // Constants
@@ -110,7 +111,7 @@ export const getDeadline = timeoutSecs => {
 export const createSslCreds = async certPath => {
   let lndCert
   if (certPath) {
-    lndCert = await fsReadFile(certPath).catch(e => {
+    lndCert = await readFile(certPath).catch(e => {
       const error = new Error(`SSL cert path could not be accessed: ${e.message}`)
       error.code = 'LND_GRPC_CERT_ERROR'
       throw error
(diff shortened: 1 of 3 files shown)


app/lib/lnd/util.js
```diff
@@ -163,15 +164,17 @@ export const waitForFile = (filepath, timeout = 1000) => {
 
   // Promise B resolves when the file has been found.
   let promiseB = new Promise(resolve => {
-    let intervalId = setInterval(() => {
+    let intervalId = setInterval(async () => {
```
korhaliv: Not obvious why searching for file is wrapped into setInterval. Maybe add some comments?

app/lib/lnd/util.js
```diff
@@ -163,15 +164,17 @@ export const waitForFile = (filepath, timeout = 1000) => {
...
       mainLog.debug('waiting for file: %s', filepath)
-      if (!fs.existsSync(filepath)) {
+      try {
+        await stat(filepath)
+        mainLog.debug('found file: %s', filepath)
+        clearInterval(intervalId)
+        clearTimeout(timeoutId)
+        resolve()
+      } catch {
```
korhaliv: is this error swallowing here intentional?

app/lib/lnd/util.js
```diff
@@ -163,15 +164,17 @@ export const waitForFile = (filepath, timeout = 1000) => {
...
   // Promise B resolves when the file has been found.
   let promiseB = new Promise(resolve => {
-    let intervalId = setInterval(() => {
+    let intervalId = setInterval(async () => {
       mainLog.debug('waiting for file: %s', filepath)
-      if (!fs.existsSync(filepath)) {
+      try {
+        await stat(filepath)
+        mainLog.debug('found file: %s', filepath)
+        clearInterval(intervalId)
```
korhaliv: Probably better to clear interval and timeout before `await stat(filepath)` incase it throws. Or even better add finally block

app/lib/lnd/util.js
mrfelton: Have added some additional comments to this method

app/lib/lnd/util.js
mrfelton: We want to clear the timers in one of 2 different places. Either here after the file has been successfully found. Or, in the other promise when the timeout has been reached. We don't want to remove the timers in the case of a throw, since in that instance we want to keep rechecking for the file, scheduled by the interval timer.

app/lib/lnd/util.js
mrfelton: Yes - have added additional comment

PR #1312, opened 2019-01-11
refactor(ui): introduce dialog component

### LINKED ISSUE #1191: Implement new Delete Wallet confirmation dialog
## Detailed Description

Implement custom design for delete wallet dialog

## Context

Currently using the system dialog to handle this and the design doesn't really fit with the app.

## Possible Implementation

Implement new Wallet Delete confirmation dialog based on Zeplin design: 
zpl://screen?sid=5bf3f08593b7393e8cf59e13&pid=5bacdbad33af886d419f1b34

image

korhaliv:
## Description:
Introduce a reusable dialog component

## Motivation and Context:
Add ability to show themed dialog instead of system one

## How Has This Been Tested?
Manually via storybook entry

## Screenshots: 
image

## Types of changes:
UI enhancement

## Checklist:


app/components/UI/Dialog.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { Box, Flex } from 'rebass'
+import { Heading, Button } from 'components/UI'
+import X from 'components/Icon/X'
+
+const Container = styled(Box)`
```
mrfelton: You could probably switch from using `Box` to `Card`, which supports the border props. See https://rebassjs.org/Card. Then you could do away with this `Container` component.

app/components/UI/Dialog.js
```diff
@@ -0,0 +1,68 @@
...
+const CloseButton = ({ onClick }) => (
+  <Flex justifyContent="space-between">
+    <Box
+      css={{ height: '40px', cursor: 'pointer', opacity: 0.6, '&:hover': { opacity: 1 } }}
+      ml="auto"
+      onClick={onClick}
+      p={2}
+      px={10}
+    >
+      <X width={15} height={15} />
```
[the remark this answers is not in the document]
korhaliv: Yes I agree. I've copied this from some other modal that we have. Maybe it's even better to extract it to keep it consistent across the board

app/components/UI/Dialog.js
korhaliv: Agreed

PR #1313, opened 2019-01-11
feat(wallet): generate and display lndconnect qrcodes

### LINKED ISSUE #1145: Generate LNDConnect QR code for remote nodes that want to use Zap iOS
Right now the process for getting Zap iOS up and running is a bit difficult. An improvement on the current user story would be:

1. User runs Node Launcher
2. User LNDConnects to Zap Desktop from Node Launcher
3. User then scans LNDConnect QR code from Zap Desktop to connect to Zap iOS

mrfelton:
## Description:

Generate and display an lndconnect QR code. Currently this only applies to remote nodes, since we don't know which IP address to use when working with local nodes.

I created https://github.com/LN-Zap/node-lndconnect to house some of the core utility methods for working with lndconnect uris.

## Motivation and Context:

Provide an easy way for users to connect the mobile app to their remote nodes

Fix #1145

## How Has This Been Tested?

Manually

## Types of changes:

New feature

## Screenshot

image

## Checklist:


app/lib/lnd/config.js
```diff
@@ -293,6 +310,31 @@ class LndConfig {
       debug('Connection type is local. Assigning settings as: %o', defaultLocalOptions)
       Object.assign(this, defaultLocalOptions)
     }
+
+    // Set the ready prop once we have calculated all of the config properties.
+    const ready = new Promise(async resolve => {
```
korhaliv: Consider avoiding using `new Promise` here. I believe this can be done in a simpler and more readable way like this: (haven't tested this code)

```
const isReady = async () => {
    try {
        if (this.type === 'local') {
            // We don't support lndconnect for local nodes.
            return Promise.resolve(true)
        }

        const { host, cert, macaroon } = this
        const [encodeCert, encodeMacaroon] = await Promise.all([
          lndconnect.encodeCert(cert),
          lndconnect.encodeMacaroon(macaroon)
        ])
        const connectionstring = lndconnect.encode({
          host,
          cert: encodeCert,
          macaroon: encodeMacaroon
        })
        _lndConnect.set(this, connectionstring)
        return Promise.resolve(true)
    }
    catch {
        return Promise.resolve(true)
    }
}

_ready.set(this, isReady())
```

app/components/UI/QRCode.js
```diff
@@ -73,17 +73,19 @@ class ZapQRCode extends React.PureComponent {
...
-    bg: 'white'
+    bg: 'white',
+    border: 1
   }
 
   render() {
-    let { bg, color, size, theme } = this.props
+    let { bg, color, size, border, theme } = this.props
 
     if (size === 'small') {
```
korhaliv: Consider using `switch` with a `default `case to make this more robust and readable. Maybe it's a style preference though.

app/components/UI/QRCode.js
```diff
@@ -93,15 +95,30 @@ class ZapQRCode extends React.PureComponent {
     }
     if (size === 'large') {
       size = 180
+      border = 2
+    }
+    if (size === 'xlarge') {
```
korhaliv: Consider extracting this into a constant. `xlarge` is also used down below

app/lib/lnd/config.js
mrfelton: Have updated this to use `async`

app/components/UI/QRCode.js
mrfelton: refactored

app/components/UI/QRCode.js
mrfelton: Have extracted all preset size strings to constants.

app/components/Wallet/ReceiveModal/ReceiveModal.js
```diff
@@ -11,25 +11,46 @@ import messages from './messages'
...
+    const {
+      pubkey,
+      address,
+      activeWalletSettings: { lndConnect },
+      intl
+    } = this.props
+    const { qrCodeType } = this.state
+
+    let qrCode, notifTitle, notifBody
+    switch (qrCodeType) {
```
korhaliv: I would add `default` case also

app/components/Wallet/ReceiveModal/ReceiveModal.js
```diff
@@ -58,116 +80,99 @@ class ReceiveModal extends React.Component {
       return null
     }
 
-    let qrCode
+    let qrCode, message
     switch (qrCodeType) {
```
korhaliv: Would be nice to have `default` here too

app/components/Wallet/ReceiveModal/ReceiveModal.js
```diff
@@ -58,116 +80,99 @@ class ReceiveModal extends React.Component {
...
+          justifyContent="space-between"
+          alignItems="center"
+          width={1}
+          bg="secondaryColor"
+          p={4}
+        >
+          <Flex as="section" flexDirection="column" alignItems="center" width={1}>
+            <Heading.h1 textAlign="center">
+              {alias && alias.length ? alias : pubkey.substring(0, 10)}
+              {network && network.name.toLowerCase() === 'testnet' && ` (${network.name})`}
```
korhaliv: Looks like it can be simplified to this
` {network && network.name.toLowerCase() === 'testnet' && network.name}`

app/components/Wallet/ReceiveModal/ReceiveModal.js
```diff
@@ -58,116 +80,99 @@ class ReceiveModal extends React.Component {
...
+            <Box my={3}>{qrCode && <QRCode value={qrCode} size="xlarge" />}</Box>
+
+            <Box width={2 / 3}>
+              <Flex bg="tertiaryColor" justifyContent="space-between" width={1}>
+                <Text
+                  p={3}
+                  width={1}
+                  fontSize="s"
+                  textAlign="center"
+                  css={{
```
korhaliv: Consider moving this to styled-comp to keep jsx cleaner and shorter

app/components/Wallet/ReceiveModal/ReceiveModal.js
```diff
@@ -185,6 +190,7 @@ ReceiveModal.propTypes = {
   pubkey: PropTypes.string,
   address: PropTypes.string,
   alias: PropTypes.string,
+  activeWalletSettings: PropTypes.object,
```
korhaliv: consider using shape here to make it clear what type of object we are expecting here

app/components/UI/QRCode.js
```diff
@@ -73,35 +78,54 @@ class ZapQRCode extends React.PureComponent {
...
 
-    if (size === 'small') {
-      size = 120
-    }
-    if (size === 'medium') {
-      size = 150
-    }
-    if (size === 'large') {
-      size = 180
+    switch (size) {
```
korhaliv: Consider adding `default `case

app/components/Wallet/ReceiveModal/ReceiveModal.js
```diff
@@ -58,116 +80,99 @@ class ReceiveModal extends React.Component {
...
+        <Flex
+          justifyContent="space-between"
+          alignItems="center"
+          width={1}
+          bg="secondaryColor"
+          p={4}
+        >
+          <Flex as="section" flexDirection="column" alignItems="center" width={1}>
+            <Heading.h1 textAlign="center">
+              {alias && alias.length ? alias : pubkey.substring(0, 10)}
```
korhaliv: I've seen `pubkey.substring(0, 10)` in multiple places (for instance in Network.js). Should be more flexible and convenient if we extract this into an util. Like `getShortPubKey`

app/components/UI/QRCode.js
mrfelton: these props already have default values set via `defaultProps`

app/components/Wallet/ReceiveModal/ReceiveModal.js
mrfelton: Network name is supposed to be displayed in brackets after the node alias.

app/components/Wallet/ReceiveModal/ReceiveModal.js
mrfelton: The default case for these is that none of them are set, which prevents this QR code from displaying. If one of the predefined QR codes is set then all 3 of these values will be set as per this switch statement. So think the default case is implicit here.

app/components/Wallet/ReceiveModal/ReceiveModal.js
mrfelton: Have added a default message for this one

app/components/Wallet/ReceiveModal/ReceiveModal.js
mrfelton: I've updated this to use the `WalletName` component, so that the name displayed here matches elsewhere in the wallet (eg in the settings menu).

app/components/Wallet/ReceiveModal/ReceiveModal.js
mrfelton: done

mrfelton:
I think that is all of those issues addressed now @korhaliv

korhaliv:
utACK

PR #1317, opened 2019-01-11
fix(ui): implement themed delete wallet dialog

### LINKED ISSUE #1191: Implement new Delete Wallet confirmation dialog

korhaliv:
fix #1191


korhaliv:
Fair enough, was just trying to follow the mockup presented. Is there an extended mockup for this ? If not I can add that information the way I see it and update the PR.

korhaliv:
It's actually WIP. This is an old one without new Ole design.

korhaliv:
Good catch. Could barely notice it when testing (prob depends on rendering speed) but can be clearly seen in redux dev tools. Fixed

app/preload.js
```diff
@@ -83,56 +83,32 @@ function killLnd() {
...
+ */
+function getWalletDir(chain, network, wallet) {
+  return join(remote.app.getPath('userData'), 'lnd', chain, network, wallet)
+}
+
 /**
  * Delete a local wallet from the filesystem.
  */
-async function deleteLocalWallet(chain, network, wallet, force = false) {
+async function deleteLocalWallet(chain, network, wallet) {
```
mrfelton: if we are removing the `force` arg, then it should be removed from the `e2e` test suite where it was being used.

app/reducers/wallet.js
```diff
@@ -172,6 +190,18 @@ const walletSelectors = {}
...
 
+walletSelectors.activeWalletDir = createSelector(
+  activeWalletSelector,
+  walletsSelector,
+  (activeWallet, wallets) => {
+    const wallet = wallets.find(w => w.id === activeWallet)
+    if (wallet && wallet.type === 'local') {
+      return window.Zap.getWalletDir(wallet.chain, wallet.network, wallet.wallet)
+    }
+    return ''
```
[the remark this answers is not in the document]
korhaliv: done

app/preload.js
korhaliv: done

PR #1319, opened 2019-01-11
refactor(lnd): promisify lnd calls

korhaliv:
## Description:
Promisify lnd calls

## Motivation and Context:
Refactor lnd methods to use promisify. This reduces boilerplate code, makes it more compact and easier to read. Also simplifies addition of new methods in future

## How Has This Been Tested?
Manually. Tested functionality that involves corresponding lnd calls

## Types of changes:
Refactor

## Checklist:

### DIFF
--- a/app/lib/lnd/methods/channelController.js
+++ b/app/lib/lnd/methods/channelController.js
@@ -1,5 +1,6 @@
 import { listPeers, connectPeer } from './peersController'
 import pushopenchannel from '../push/openchannel'
+import { promisifiedCall } from '../../utils'
 
 function ensurePeerConnected(lnd, pubkey, host) {
   return listPeers(lnd).then(({ peers }) => {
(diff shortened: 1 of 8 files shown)


mrfelton:
@korhaliv this is a good improvement. I think you should extract the promisify wrapper method to a separate file and apply the same treatment across the rest of the lnd method proxies.

korhaliv:
Agree, I was going to apply the same kind of refactor in some other parts of the app. One thing that stopped me from extracting the method at this stage is because lnd expects us to pass args as an object (as opposed to spreading them). This may not be the case in other places, so need to check

mrfelton:
Have you seen https://github.com/carlessistare/grpc-promise @korhaliv ? Not saying that we should implement necessarily, but with a look to see what they are doing at least.

Also some discussion about this topic here https://github.com/grpc/grpc-node/issues/54

korhaliv:
No, I will check that. I prefer to do as much as possible using standard tools (like node promisify in this case). If you can't do something with a reasonable complexity with standard packages, only then I look for 3rd party packages. Also another thing is that particular package you are referring to doesn't seem very popular judging from a number of stars (may be not the best metric, but it's another thing in my checklist when looking for 3rd party packages).

mrfelton:
Agreed. no need for additional dependency here, but if there are issues you are facing the chances are that they have been solved by the people working on that package, so could be a good point of reference.

korhaliv:
Extracted `promisifiedCall`

korhaliv:
Promisified the entire lnd/methods folder

Issue #1195, opened 2018-12-29
Ability to close modal overlays by pressing escape key


mrfelton:
## Detailed Description

I should be able to close a modal overlay by pressing the escape key.

## Context

It's pretty standard to be able to close a modal overlay by hitting the escape key. We should support this to allow users to navigate the app more quickly.

mrfelton:
@bolatovumar it looks like this story wasn't properly linked to from the commit message of the PR that fixes it. Need to use one of the dedicated keywords in the commit message in order to have this issue get closed automatically. See https://help.github.com/articles/closing-issues-using-keywords/

Closing as this was resolved via #1321

PR #1321, opened 2019-01-13
feat(ui): enable modal close on escape key

bolatovumar:
Address #1195


bolatovumar:
This gets it working. I wasn't able to get it to work without using `document.addEventlistener` since adding a `onKeyPress` handler in JSX seems to get ignored on non-input elements?

Anyway, I think adding an event through `document.addEventlistener` should be fine here as the event is removed on `componentWillUnmount`.

app/components/UI/Modal.js
```diff
@@ -28,6 +28,27 @@ class Modal extends React.Component {
     withHeader: false
   }
 
+  constructor(props) {
```
mrfelton: This `constructor` method can be removed if you change the `closeOnEscape` method to be a fat arrow function:

```js
  closeOnEscape(e) {
    if (e.key === 'Escape') {
      const { onClose } = this.props

       onClose()
    }
  }
```
becomes: 
```js
  closeOnEscape = e => {
    if (e.key === 'Escape') {
      const { onClose } = this.props

      onClose()
    }
  }
```

app/components/UI/Modal.js
```diff
@@ -28,6 +28,27 @@ class Modal extends React.Component {
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.closeOnEscape)
+  }
+
+  closeOnEscape(e) {
+    if (e.key === 'Escape') {
+      const { onClose } = this.props
+
+      onClose()
```
mrfelton: `onClose` is not a required prop, and there are places where we display the modal without this prop set (such as in the onboarding process for brand new users). We need to ensure it's defined here before trying to call it.

app/components/UI/Modal.js
bolatovumar: Sure, I can do that.

app/components/UI/Modal.js
bolatovumar: Will do.

PR #1323, opened 2019-01-13
fix(btc): use correct parameter when converting from millisats to fiat

### LINKED ISSUE #1322: Non-existent variable used in btc.js on line 123
screen shot 2019-01-12 at 10 16 48 pm

The variable `satoshis` doesn't exist here. I'm guessing the `msats` parameter was supposed to be used instead.

bolatovumar:
Fix #1322


JimmyMow:
Concept ACK 5292312

PR #1337, opened 2019-01-14
fix(wallet): satoshi to millisatoshi conversion

### LINKED ISSUE #1332: Convertation from satoshis to millisatoshis incorrect?
I'm trying to write unit tests for btc.js and running into this issue:
screen shot 2019-01-13 at 3 04 21 pm

I expect 0.001 satoshis to equal to 1 millisatoshi (as per the [spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#human-readable-part)) but get 0.01 instead. It looks like the code thinks there are 10 millisatoshis per 1 satoshi when there are a 1000. Am I misunderstanding something here?

screen shot 2019-01-13 at 3 17 50 pm

Additional links: 
* [https://bitcoin.stackexchange.com/questions/69700/what-is-a-millisatoshi-i-thought-satoshi-is-the-smallest-unit](https://bitcoin.stackexchange.com/questions/69700/what-is-a-millisatoshi-i-thought-satoshi-is-the-smallest-unit)
* [https://en.bitcoin.it/wiki/Units#cite_note-4](https://en.bitcoin.it/wiki/Units#cite_note-4)

mrfelton:
## Description:

Millisatoshi should be satoshi * 1000 (we were converting satoshi to bits first, and then multiplying by 1000 which was resulting in an incorrect conversion.

## Motivation and Context:

Fix #1332

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/lib/utils/btc.js
+++ b/app/lib/utils/btc.js
@@ -84,7 +84,7 @@ export function satoshisToBits(satoshis) {
 export function satoshisToMillisatoshis(satoshis) {
   if (isEmptyAmount(satoshis)) return null
 
-  return satoshisToBits(satoshis) * 1000 || 0
+  return satoshis * 1000 || 0
 }
 
 export function satoshisToFiat(satoshis, price) {


app/lib/utils/btc.js
```diff
@@ -84,7 +84,7 @@ export function satoshisToBits(satoshis) {
 export function satoshisToMillisatoshis(satoshis) {
   if (isEmptyAmount(satoshis)) return null
 
-  return satoshisToBits(satoshis) * 1000 || 0
+  return satoshis * 1000 || 0
```
korhaliv: I wonder, why do we return `null` if amt is empty but use fallback ` || 0 `when something happens to `satoshis * 1000` ?
Probably not a big deal, but for consistency sake.

mrfelton: TBH, the fallback value probably isn't required here at all. I added it for the sake of consistency with the rest of this module since all the other methods do it that way and I don't want to extend the scope of this package to try and change that since personally I think this entire file needs to be reworked at some point.

PR #1339, opened 2019-01-14
feat(wallet): full support for lndconnect links

### LINKED ISSUE #1147: Add full support for lndconnect
## Detailed Description

Add full support for lndconnect, enabling users to connect Zap to their lnd nodes without needing to copy any files to their localhost.

https://github.com/LN-Zap/lndconnect

## Context

Currently we support a form of lndconnect in which the cert and macaroon are provided as paths to files. This means the user is expected to copy the cert and macaroon files to their local machine before being able to connect.

lndconnect allows for providing encoded cert and macaroons directly in the connection string, which would enable users to connect more easily.

## Possible Implementation

Enhance `ConnectionDetails` onboarding screen to support encoded cert/macaroon data in addition to paths to files.

mrfelton:
## Description:

Add full support for lndconnect links. This enables a 1 click connection process with remote nodes when used in conjunction with the `lndconnect` binary or anything else that generates lndconnect links.

**Note: This builds on the work in #1313 so that PR should be merged first, ideally.**

## Motivation and Context:

- Fix #1147
- Consolidate wallet config styles

## Task List

 - [x] Consolidate and standardize wallet config handling to use lndconnect format for all
 - [x] Write database migration to update existing wallet configs
 - [x] Write library to handle encoding/decoding/passing/formatting of lndconnect links (https://github.com/LN-Zap/node-lndconnect)
 - [x] Adapt lnd utils to support macaroons and certificates supplied as strings in addition to files
 - [x] Customise `ConnectionDetails` form when an lndconnect link has been used
 - [x] Provide a way to switch between the manual entry and lndconnect forms.
 - [x] Auto submit the lndconnect form as soon as it has been populated with a valid link
 - [x] Extract connection string input to field component
 - [x] Pretty print btcpay server config on paste
 - [x] Update test suite

## How Has This Been Tested?

1. Generate an `lndconnect` link on your remote node
2. Ctrl+click the link
3. Conform connection in Zap 

## Types of changes:

New feature

## Checklist:


app/components/Onboarding/Steps/ConnectionConfirm.js
```diff
@@ -58,10 +60,16 @@ class ConnectionConfirm extends React.Component {
...
       macaroon: connectionMacaroon
     }
+    // if we have a connection string, parse the connection details from that.
     if (connectionString) {
       const { host, port, macaroon } = parseConnectionString(connectionString)
       options = { type: connectionType, host: `${host}:${port}`, macaroon }
     }
+    // Otherwise, if we have a connection uri, parse the connection details from that.
+    else if (connectionUri) {
+      const { host, cert, macaroon } = decode(connectionUri)
```
korhaliv: We may want to use Error Boundaries here 
[https://reactjs.org/docs/error-boundaries.html](https://reactjs.org/docs/error-boundaries.html)
In case `decode` or `parseConnectionString` throw, which they, judging from their name could, the whole component tree will unmount and it will get ugly

mrfelton: What I have pushed so far was just the result of hacking until I had something that works. This whole PR is being reworked. I've read about Error Boundaries  before but not implemented them. I'll see how I can work into this in order to prevent the type of possible issue you mentioned.

korhaliv: Gotcha

app/components/UI/LndConnectionStringInput.js
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * Check for a valid BtcPayServer connection string.
+   * @param  {String}  value String to validate.
+   * @return {Boolean}       Boolean indicating wether the string is a valid or not.
+   */
+  isValidBtcPayConfig = value => {
+    let config = {}
+    try {
+      config = JSON.parse(value)
+      const configs = get(config, 'configurations', [])
```
korhaliv: Ideally configs should be named differently. like `allConfigs` to avoid very close looking name with `config`

app/components/UI/LndConnectionStringInput.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  /**
+   * Check for a valid BtcPayServer connection string.
+   * @param  {String}  value String to validate.
+   * @return {Boolean}       Boolean indicating wether the string is a valid or not.
+   */
+  isValidBtcPayConfig = value => {
+    let config = {}
```
korhaliv: `config` can be moved as a const inside `try`

app/lib/utils/btcpayserver.js
```diff
@@ -0,0 +1,19 @@
+import get from 'lodash.get'
+
+const parseConnectionString = value => {
+  let config = {}
+  try {
+    config = JSON.parse(value)
+  } catch (e) {
+    return new Error('Invalid connection string')
+  }
+  const configs = get(config, 'configurations', [])
```
korhaliv: Ideally configs should be named differently. like `allConfigs` to avoid very close looking name with `config`

app/lib/utils/btcpayserver.js
```diff
@@ -0,0 +1,19 @@
...
+  const params = configs.find(c => c.type === 'grpc' && c.cryptoCode === 'BTC') || {}
+  const { host, port, macaroon } = params
+  if (!host || !port || !macaroon) {
+    return new Error('Invalid connection string')
```
korhaliv: Do we really intend to return an `Error` here rather than `throw` ?

app/lib/utils/btcpayserver.js
```diff
@@ -0,0 +1,19 @@
...
+
+const parseConnectionString = value => {
+  let config = {}
+  try {
+    config = JSON.parse(value)
+  } catch (e) {
+    return new Error('Invalid connection string')
+  }
+  const configs = get(config, 'configurations', [])
+  const params = configs.find(c => c.type === 'grpc' && c.cryptoCode === 'BTC') || {}
```
korhaliv: May be a good idea to extract 'BTC' into a function argument (e.g `coin` or `network`) so it's more pronounced. And make it default to 'BTC'

app/lib/utils/sanitize.js
```diff
@@ -0,0 +1,17 @@
...
+}
+
+const sanitize = data => {
+  if (isObject(data)) {
+    const sanitizedData = { ...data }
+    const props = ['lndconnectUri', 'lndconnectQRCode']
+    props.forEach(key => {
+      sanitizedData[key] = truncate(data[key])
+    })
+  }
```
korhaliv: To me looks like this function doesn't actually do anything. I'd assume it meant to return `sanitizedData`. If that is the case another comment would be that `const sanitizedData = { ...data }` will destroy prototype data so in cases like with `lndConfig` instances it can lead to all sorts of obscure bugs

app/reducers/lnd.js
```diff
@@ -171,6 +176,12 @@ export function setStartLndError(errors) {
   }
 }
 
+export function clearStartLndError() {
```
korhaliv: Style comment. Other action creators use arrow notation here. So this can be shortened to 
`export const clearStartLndError = () => ({
    type: CLEAR_START_LND_ERROR
  })`

app/store/db.js
```diff
@@ -15,6 +18,39 @@ db.version(1).stores({
   nodes: 'id'
 })
 
+// Migrate custom wallets to lndconnect.
+db.version(2).upgrade(tx =>
+  tx.wallets.toCollection().modify(wallet => {
+    if (wallet.type === 'local') {
```
korhaliv: Could be a good idea to switch to `switch` or use `else` more consistently (there is one `if` and one `else if`

app/components/Onboarding/Steps/ConnectionDetailsString.js
```diff
@@ -0,0 +1,173 @@
...
+        {...rest}
+        getApi={formApi => {
+          this.setFormApi(formApi)
+          if (getApi) {
+            getApi(formApi)
+          }
+        }}
+        onChange={onChange && (formState => onChange(formState, currentItem))}
+        onSubmit={values => {
+          this.handleSubmit(values)
```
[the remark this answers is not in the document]
mrfelton: It's written this way deliberately. The idea is to make it clear that we are fist running our own onSubmit handler (`this.handleSubmit`) and then running `onSubmit` provided by `wizardApi`. This is a convention that we use for all of the Onboarding form Steps

app/components/Onboarding/Steps/ConnectionConfirm.js
```diff
@@ -43,26 +30,57 @@ class ConnectionConfirm extends React.Component {
...
+          host: connectionHost,
+          cert: connectionCert,
+          macaroon: connectionMacaroon
+        })
+      })
     }
-    if (connectionString) {
+
+    // If have a connection uri, parse the connection details from that.
+    else if (connectionString) {
```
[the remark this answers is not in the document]
mrfelton: updated

app/components/UI/LndConnectionStringInput.js
```diff
@@ -0,0 +1,106 @@
...
+   * Check for a valid lndconnect uri.
+   * @param  {String}  value String to validate.
+   * @return {Boolean}       Boolean indicating wether the string is a valid or not.
+   */
+  isValidLndConnectUri = value => {
+    let config = {}
+    try {
+      config = decode(value)
+      const { host } = config
+      if (host) {
```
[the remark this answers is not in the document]
mrfelton: Updated

app/components/UI/LndConnectionStringInput.js
```diff
@@ -0,0 +1,106 @@
...
+   * @return {Boolean}       Boolean indicating wether the string is a valid or not.
+   */
+  isValidBtcPayConfig = value => {
+    let config = {}
+    try {
+      config = JSON.parse(value)
+      const configs = get(config, 'configurations', [])
+      const params = configs.find(c => c.type === 'grpc' && c.cryptoCode === 'BTC') || {}
+      const { host, port, macaroon } = params
+      if (host && port && macaroon) {
```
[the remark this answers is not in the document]
mrfelton: Updated

app/lib/utils/btcpayserver.js
mrfelton: Good call. Much better to throw. Updated.

app/lib/utils/sanitize.js
mrfelton: I've updated this to return `sanitizedData`. The strange thing though was that it was working as expected without.

> If that is the case another comment would be that `const sanitizedData = { ...data }` will destroy prototype data so in cases like with `lndConfig` instances it can lead to all sorts of obscure bugs

This primary purpose of this method is to allow us to prevent sensitive data from being logged on the terminal, such as when we are logging the content of an `LndConfig` object.

`const sanitizedData = { ...data }` ensure that we modify a clone of the original object rather than modifying the original object directly, which we don't want to do.

We aren't intending to be left with a functional `LndConfig` object after running through this method.

app/lib/utils/btcpayserver.js
mrfelton: Good call. Have updated

app/components/UI/LndConnectionStringInput.js
```diff
@@ -0,0 +1,106 @@
...
+   * Check for a valid lndconnect uri or BtcPayServer connection string.
+   * @param  {String}  value String to validate.
+   * @return {Boolean}       Boolean indicating wether the string is a valid or not.
+   */
+  validate = value => {
+    const { intl } = this.props
+    const isValid = this.isValidLndConnectUri(value) || this.isValidBtcPayConfig(value)
+    if (!isValid) {
+      return intl.formatMessage({ ...messages.invalid_lnd_connection_string })
+    }
```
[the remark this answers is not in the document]
mrfelton: yes. This is how `informed` validation handlers work. They should either return an error message string, or `undefined` of there are no errors.

mrfelton:
Thanks for the review @korhaliv . I've made some adjustments based on your comments.

app/lib/utils/sanitize.js
korhaliv: > ensure that we modify a clone of the original object

That's not the question. The intent is obvious. Was just pointing out that there exists a side effect which removes a proto. If, as you say, this is util has a very narrow scope of use it'd be good to provide an extensive description/comment pointing out that specific, since it's placed in `app/lib/utils/sanitize` which is a top level util dir and those utils it can potentially be widely used across the app. Or another option is to move it somewhere closer to where it's actually being used.

app/components/UI/LndConnectionStringInput.js
mrfelton: Updated.

app/components/UI/LndConnectionStringInput.js
mrfelton: Updated.

app/lib/utils/sanitize.js
mrfelton: I've added some documentation to this method and included a note about the potential unintended side effects.

mrfelton:
Addressed the remainder of your points @korhaliv

PR #1342, opened 2019-01-14
refactor(ui): introduce checkbox component

korhaliv:
## Description:
Introduce `Checkbox `component

## Motivation and Context:
Allows to use checkbox in various pieces of UI

## How Has This Been Tested?
Via storybook

image

## Types of changes:
New UI component

## Checklist:


korhaliv:
Yes I was thinking about that too. Was not sure if this should be a part of this initial PR.

mrfelton:
> Yes I was thinking about that too. Was not sure if this should be a part of this initial PR.

I think it should be @korhaliv. Since all of our forms are handled in this way it makes sense that if we are introducing a new form element that it is ready to be used in conjunction with our forms.

korhaliv:
Ok let's do it

korhaliv:
Checkbox is now informed compatible

mrfelton:
@korhaliv here you can see the different states for the checkbox (see last 4 here)

image

normal: gray outline
hover: orange outline
selected: orange checked
disabled: grey with line through middle

mrfelton:
Here is a clearer version:

image

korhaliv:
Updated

mrfelton:
This is still showing a conflicted file `app/components/UI/index.js` which is preventing a merge

stories/forms/form.stories.js
```diff
@@ -17,7 +17,8 @@ import {
   Toggle,
   Radio,
   RadioGroup,
-  Range
+  Range,
+  Checkbox
```
mrfelton: I'd like to keep this list in alphabetical order to make it easier to scan and find things, and less subjective as to where to place new items - can you move this up to the right slot based on that please?

app/components/UI/Checkbox.js
```diff
@@ -0,0 +1,142 @@
...
+      width={18}
+      disabled={disabled}
+      color={disabled ? 'gray' : 'primaryColor'}
+      checked={checked}
+      alignItems="center"
+      justifyContent="center"
+    >
+      <Checkmark checked={checked} disabled={disabled} />
+    </CheckboxOutline>
+    <Text ml={2} color="gray">
```
mrfelton: I think the text colour should be `primaryColor` here. That would match the radios.

app/components/UI/Checkbox.js
```diff
@@ -0,0 +1,142 @@
...
+  </Container>
+)
+
+Checkbox.defaultProps = {
+  checked: false,
+  disabled: false
+}
+
+Checkbox.propTypes = {
+  text: PropTypes.string.isRequired,
```
mrfelton: The radios have a `label` and a `description` props. The label is the primary colour, whilst the description uses the gray colour. It would good if we could mirror that here.

At a minimum we should use the prop name `label` rather than `text` which would match other inputs, such as the Radios. It would be nice (although I don't think it's needed at this stage) if we could also add a `description`, which would show below the label in the more muted gray colour.

korhaliv: Changed text to label. We can also added description later if needed

app/components/UI/Checkbox.js
korhaliv: removed color entirely so it now uses default Text color (same as Radio)

stories/forms/form.stories.js
korhaliv: Sorted

app/components/UI/index.js
```diff
@@ -42,3 +42,4 @@ export Value from './Value'
 export Wizard from './Wizard'
 export Dialog from './Dialog'
 export ActionBar from './ActionBar'
+export Checkbox from './Checkbox'
```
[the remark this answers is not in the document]
korhaliv: Done

PR #1348, opened 2019-01-15
fix(wallet): show as unconfirmed transactions

### LINKED ISSUE #1344: Transaction modal copy bug
After sending an on-chain transaction I clicked on it to view the details. In the details the modal showed `Confirmed in Block: 0`. It was a fresh transaction that was unconfirmed. The app shouldn't show this line if it is unconfirmed or maybe like `Status: Confirmed`. I'd like to hear other opinions but I personally consider the current functionality a bug.

See screenshot:

screen shot 2019-01-14 at 9 14 40 pm

mrfelton:
## Description:

Show unconfirmed transactions as "Unconfirmed" rather than being in block 0.

## Motivation and Context:

Fix #1344 

## How Has This Been Tested?

Send an onchain transaction and then view the transaction in the transaction modal right after sending. Should show as `Confirmed in Block: Unconfirmed` rather than  `Confirmed in Block: 0`

image

## Types of changes:

Bug fix

## Checklist:


JimmyMow:
@mrfelton this may be picky but the copy still may be a bit misleading/weird. If it is unconfirmed what about `Status: Unconfirmed` and if confirmed then `Confirmed in block 123456`

mrfelton:
I've updated that @JimmyMow 

image

PR #1349, opened 2019-01-15
refactor(lib): apply notification.onClick properly

### LINKED ISSUE #1336: showNotification function accepts an onClick parameter which is used incorrectly
See [line 4 of notifications.js](https://github.com/LN-Zap/zap-desktop/blob/master/app/lib/utils/notifications.js#L4). The `onClick` parameter is being passed as an option to the `Notification` constructor but such an option does not exist. See here: [https://developer.mozilla.org/en-US/docs/Web/API/notification/Notification](https://developer.mozilla.org/en-US/docs/Web/API/notification/Notification).

An onclick property does exist but it should be assigned after the notification has been created. Like so:
screen shot 2019-01-13 at 6 48 07 pm

It doesn't look like this parameter ever gets passed in into this function from anywhere so I don't think there are any bugs caused by this.

mrfelton:
## Description:

Correctly apply `Notification.onClick` property. We don't actually use the onClick anywhere currently, but the current code would not work if we did.

## Motivation and Context:

Fix #1336 

## How Has This Been Tested?

Manually - verify all notifications work correctly

## Types of changes:

Refactor / Fix

## Checklist:

### DIFF
--- a/app/lib/utils/notifications.js
+++ b/app/lib/utils/notifications.js
@@ -1,8 +1,7 @@
 export const showNotification = (title, body, onClick) => {
-  new Notification(title, {
-    body,
-    onClick
-  })
+  const notification = new Notification(title, { body })
+
+  notification.onClick = onClick
 }
 
 export default { showNotification }


korhaliv:
Concept ACK [375001e](https://github.com/LN-Zap/zap-desktop/pull/1349/commits/375001e1fa9ecba3dc672e38eff23ae1a8931018)

PR #1352, opened 2019-01-15
perf(ui): reduce superfluous component renders

mrfelton:
## Description:

There are a number of places where the app is re-rendering things needlessly. Sometimes multiple times per second.

There are many more improvements that could be made, but to keep the scope of this PR relatively focused I have focused on a handful of the main offenders:

- Separate out the Wallet and Activity containers in order to reduce unnecessary re-renders of these components.
- Rather than calling describeNetwork periodically and replacing our entire node list with an updated list, subscribe to channel updates and only update node data for specific nodes that have been updated or added.
- reduce network pane re-renders by splitting out containers and connecting close to the source.
- prevent rerenders on balance fetch by using reselect to calculate the total balance

## Motivation and Context:

Improve app performance

## How Has This Been Tested?

Enable component update highlighting in react dev tools and connect to a wallet with several channels and transactions.

Compare renders before and after the patch.

## Types of changes:

Performance improvements

## Checklist:

### DIFF
--- a/app/reducers/channels.js
+++ b/app/reducers/channels.js
@@ -5,6 +5,7 @@ import { requestSuggestedNodes } from 'lib/utils/api'
 import { setError } from './error'
 import { fetchBalance } from './balance'
 import { walletSelectors } from './wallet'
+import { updateNodeData } from './network'
 
 // ------------------------------------
 // Constants
@@ -290,8 +291,13 @@ const throttledFetchChannels = throttle(dispatch => dispatch(fetchChannels()), 1
 export const channelGraphData = (event, data) => (dispatch, getState) => {
   const { info, channels } = getState()
   const {
-    channelGraphData: { channel_updates }
+    channelGraphData: { channel_updates, node_updates }
   } = data
+
+  // Process node updates.
+  if (node_updates.length) {
+    dispatch(updateNodeData(node_updates))
+  }
 
   // if there are any new channel updates
   let hasUpdates = false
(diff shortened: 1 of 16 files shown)


app/reducers/channels.js
```diff
@@ -290,9 +291,14 @@ const throttledFetchChannels = throttle(dispatch => dispatch(fetchChannels()), 1
 export const channelGraphData = (event, data) => (dispatch, getState) => {
   const { info, channels } = getState()
   const {
-    channelGraphData: { channel_updates }
+    channelGraphData: { channel_updates, node_updates }
   } = data
 
+  // Process node updates.
+  node_updates.forEach(nodeData => {
```
korhaliv: It maybe outside of scope of this PR, but later we may want to enhance`updateNodeData` to accept multiple nodes to mutate state in one go to avoid potential extra re-renders in case data comes in bulk.

korhaliv:
This looks like a solid perf PR @mrfelton !
I have just a couple of very minor comments

app/reducers/channels.js
mrfelton: Good idea, I'll have a look to see if this can be done easily. In my testing I have only ever see a single update come through at once, though given it is an array I assume it is possible for multiple updates to come through together.

app/reducers/network.js
```diff
@@ -164,6 +173,32 @@ export const receiveInvoiceAndQueryRoutes = (event, { routes }) => dispatch =>
 // Action Handlers
 // ------------------------------------
 const ACTION_HANDLERS = {
+  [UPDATE_NODE_DATA]: (state, { nodeData }) => {
```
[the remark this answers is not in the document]
mrfelton: I've updated this implementation. Much cleaner this way.

app/reducers/channels.js
mrfelton: Have updated to handle multiple node updates in a single dispatch

korhaliv: Looks good!

mrfelton:
Closing in favour of https://github.com/LN-Zap/zap-desktop/pull/1392, https://github.com/LN-Zap/zap-desktop/pull/1391 and https://github.com/LN-Zap/zap-desktop/pull/1390

PR #1374, opened 2019-01-17
fix(btc): make sure non-round fiat conversion results are handled well

### LINKED ISSUE #1334: btc to fiat conversion isn't handled well for non-round values
I'm writing unit tests for btc.js and getting this result when testing the `bitsToFiat` function:
screen shot 2019-01-13 at 3 47 23 pm

I would expect to see 1.58 as this makes sense for a fiat values. The issue is in the btcToFiat function on [line 32 ](https://github.com/LN-Zap/zap-desktop/blob/master/app/lib/utils/btc.js#L32). Some rounding to the nearest second decimal place should be done there.

bolatovumar:
Fix #1334

### DIFF
--- a/app/lib/utils/btc.js
+++ b/app/lib/utils/btc.js
@@ -29,7 +29,7 @@ export function btcToMillisatoshis(btc) {
 export function btcToFiat(btc, price) {
   if (isEmptyAmount(btc)) return null
 
-  return parseFloat(btc * price) || 0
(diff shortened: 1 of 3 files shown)


mrfelton:
@bolatovumar it looks like you need to update the test snapshots by running the tests with `yarn test-unit -u`

bolatovumar:
Pushed the branch with updated snapshots.

PR #1378, opened 2019-01-17
fix(ui): highlight border of selected inputs

mrfelton:
## Description:

 - Ensure that input field borders are correctly highlighted when you click into them.
 - Refactor `Input` and `TextArea` fields to be more consistent
 - Extract some common input behavior out into an HOC

## Motivation and Context:

This fixes an issue with the inputs where by they don't properly get highlighted when you click in to them.

## How Has This Been Tested?

 - Added several more examples to the storybook.
 - Test behavior of various input fields throughout the app

**Default state Before:**
initial-before

**Default state After:**
initial-after

**With errors Before:**
with-error-before

**With errors  After:**
with-error-after

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/components/UI/Input.js
+++ b/app/components/UI/Input.js
@@ -2,38 +2,83 @@
 
 import React from 'react'
 import { asField } from 'informed'
-import { withTheme } from 'styled-components'
-import system from '@rebass/components'
+import styled, { withTheme } from 'styled-components'
 import { styles } from 'styled-system'
-import * as yup from 'yup'
+import system from '@rebass/components'
 import { Flex } from 'rebass'
 import { Message, Label, Span, Text } from 'components/UI'
+import withRequiredValidation from 'components/withRequiredValidation'
+
+function isFieldValid({ value, error, asyncError, touched }) {
+  return value && !error && !asyncError && touched
+}
+
+function mapDefaultBorderColor(props) {
+  const {
+    disabled,
+    readOnly,
+    fieldState,
+    fieldState: { error, asyncError },
+    theme: {
+      colors: { gray, superGreen, superRed }
+    }
+  } = props
+
+  let borderColor = gray
+  if (readOnly || disabled) {
+    borderColor = gray
+  } else if (error || asyncError) {
+    borderColor = superRed
+  } else if (isFieldValid(fieldState)) {
+    borderColor = superGreen
+  }
+  return borderColor
+}
+
+function mapFocusBorderColor(props) {
+  const {
+    fieldState,
+    theme: {
+      colors: { lightningOrange, superGreen }
+    }
+  } = props
+  return isFieldValid(fieldState) ? superGreen : lightningOrange
+}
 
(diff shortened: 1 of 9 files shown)


app/components/UI/Input.js
```diff
@@ -140,12 +151,20 @@ class Input extends React.Component {
             )}
           </Label>
         )}
+
         <SystemInput
+          opacity={readOnly ? 0.6 : null}
```
korhaliv: Probably would be a good idea to extract this into a styled-comp to keep it cleaner

mrfelton: I've moved the `outline` and `opacity ` style props over to the styled component.

Unfortunately, I can't move the dynamic border color props to the same place as the utility methods that calculate the correct color  need access to the `formState` prop, which doesn't get passed on to the styled component. So I've had to leave those applied via the `css` prop.

korhaliv: Maybe I'm missing something but why can't you path form props down to styled component and map it with the same function you already have?

mrfelton: The styled component in this case is a rebas system component, wrapped in a styled component (see https://github.com/LN-Zap/zap-desktop/pull/1378/files#diff-9406bc9468bcaded7a7aec88441c833eR49), which outputs an `input` dom element.

For example, here is a stripped down version of what we have (psudocode)

First we create a system component, and wrap it with a rebass system component.
```js
const StyledInput = styled(
  system(
    {
      as: 'input',
      border: 1,
      borderColor: 'gray',
      borderRadius: 5,
    },
    'borders',
    'borderColor',
    'borderRadius',
  )
)`
  opacity: ${props => (props.disabled || props.readOnly ? '0.6' : 'inherit')};
  outline: none;
`
```

Then we use it in our Input component:

```js
// Integrate the field with informed.
const Input = asField(({
   fieldApi,
   fieldState,
   ...rest
 }) => {
return {
  <StyledInput
    css={Object.assign(
      {
        'border-color': mapDefaultBorderColor(this.props),
        '&:not([readOnly]):not([disabled]):focus': {
          'border-color': mapFocusBorderColor(this.props)
        }
      },
      css
    )}
    {...rest}
    />
})
```

`mapDefaultBorderColor` needs access to things that are not being passed on to the `StyledInput`, such as `formState`.

If I pass additional props such as `formState` to the styled component, that are not intended to end up on the underlaying dom element (`input` in this case) I will get warnings about unknown properties being set in the dom.

I'm not sure how I could pass the props to the styled component purely for the purpose of calculating the css value, and then strip those properties out so that they don't get passed onto the dom. I could possibly do it by introducing another intermediary component but it seems like overkill.

Am I missing something?

mrfelton: @korhaliv I've updated this as per our slack chat. Works like a charm :)

korhaliv:
Tested ACK [3b4f586](https://github.com/LN-Zap/zap-desktop/commit/3b4f5869752583a6f6b6280f05d1243e8a170782)

PR #1393, opened 2019-01-18
refactor(ui): decompose network jsx into components

korhaliv:
## Description:
Refactor `Network.js` jsx code

## Motivation and Context:
Decompose `Network.js` jsx code into manageable parts in order to be able to easily modify/improve/do bugfix in the future

## How Has This Been Tested?
Manually

## Types of changes:
Refactor

## Checklist:

### DIFF
--- a/app/components/Contacts/Network/BrowseChannelButton.js
+++ b/app/components/Contacts/Network/BrowseChannelButton.js
@@ -1,0 +1,16 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import ExternalLink from 'components/Icon/ExternalLink'
+import { Button } from 'components/UI'
+
+const BrowseChannelButton = ({ onClick }) => (
+  <Button variant="secondary" size="small" ml="auto" px={0} py={0} onClick={onClick}>
+    <ExternalLink />
+  </Button>
+)
+
+BrowseChannelButton.propTypes = {
+  onClick: PropTypes.func.isRequired
+}
+
+export default BrowseChannelButton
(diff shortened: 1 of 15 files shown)


app/components/Contacts/Network/BrowseChannelButton.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import ExternalLink from 'components/Icon/ExternalLink'
+import { Button } from 'components/UI'
+
+const BrowseChannelButton = ({ isVisible, onClick }) =>
+  isVisible && (
```
mrfelton: Why add the `isVisible` prop to each component? It seems a non-reacty thing to do. Shouldn't it be the parent sets wether the prop renders of not in its jsx markup?

```jsx
{isVisible && <BrowseChannelButton />}
```

The `isVisible` seems redundant

korhaliv: Hm, maybe you are right, but I find
`{isVisible && <BrowseChannelButton />}`
Not so readable as 
`<BrowseChannelButton isVisible={isVisible}/>`
because of those curly braces
I understand your concern that it's like a black box and you don't know how this flag is handled by each component. Is this what concerns you?

mrfelton: Yes, pretty much. It requires the user of the component to know about the prop, what it is and what it does. It means that readers of the parent component don't inherently know that one of it's children will not be rendered because all you can see is that it is included in the markup and has a bunch of props - but what do those props do?

With jsx it is immediately obvious to the reader what will render and what will not and there is no guesswork.

It's also a bit inconsistent for us to handle visibility in this way for these components when we don't do that anywhere else in the codebase.

korhaliv: Ok, I understand your point. I will extract those props back

PR #1398, opened 2019-01-19
fix(btc): use rounding in fiat conversions

### LINKED ISSUE #1334: btc to fiat conversion isn't handled well for non-round values

mrfelton:
## Description:

When converting crypto amounts to fiat, use rounding rather than string conversion to ensure that the result remains as a Number type.

Fix #1334

## Motivation and Context:

In #1374 we changed how we handle conversions to fiat to use `toFixed`, which results in a String. We want to the the result as a number so that we can work with it as a number. Conversion top string with fixed dp should happen at display time rather than at time of conversion.

## How Has This Been Tested?

This fixes a bug that has broken the Request page. Currently in master, the app crashes when trying to create an invoice because in the `Value` component (which we use for displaying value amount) we convert numbers to the correct number of decimal places. But in the case of fiat the value is already a string. This causes a fatal error since `toString` can not be called on a string.

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/lib/utils/btc.js
+++ b/app/lib/utils/btc.js
@@ -29,7 +29,7 @@ export function btcToMillisatoshis(btc) {
 export function btcToFiat(btc, price) {
   if (isEmptyAmount(btc)) return null
 
-  return (btc * price).toFixed(2)
+  return Math.round(btc * price * 100) / 100
 }
 
 ////////////////////////////
--- a/test/unit/components/Pay/__snapshots__/PaySummaryLightning.spec.js.snap
+++ b/test/unit/components/Pay/__snapshots__/PaySummaryLightning.spec.js.snap
@@ -55,7 +55,7 @@ exports[`component.Form.PaySummaryLightning should render correctly 1`] = `
           <FormattedNumber
             currency="USD"
             style="currency"
-            value="0.65"
+            value={0.65}
           />
         </Text>
       </styled.div>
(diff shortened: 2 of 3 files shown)


mrfelton:
@bolatovumar interested to get your thoughts on this change

mrfelton:
@bolatovumar we already have the [`Value`](https://github.com/LN-Zap/zap-desktop/blob/master/app/components/UI/Value.js) component that we use when rendering amounts. See

Also, we use the [`FormattedNumber`](https://github.com/yahoo/react-intl/wiki/Components#number-formatting-components) component from `react-intl` when rendering fiat values, which supports rendering currency values according to locale specific rules.

In Slack, we also discussed using [bignumber.js](https://github.com/MikeMcl/bignumber.js/) to better handle the arithmetic when converting from one currency to another which would probably give a better degree of control and mitigate against issues with rounding errors.

korhaliv:
Tested ACK [e961d60](https://github.com/LN-Zap/zap-desktop/pull/1398/commits/e961d60be963c8d88626da7dab271e3234e0dd9c)

PR #1403, opened 2019-01-19
refactor(lint): fix lint warnings

korhaliv:
## Description:
Address lint warnings

## Motivation and Context:
Fix nested promise(and improve it's readability along the way) and import warnings. Silence proptype warning.

## How Has This Been Tested?
Request invoice functionality, one of the parts this PR addresses, was tested manually

## Before
` 6 warnings`
## After
`no warnings`
## Types of changes:
Refactor

## Checklist:

### DIFF
--- a/app/lib/lnd/methods/index.js
+++ b/app/lib/lnd/methods/index.js
@@ -143,34 +143,31 @@ export default function(lnd, log, event, msg, data) {
         .catch(error => log.error('balance:', error))
       break
     case 'createInvoice':
-      // Invoice looks like { r_hash: Buffer, payment_request: '' }
-      // { memo, value } = data
       invoicesController
         .addInvoice(lnd, data)
         .then(newinvoice =>
-          invoicesController
-            .getInvoice(lnd, { pay_req: newinvoice.payment_request })
-            .then(decodedInvoice =>
-              event.sender.send(
-                'createdInvoice',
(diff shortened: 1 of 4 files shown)


app/lib/lnd/methods/index.js
```diff
@@ -5,6 +5,33 @@ import * as peersController from './peersController'
 import * as paymentsController from './paymentsController'
 import * as networkController from './networkController'
 
+const createInvoice = async ({ lnd, data, event, log }) => {
```
mrfelton: This is a nice refactor.

My only slight issue is that this is the only method that that we declare at the top of the file. All of the others are defined inline.

This file is a complete mess as it is anyway, and we really need to look at refactoring this whole thing. But, until we do that I think we should try to keep the consistency at least.

Could you move the definition so that it sits inline where it is used in the big switch statement?

korhaliv: Technically I can, but there is a couple of things:
1. I thought it's cleaner when you move that method outside of `switch` to start reducing it. So this could be a start of the refactor you are talking about
2. There is a lint rule that doesn't allow me to put it inside  `Unexpected lexical declaration in case block.  no-case-declarations` I can of course silence it if we decide to put this inside.

korhaliv: Updated as per slack discussion

PR #1431, opened 2019-01-26
fix(lnd): support latest version of lnd

mrfelton:
## Description:

Update internal lnd.

Store multiple rpc.proto files and select the most appropriate one based on the version of lnd that we are connecting to. Adapt api calls as needed in order or support backwards incompatible changes in lnd.

## Motivation and Context:

Support latest version of lnd

## How Has This Been Tested?

Manually

## Types of changes:

Lnd Update

## Checklist:


app/lib/lnd/lightning.js
```diff
@@ -87,57 +98,40 @@ class Lightning {
...
-            longs: Number,
-            enums: String,
-            defaults: true,
-            oneofs: true
+        .then(() => {
+          if (type === 'local') {
+            return waitForFile(macaroon, 20000)
           }
-          const packageDefinition = await load(rpcProtoPath, options)
+          return
```
korhaliv: is there a reason why we have `return` here and in a couple of other places? If we omit return it will still return undefined

app/lib/lnd/util.js
```diff
@@ -184,3 +188,86 @@ export const waitForFile = (filepath, timeout = 1000) => {
...
+}
+
+/**
+ * Find the closest matching rpc.proto version based on an lnd version string.
+ * @param  {[type]}  info [description]
+ * @return {Promise}      [description]
+ */
+export const getClosestProtoVersion = async versionString => {
+  let version = versionString.split(' ')[0]
+  const commitString = versionString.split(' ')[1]
```
korhaliv: consider using `const[version , commitString] = versionString.split(' ')`

app/lib/lnd/walletUnlocker.js
```diff
@@ -115,6 +87,39 @@ class WalletUnlocker {
...
+    const rpc = loadPackageDefinition(packageDefinition)
+
+    // Create ssl credentials to use with the gRPC client.
+    const sslCreds = await createSslCreds(cert)
+
+    // Create a new gRPC client instance.
+    this.service = new rpc.lnrpc.WalletUnlocker(host, sslCreds)
+
+    // Wait upto 20 seconds for the gRPC connection to be established.
+    return new Promise((resolve, reject) => {
```
korhaliv: Looks like this can be promisified to avoid `new Promise`

app/lib/lnd/walletUnlocker.js
```diff
@@ -115,6 +87,39 @@ class WalletUnlocker {
   // Helpers
   // ------------------------------------
 
+  /**
+   * Establish a connection to the Lightning interface.
+   */
+  async establishConnection() {
+    const { host, cert } = this.lndConfig
```
korhaliv: This looks like a complicated method. What happens if we get an exception somewhere there?  Is it handled somewhere so the user is aware that something went wrong?

app/lib/lnd/lightning.js
```diff
@@ -173,11 +167,48 @@ class Lightning {
...
   // ------------------------------------
 
+  /**
+   * Establish a connection to the Lightning interface.
+   */
+  async establishConnection(version: ?string) {
+    const { host, cert, macaroon } = this.lndConfig
+
+    // Find the rpc.proto file.. If no version was supplied, attempt to use the latest version.
+    this.version = version || (await getLatestProtoVersion())
```
korhaliv: can't find where else `this.version` is used other than the next line. probably somewhere outside this module. In that case it may be useful to convert it into a getter in order to  have more control over it's value. If it's not used anywhere outside and is not meant to be used as such, consider converting it into a local const

app/lib/lnd/lightning.js
mrfelton: Our lint rules require an explicit return inside of promise chains.

app/lib/lnd/walletUnlocker.js
mrfelton: This PR is off master which doesn't have the new `promisifiedCall` method. Given that the PR to add that was fairly large and we want to keep changes to a minimum in master, I'd prefer to leave this as it is for now.

Note, I didn't write this method in the PR - I have just pulled it out from another method where it was previously embedded to make the code a little clearer.

app/lib/lnd/walletUnlocker.js
mrfelton: This class is fairly low level, and the caller of this method is expected to handle any errors from it.

We do handle these already and ensure that we pass on appropriate errors to the client where possible. For example, see `onBeforeStartRemoteLnd` where we parse some of the known errors from failed connection attempts.

app/lib/lnd/lightning.js
mrfelton: Correct - this is used outside of this module. I will make it a readonly getter.

mrfelton: Updated.

app/lib/lnd/util.js
mrfelton: done

mrfelton:
@JimmyMow that is what happens when you use a newer `rpc.proto` file with an older lnd. Could you run with `DEBUG_LEVEL=debug DEBUG=zap*` and share the first part of your startup logs. It should show exactly which `rpc.proto` file is being selected.

mrfelton:
@JimmyMow I found the cause of this issue. The problem is that lnd doesn't actually print it's version string in a fully valid semver format. For example, take this version string from the lnd version that you are connecting to:

```
0.5.1-beta commit=v0.5.1-beta-40-gfc4fe070100a66ea220a14e74f52f9c9d1550636
```

If you try to extract and parse a semver from that string, you would get:

```
major: 0
minor: 5
patch: 1
prerelease: beta-40-gfc4fe070100a66ea220a14e74f52f9c9d1550636 
```

The problem here is that the prerelease identifier is a single simple string, but in reality this would be better represented as two unique prerelease components `beta` and `40` (the commit hash should be added as a build identifier).

So, the fully valid semver should be more like `v0.5.1-beta.40+gfc4fe070100a66ea220a14e74f52f9c9d1550636`

This is the difference between `0.5.1-beta-40` and `0.5.1-beta.40`. When comparing versions in the former case prerelease is being treated as a single string `beta-40`. In the later case, the prerelease would be compared as 2 unique components `beta` and `40` which enables us to do more accurate version matching.

I have updated our processing of the version string to coerce the string into a more accurate semver and I have written some unit tests to prove that our processing of this data works as expected with various different example version strings as printed by lnd.

mrfelton:
Please check again @JimmyMow - have made some updates to resolve this.

Issue #1425, opened 2019-01-24
Implement new channel (summary) list view


JimmyMow:
Implement the newly designed channel management (card) list view.

Design can be found here: 
https://app.zeplin.io/project/5bacdbad33af886d419f1b34/screen/5c4b53e31f32ea378b51d212

mrfelton:
All PR's relating to the summary view have been merged into `next`, so closing this issue now.

PR #1439, opened 2019-01-28
 refactor(ui): add new ChannelSummaryListItem component

mrfelton:
## Description:

Add a new ChannelSummaryListItem component that can be used to display a summary of channel data in a list view format.

## Motivation and Context:

Re #1425

## How Has This Been Tested?

Storybook

image

## Types of changes:

New components

## Checklist:

### DIFF
--- a/app/components/Channels/ChannelMoreButton.js
+++ b/app/components/Channels/ChannelMoreButton.js
@@ -1,0 +1,11 @@
+import React from 'react'
+import { FormattedMessage } from 'react-intl'
+import { Button } from 'components/UI'
+import messages from './messages'
+
+const ChannelMoreButton = props => (
+  <Button type="button" size="small" {...props}>
+    <FormattedMessage {...messages.more_button_text} />
+  </Button>
+)
+export default ChannelMoreButton
--- a/app/components/Channels/ChannelSummaryListItem.js
+++ b/app/components/Channels/ChannelSummaryListItem.js
@@ -1,0 +1,65 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Box, Flex } from 'rebass'
+import { Card, Heading, Text } from 'components/UI'
+import { ChannelCapacity, ChannelMoreButton, ChannelStatus } from 'components/Channels'
+import withEllipsis from 'components/withEllipsis'
+
+const ClippedHeading = withEllipsis(Heading.h1)
+const ClippedText = withEllipsis(Text)
+
+const ChannelSummaryListItem = ({
+  channelId,
+  channelName,
+  channelPubKey,
+  channelPubKeyShort,
+  cryptoCurrency,
+  cryptoCurrencies,
+  localBalance,
+  remoteBalance,
+  status,
+  setCryptoCurrency,
+  showChannelDetail,
+  ...rest
+}) => (
+  <Card {...rest}>
+    <Flex>
+      <Box width={8 / 20}>
+        <ChannelStatus status={status} />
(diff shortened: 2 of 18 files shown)


app/components/Channels/ChannelSummaryListItem.js
```diff
@@ -0,0 +1,70 @@
...
+  cryptoCurrency,
+  cryptoCurrencies,
+  localBalance,
+  remoteBalance,
+  status,
+  setCryptoCurrency,
+  showChannelDetail,
+  ...rest
+}) => {
+  return (
```
JimmyMow: can remove the return here and make this a pure component

mrfelton: Updated.

app/components/Channels/ChannelSummaryListItem.js
```diff
@@ -0,0 +1,68 @@
...
+        remoteBalance={remoteBalance}
+        cryptoCurrency={cryptoCurrency}
+        cryptoCurrencies={cryptoCurrencies}
+        setCryptoCurrency={setCryptoCurrency}
+        width={1 / 2}
+        pl={4}
+        pr={4}
+      />
+
+      <Flex width={2 / 20} alignSelf="stretch">
```
korhaliv: I think `alignSelf="stretch"` can be safely removed as it basically a default

app/components/Channels/ChannelMoreButton.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { FormattedMessage } from 'react-intl'
+import { Button } from 'components/UI'
+import messages from './messages'
+
+const ChannelMoreButton = ({ showChannelDetail, channelId, ...rest }) => (
+  <Button type="button" size="small" onClick={() => showChannelDetail(channelId)} {...rest}>
```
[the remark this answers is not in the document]
mrfelton: I've removed these 2 props completely in favour of just using the `onClick` prop in on the caller.

app/components/withEllipsis/withEllipsis.js
```diff
@@ -0,0 +1,34 @@
...
+      children: PropTypes.node
+    }
+
+    render() {
+      const { children, ...rest } = this.props
+
+      return (
+        <Component
+          css={{
+            overflow: 'hidden',
```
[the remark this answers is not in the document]
mrfelton: Updated.

app/components/Channels/ChannelSummaryListItem.js
```diff
@@ -0,0 +1,68 @@
...
+  status,
+  setCryptoCurrency,
+  showChannelDetail,
+  ...rest
+}) => (
+  <Card {...rest}>
+    <Flex>
+      <Box width={8 / 20}>
+        <ChannelStatus status={status} />
+        <ClippedHeading my={1}>{channelName || channelPubKey.substring(0, 10)}</ClippedHeading>
```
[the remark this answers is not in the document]
mrfelton: Actually, I had reworked this already in a later commit but didn't include in this PR as it's outside of the scope of what I'm doing here.

The approach that I have taken is to inject a new `pubKeyShort` prop into the channel object in the reducers/selectors so that we already have a prop available for this and can just use it directly.

I've updated the component in this PR to use a dedicated prop for the short pubkey rather than truncating within the component render method.

app/components/Channels/ChannelSummaryListItem.js
mrfelton: removed.

PR #1447, opened 2019-01-29
refactor(ui): add new ChannelSummaryList component

mrfelton:
## Description:

Add a new ChannelSummaryList component that can displays a list of channels in a summary view.

## Motivation and Context:

Needed for new Channel UI - Re #1425

## How Has This Been Tested?

Storybook

image

## Types of changes:

New components

## Checklist:

### DIFF
--- a/app/components/Channels/ChannelSummaryList.js
+++ b/app/components/Channels/ChannelSummaryList.js
@@ -1,0 +1,52 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Box } from 'rebass'
+import { ChannelSummaryListItem } from 'components/Channels'
+
+const ChannelSummaryList = ({
+  channels,
+  cryptoCurrency,
+  cryptoCurrencies,
+  setCryptoCurrency,
+  showChannelDetail,
+  ...rest
+}) => (
+  <Box as="article" {...rest}>
+    {channels.map((channelObj, index) => {
+      const channel = channelObj.channel || channelObj
+
+      return (
+        <ChannelSummaryListItem
+          key={channel.chan_id || `${channel.remote_pubkey_short}-${index}`}
+          cryptoCurrency={cryptoCurrency}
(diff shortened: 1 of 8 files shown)


app/components/Channels/ChannelSummaryList.js
```diff
@@ -0,0 +1,54 @@
...
+}) => (
+  <Box as="article" {...rest}>
+    {channels.map((channelObj, index) => {
+      const channel = Object.prototype.hasOwnProperty.call(channelObj, 'channel')
+        ? channelObj.channel
+        : channelObj
+
+      return (
+        <ChannelSummaryListItem
+          key={index}
```
korhaliv: Can we use something other than index? `chan_id` looks good for this purpose. Using indexes as a key is generally an antipattern

app/components/Channels/ChannelSummaryList.js
```diff
@@ -0,0 +1,54 @@
...
+  channels,
+  cryptoCurrency,
+  cryptoCurrencies,
+  setCryptoCurrency,
+  showChannelDetail,
+  ...rest
+}) => (
+  <Box as="article" {...rest}>
+    {channels.map((channelObj, index) => {
+      const channel = Object.prototype.hasOwnProperty.call(channelObj, 'channel')
```
korhaliv: Is there a need for `hasOwnProperty` vs `in` here ? If we indeed don't want to take into account proto property then maybe it's worth to add a comment why

mrfelton: Have simplified this.

app/components/Channels/ChannelSummaryList.js
mrfelton: `chan_id` isn't always set (not set for pending nodes for example). I've created a new key which is a combination of a few different things which I think should be guaranteed to be unique.

PR #1450, opened 2019-01-29
perf(activity): optimize activity list rendering

### LINKED ISSUE #1296: UI elements lag when transaction count grows large
All UI elements lag when transaction count gets into the thousands.

## Description

I have over 5000 transactions now on one LND node. When connecting to this node, Zap becomes unresponsive while it apparently loads the transactions (for about 5 seconds), and then all UI elements become constantly laggy e.g. the menu drop down, the add new channel button. It takes 3 seconds to open a drop down, about a second for the cursor to change from pointer to hand, etc.

When switching to other LND nodes (with far fewer transactions) using the same ZAP instance, the UI returns to its normal responsive self.

## Expected Behavior

UI Elements should be responsive regardless of trransaction count.

## Actual Behavior

UI elements lag when there is a large transaction count.

## Possible Fix

Perhaps some sort of paging for transactions would stop the UI from slowing down.

## Steps to Reproduce

- 1. Connect to a LND node with more than 5,000 transactions
- 2. Click the UI Elements and observe the unresponsiveness.

## Context

Managing a node with a large number of transactions. The node also has a reasonably large number of channels (around 70) so potentially that could be the problem too, but it is probably more likely the large transaction count.

## Your Environment

- Zap Desktop version: 0.3.2.0
- Operating System and version: Windows 10 Build 17763

korhaliv:
## Description:
Optimize rendering of activity list

## Motivation and Context:
fix #1296 
Deploy [react-virtualized](https://github.com/bvaughn/react-virtualized) library to optimize rendering process and also optimize reselect selectors for flat structure and faster initial fetching

## How Has This Been Tested?
Manually
Created a remote node with 6000 txs and tested performance in production. It is important to test performance in production mode since react/redux devtools slow down rendering a lot and eventual results are distorted a lot.

## Types of changes:
Optimization

## Checklist:

### DIFF
--- a/app/components/Activity/Activity.js
+++ b/app/components/Activity/Activity.js
@@ -1,6 +1,9 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space } from 'styled-system'
 import debounce from 'lodash.debounce'
+import { List, AutoSizer, CellMeasurer, CellMeasurerCache } from 'react-virtualized'
 import { FormattedMessage, injectIntl, FormattedDate } from 'react-intl'
 import { Box, Flex } from 'rebass'
 import { Bar, Button, Form, Heading, Input, Panel, Spinner, Text } from 'components/UI'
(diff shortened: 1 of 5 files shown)


app/components/Activity/Activity.js
```diff
@@ -10,15 +12,31 @@ import X from 'components/Icon/X'
 import messages from './messages'
 import ActivityListItem from './ActivityListItem'
 
+const StyledList = styled(List)`
+  outline: none;
+  padding-right: 45px;
```
mrfelton: Can we do this using our theme spacing units rather than custom values? You could add the styled-system spacing props onto this element like so:

```js
import styled from 'styled-components'
import { space } from 'styled-system'

const StyledList = styled(List)(space)
```

Which will allow you to use all of our themes predefined spacing properties rather than custom units. (`<StyledList  pr={5}/>` would be equivalent to `padding-right: 45px;` for example), and would allow those props to be applied by the caller rather than being preset on the element definition.

See https://github.com/jxnblk/styled-system

korhaliv:
Updated

PR #1453, opened 2019-01-30
feat(ui): implement wallet settings save

### LINKED ISSUE #1190: Implement new Wallet Settings actions bar (save, cancel)
## Detailed Description

Users should be able to save changes to wallet configurations without launching the wallet.

## Context

It's not convenient or obvious that you need to launch a wallet in order to save changes to it's configuration.

## Possible Implementation

Implement new Lightning Payment Received modal based on Zeplin design:
zpl://screen?sid=5c065f4f7324832ffb782ff2&pid=5bacdbad33af886d419f1b34

image

korhaliv:
## Description:
fix #1190 

## How Has This Been Tested?
Manually

image

## Types of changes:
New UI feature

## Checklist:


app/components/Home/WalletLauncher.js
```diff
@@ -69,66 +131,143 @@ class WalletLauncher extends React.Component {
...
+
+    let updatedSettings = values
+
+    // if autopilot is off and no defaults are present, ensure they are included into config
+    // to properly fill in the form next time user decides to edit settings.
+    if (wallet.type === 'local' && !values.autopilot) {
+      updatedSettings = Object.assign({}, autopilotDefaults, values)
+    }
+
+    putWallet(formToWalletFormat(updatedSettings))
```
mrfelton: With this, we will always be saving the default options onto the wallet config when the wallet form, is saved. This means that if the default settings are changed in the future (eg, when lnd increase the default max channel size) users that created wallets prior to this will not assume the new defaults as the old defaults will have been hardcoded into their wallet config.

app/preload.js
```diff
@@ -95,16 +95,30 @@ function getWalletDir(chain, network, wallet) {
...
-async function deleteLocalWallet(chain, network, wallet) {
-  try {
-    assert(chain && network && wallet)
-  } catch (err) {
-    throw new Error(`Unknown wallet: (chain: ${chain}, network: ${network}, wallet: ${wallet}`)
+async function deleteLocalWallet({ chain, network, wallet, dir }) {
+  // returns wallet location based on arguments configuration
+  const getDir = () => {
+    if (typeof dir === 'string') {
+      return dir
```
mrfelton: This has some security considerations.

This method gets added to the global `window` namespace and essentially exposes the `rimraf` (`rm -rf`) method.

If a malicious user were able to gain  the ability to execute js code (eg, by some kind of injection exploit) then they would have the ability to force remove any directory on the system.

The previous version of this method attempted to mitigate that possibility by only allowing the deletion of wallet directories (although I'm not confident that the mitigation was foolproof as you could potentially run it like:

```js
deleteLocalWallet({ chain: 'bitcoin', network: 'testnet', wallet: '../../../../../../../../'})`
```

I have not tested this but I suspect it may work.

In this updated version of the code however, the risk is even more pronounced as a user could run the following and trash the file system.

```js
deleteLocalWallet({dir: '/'})
```

Also untested, but it looks to be the case.

Can you think of a way that we can provide some additional security guarantees around this?

One thing we could consider is to only expose this method on `window` if `NODE_ENV === 'test'`.

mrfelton: That would definitely be a good start. This should really apply to both cases regardless of wether the method is passed an `lndDir` path or the parameters to look up a wallet by type.

app/components/Home/WalletLauncher.js
korhaliv: updated

app/preload.js
korhaliv: Security check added

test/unit/utils/isSubDir.spec.js
```diff
@@ -0,0 +1,35 @@
...
+  const tests = [
+    ['/foo', '/foo'],
+    ['/foo', '/bar'],
+    ['/foo', '/foobar'],
+    ['/foo', '/foo/bar'],
+    ['/foo', '/foo/../bar'],
+    ['/foo', '/foo/./bar'],
+    ['/bar/../foo', '/foo/bar'],
+    ['/foo', './bar'],
+    ['C:\\Foo', 'C:\\Foo\\Bar'],
```
mrfelton: These tests will not run as expected on a non-windows machine (test is failing on linux on travis). You probably should make these tests tailored to the value of `os.platform()`

app/components/Home/WalletLauncher.js
```diff
@@ -69,66 +132,128 @@ class WalletLauncher extends React.Component {
...
+
+  saveSettings = () => {
+    const { putWallet, showNotification, showError, intl } = this.props
+    try {
+      const formState = this.formApi.getState()
+      const { values } = formState
+      putWallet(formToWalletFormat(values))
+      const message = intl.formatMessage({ ...messages.saved_notification })
+      showNotification(message)
+    } catch {
```
mrfelton: This is causing failures for the e2e tests. Should be:

```
try {
} catch(e) {
}
```

korhaliv:
e2e fixed with a temp patch as per slack discussion

app/components/Home/WalletLauncher.js
```diff
@@ -69,66 +129,135 @@ class WalletLauncher extends React.Component {
...
+            </>
+          )}
+        </Form>
+      </Box>
     )
   }
 }
 
-export default withRouter(WalletLauncher)
+export default compose(
```
mrfelton: Nice. I never knew about this

korhaliv: Makes it a lot easier to comprehend rather than `hoc1(hoc2(hoc3(cmp)))` imo

PR #1440, opened 2019-01-28
Added TROUBLESHOOTING.md doc

alvistar:
## Description:

Added TROUBLESHOOTING.md with compilation issues

## Motivation and Context:

Provides documentation for common issues

## How Has This Been Tested?

Just documentation

## Types of changes:

Docs

## Checklist:

- [ X] My code follows the code style of this project.
- [ X] I have reviewed and updated the documentation accordingly.
- [X ] I have read the _CONTRIBUTING_ document.
- [ ] I have added tests to cover my changes where needed.
- [ ] All new and existing tests passed.
- [ ] My commits have been squashed into a concise set of changes.

PR #1466, opened 2019-02-02
docs(help): start a troubleshooting guide

mrfelton:
## Description:

Create skeleton for troubleshooting guide and and add a cleaned up #1440

## Motivation and Context:

Ref #1440

## Types of changes:

Docs


korhaliv:
Looks like warning of wallet removal was removed ? Was this edit intentional?

mrfelton:
> Looks like warning of wallet removal was removed ? Was this edit intentional?

I updated the code sample to delete the extensions subdirectory specifically, rather than removing the entire electron data directory which is overkill and unnecessary.

korhaliv:
Overlooked that. Ok that's even better

PR #1500, opened 2019-02-05
feat(channels): new channel list search filters

mrfelton:
## Description:

Add channel search and filter widgets to the new channels list page.

**NOTE: This PR builds on #1491 which should be merged first**

## Motivation and Context:

Ref #1425

## How Has This Been Tested?

Manually

image

## Types of changes:

New features

## Checklist:


mrfelton:
> Is there a reason the channel filter dropdown is orange on hover instead of how we style our other dropdowns? I think we should keep the style consistent.

@JimmyMow the `<Select>` widget was built out a while back, when I was first building out the storybook and creating all of our reusable form widgets. But the Select widget has never actually been used outside of the storybook. I will update the style so that the hover styles more closely match the styles of the dropdown menus used elsewhere in the app.

> Can we make the `x` within the search bar `pointer` on hover?

This is actually added by the browser - it is part of html5's '<input type="search"/> widget - so I don't have any control over this unfortunately.

mrfelton:
Actually, I take that back about the search input `x` icon. Looks like I may be able to control it

mrfelton:
@JimmyMow I've addressed both of those style issues

mrfelton:
image

mrfelton:
> Seems as if the `Funding Transaction` link doesn't link to an explorer. Is that intentional?

That probably belongs in #1539 @JimmyMow - this story only implements the search and filter widgets.

> Also it seems there is a bit of a performance hit when using with my bigger remote nodes. Toggling between `online` and `all` stalls for a few seconds. Anyone else experiencing this?

Yes, there is a performance issue and I have looked into it. The cause is that currently we are rendering both the new and the old versions of the network list (the sidebar is still being rendered even though you can not see it (it's under the modal). The old version is relatively slow on it's own, and with both versions being rendered that his is multiplied.

I have tested and confirmed that by removing the old network list, the performance issue goes away and switching tabs feels very responsive.

However, I don't want to remove the old sidebar until we have at least the same scope of functionality that it provides available in the new version.

JimmyMow:
Fair enough. This LGTM then 🎊

app/components/Channels/Channels.js
```diff
@@ -1,35 +1,64 @@
...
+            channels={allChannels}
+            channelBalance={channelBalance}
+            filter={filter}
+            filters={filters}
+            updateChannelSearchQuery={this.updateChannelSearchQuery}
+            searchQuery={searchQuery}
+            changeFilter={changeFilter}
+          />
+        </Panel.Header>
+        <Panel.Body px={4} css={{ 'overflow-y': 'auto' }}>
```
korhaliv: Should we go with `'overflow-y': 'auto', 'overflow-x':'hidden'` instead as we recently did in master ?

app/components/UI/Input.js
```diff
@@ -42,9 +48,21 @@ function mapFocusBorderColor(props) {
...
+  if (!props.highlightOnValid) {
+    return lightningOrange
+  }
+
+  return fieldState.touched && isFieldValid(fieldState) ? superGreen : lightningOrange
 }
 
+const SearchIcon = styled(Search)`
+  margin-right: -46px;
+  width: 46px;
```
korhaliv: Look like `margin-right` is equal to `width` to do some kind of offset? In that case could be convenient to introduce some kind of a prop (size or smth) `<SearchIcon size={46} />` that will adjust both of this simultaneously

app/components/UI/Input.js
```diff
@@ -167,44 +190,49 @@ class Input extends React.Component {
...
-            const newHasFocus = document.activeElement === this.inputRef.current
-            if (hasFocus !== newHasFocus) {
-              this.setState({ hasFocus: newHasFocus })
-            }
-            if (onFocus) {
-              onFocus(e)
-            }
-          }}
-        />
+        <Flex flexDirection="row" alignItems="center">
```
korhaliv: `flexDirection="row"` is a default so can be safely removed

app/components/UI/Select.js
```diff
@@ -84,19 +86,53 @@ class Select extends React.PureComponent {
...
+      initialSelectedItem,
+      ...rest
+    } = this.props
+    const { setValue, setTouched } = fieldApi
+
+    let initialInputValue
+    if (initialSelectedItem) {
+      initialSelectedItem = items.find(i => i.key === initialSelectedItem)
+      initialInputValue = itemToString(initialSelectedItem)
+    }
```
korhaliv: Could be simplified to
```
const findByKey = ({key}) => key === initialSelectedItem
const initialInputValue = initialSelectedItem && itemToString(items.find(findByKey))
```
to avoid `let`
But if we need to ensure `initialInputValue` equals `undefined` in case  `initialSelectedItem` is `false` `null` or empy string ternary operator should be used or just leave it as it is.

app/components/UI/Select.js
```diff
@@ -123,25 +159,39 @@ class Select extends React.PureComponent {
                   onFocus: openMenu,
                   onMouseDown: toggleMenu
                 })}
+                initialValue={selectedItem ? selectedItem.value : null}
+                forwardedRef={this.inputRef}
+                css={{
+                  cursor: 'pointer',
+                  color: 'transparent',
+                  'text-shadow': `0 0 0 ${color || theme.colors.primaryText}`
```
korhaliv: consider replacing `css` with `styled`

app/components/UI/Select.js
```diff
@@ -123,25 +159,39 @@ class Select extends React.PureComponent {
...
+                }}
               />
             </div>
             <SelectOptionList {...getMenuProps({}, { suppressRefError: true })}>
               {isOpen
                 ? items.map((item, index) => (
```
korhaliv: Using bulky maps in-place tend to make `JSX` harder to read and modify. Consider extracting this map into a local function like `selectOptions`

app/components/UI/Input.js
mrfelton: updated.

app/components/UI/Select.js
mrfelton: In this case, both of these values need to be used independently and conditionally based on each other. I tried your suggestion and it doesn't produce the same results, and I tried several other things but this still seems the most straightforward and readable option to me.

app/components/UI/Select.js
mrfelton: updated.

app/components/UI/Select.js
mrfelton: I've pulled this out into a separate render method.

Issue #1291, opened 2019-01-08
Add contextual help to form fields to help users understand how to fill out forms


mrfelton:
## Detailed Description

Add additional help text to form fields where more information would be useful, such as the fields in the onboarding forms and and the request/pay forms.

## Context

There are many parts of the app where we need to collect data from users. Sometimes it is not always obvious what we are asking for or where the user should get the data or what format they should provide it in.

## Possible Implementation

- Add tooltip icon next to fields where additional help is available
- Show additional help on hover

![image](https://user-images.githubusercontent.com/200251/50822475-1c8ff600-1332-11e9-8fe1-31792dbe35eb.png)

![image](https://user-images.githubusercontent.com/200251/50822506-2d406c00-1332-11e9-983e-34ff0975d16a.png)

bolatovumar:
@mrfelton would a simple first step be to add the "title" attribute to form fields so you get the built-in tooltip when you hover over the input field?

mrfelton:
Closed via https://github.com/LN-Zap/zap-desktop/pull/1737

PR #1501, opened 2019-02-06
feat(ui): add contextual help component

bolatovumar:
Addresses #1291


bolatovumar:
I took a stab at #1291  and added a new component and an example help text message. Let me know if anything should be adjusted.

app/components/UI/TooltipButton.js
```diff
@@ -0,0 +1,86 @@
...
+`
+StyledTooltipWrapper.displayName = 'TooltipWrapper'
+
+/**
+ * @render react
+ * @name TooltipButton
+ * @example
+ * <TooltipButton>Some explanation text here</TooltipButton>
+ */
+class TooltipButton extends React.Component {
```
mrfelton: Let's call this `Tooltip` rather than `TooltipButton` since it encompasses both the button and the tooltip.

app/components/UI/TooltipButton.js
```diff
@@ -0,0 +1,86 @@
...
+          onMouseLeave={this.hoverOff}
+          {...rest}
+        >
+          <Text fontWeight="bold" fontFamily="sans">
+            ?
+          </Text>
+        </StyledTooltipIconBox>
+        {hover ? (
+          <StyledTooltipWrapper bg="secondaryColor">
+            <Text fontWeight="normal" fontFamily="sans" lineHeight="1.4">
```
mrfelton: `fontWeight` should be `light` and I think you can probably remove the `fontFamily` declaration.

app/components/UI/TooltipButton.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Box, Text } from 'rebass'
+
+const StyledTooltipIconBox = styled(Box)`
+  border-radius: 50%;
+  height: 16px;
```
mrfelton: I think this should be slightly smaller. In the mockup it is 13px.

app/components/Request/Request.js
```diff
@@ -130,11 +140,12 @@ class Request extends React.Component {
     const { intl } = this.props
     return (
       <Box>
-        <Box pb={2}>
+        <Flex pb={2}>
           <Label htmlFor="memo">
             <FormattedMessage {...messages.memo} />
           </Label>
-        </Box>
+          <TooltipButton>{intl.formatMessage({ ...messages.memo_help })}</TooltipButton>
```
mrfelton: Can I suggest that we build this functionality into the `Input` component, so that you can use it like:

```js
<Input tooltip={intl.formatMessage({ ...messages.memo_help })}/>
```

Note that the `Input` element already supports a `label` property which can be used to add the `<label>` element automatically. I'm not sure why that isn't being used for this input but we should probably update this so that it does.

app/components/Request/messages.js
```diff
@@ -11,6 +11,8 @@ export default defineMessages({
   payment_request: 'Payment Request',
   total: 'Total',
   memo: 'Memo',
+  memo_help:
```
mrfelton: When adding new strings you should run `yarn extract-messages` so that they get extracted into our translation files at `/app/translations`.

Also, if we are calling the element a `tooltip` then I think the translation key would be better named as `memo_tooltip` for consistency.

app/components/UI/TooltipButton.js
```diff
@@ -0,0 +1,86 @@
...
+  hoverOn = () => {
+    this.setState({ hover: true })
+  }
+
+  hoverOff = () => {
+    this.setState({ hover: false })
+  }
+
+  render() {
+    const { children, ...rest } = this.props
```
korhaliv: Is there a reason why props are destructured onto inner container rather than wrapping `Box` ?

app/components/UI/TooltipButton.js
```diff
@@ -0,0 +1,86 @@
...
+    const { hover } = this.state
+
+    return (
+      <Box
+        style={{
+          position: 'relative'
+        }}
+      >
+        <StyledTooltipIconBox
+          bg="white"
```
korhaliv: Should be `primaryText` to keep it compatible with other themes

app/components/UI/TooltipButton.js
korhaliv: `lineHeight="1.4"` why use custom line height here ? We have a clear style guidelines which are achieved with a set of defaults in most of the components. Please see `npm run storybook` for the reference.  One shouldn't use custom styles unless absolutely necessary

bolatovumar: The default line height seemed too small and didn't look like the mock-up.  I will adjust it to match the style guide.

app/components/UI/TooltipButton.js
bolatovumar: Will do.

app/components/UI/TooltipButton.js
bolatovumar: No, that's my bad. It should go to the outer box.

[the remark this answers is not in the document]
bolatovumar: I'm not sure what you mean by this. Do you mean propTypes?

app/components/UI/TooltipButton.js
```diff
@@ -0,0 +1,86 @@
...
+          color="primaryColor"
+          onMouseEnter={this.hoverOn}
+          onMouseLeave={this.hoverOff}
+          {...rest}
+        >
+          <Text fontWeight="bold" fontFamily="sans">
+            ?
+          </Text>
+        </StyledTooltipIconBox>
+        {hover ? (
```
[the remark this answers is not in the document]
bolatovumar: Sure thing.

bolatovumar:
@korhaliv @mrfelton how would you prefer the changes made? Squash them into one commit and push the branch? If I do that then the context around your comments will get lost, right (since the code will be different)? Does this matter?

app/components/UI/TooltipButton.js
bolatovumar: I made it 13px at first but it looked small to me. I will change it to 13px.

app/components/UI/TooltipButton.js
bolatovumar: I'm looking at the style guide and can't find anything about line height there. Also searched the source for `lineHeight` and `line-height` to see if there are any line height global variables defined but didn't find anything. Am I missing something here?

app/components/Request/Request.js
bolatovumar: I will do that.

In `Request.js` the memo field is actually a textarea hence why label component is used instead of the label prop. `<Input />` component can't render a textarea, right?

app/components/Request/messages.js
bolatovumar: Will do.

app/components/UI/TooltipButton.js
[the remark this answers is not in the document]
korhaliv: Yes, `propTypes`.

korhaliv:
@bolatovumar Squashing is preferable. This PR is rather small so there won't be much of an issue to re-review once you update.

app/components/UI/TooltipButton.js
korhaliv: What I meant to say is that `Text` component already has a proper line height, so usually no need to set it manually

bolatovumar: Ok, I will just remove it then.

bolatovumar:
Implemented your suggestions and also added a story component showing a label with tooltip and and an input.

bolatovumar:
I just noticed that a couple of other messages have been added alongside my new memo tooltip message after I ran `yarn extract-messages`. I rebased on top of the latest `next` branch so I'm not sure why these messages weren't there in the first place.

mrfelton:
Hi @bolatovumar .

Regarding the extra messages that were being extracted into the language files when you run `yarn extract-messages` - this was because there were new strings in `next` that were no properly extracted. I have fixed this in the next branch so please rebase from the next HEAD and re run the message extract script.

app/components/Request/Request.js
mrfelton: The Pay/Request forms we built before the `label` property was added to the `Input` and `TextArea` fields and these forms had not yet been updated to use the new `label` property.

We have now updated the code in `next` so that these these forms both now properly use the `label` property - which is preferable as it ensures consistent styling of the labels elements.

So - we will need to add the tooltip support directly to these Input/TextArea elements and use via those, so that they can be applied along with the label in a standard and consistent way throughout the app's forms.

app/components/UI/Tooltip.js
```diff
@@ -0,0 +1,89 @@
...
+
+const StyledTooltipWrapper = styled(Box)`
+  border: 1px solid white;
+  border-radius: 4px;
+  padding: 5px 5px 5px 25px;
+  position: absolute;
+  top: -5px;
+  width: 225px;
+  z-index: 1;
+  transform: translateZ(0);
```
korhaliv: Is this related to https://stackoverflow.com/questions/10814178/css-performance-relative-to-translatez0 ?
If so then it seems like it's not a recommended approach

app/components/UI/Tooltip.js
```diff
@@ -0,0 +1,89 @@
...
+        }}
+        {...rest}
+      >
+        <StyledTooltipIconBox
+          bg="primaryText"
+          color="primaryColor"
+          onMouseEnter={this.hoverOn}
+          onMouseLeave={this.hoverOff}
+        >
+          <Text fontWeight="bold" fontFamily="sans">
```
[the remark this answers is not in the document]
bolatovumar: No. Probably just saw this being used elsewhere and copy-pasted it. Will remove if not necessary.

app/components/UI/Tooltip.js
bolatovumar: No, this was to put it on its own rendering layer so that when I use z-index on the icon and the surrounding layer they are relative to each other. Without this trick it doesn't seem to work.

bolatovumar: Actually, just checked and it seems to work fine without the `translateZ`. Removed.

bolatovumar:
Sorry for the delay.

I rebased on top of the latest in the `next` branch and when I run the `extract-messages` command I still get several message properties that I'm not expecting to see alongside the new one that I added. These other messages are:

- `"components.Activity.TransactionModal.maturity_height"`
- `"components.Channels.closing_transaction_id_description"`
- `"components.Channels.closing_transaction_id_label"`

I'm guessing someone else need to run the `extract-messages` command first and merge-in the resulting change before I can push mine?

mrfelton:
@bolatovumar you are right about the translation strings. I have updated `next` to ensure that all new transaction strings are included. Please rebase again! Thanks.

app/components/UI/Tooltip.js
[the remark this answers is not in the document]
bolatovumar: Removed.

bolatovumar:
Ok, looking better now.

mrfelton:
@bolatovumar I rebased this PR and made a couple of tweaks to bring the style slightly more inline with the rest of the styles, and reuse our theme style props where possible.

https://github.com/LN-Zap/zap-desktop/pull/1737

This is now merged, Thanks!

PR #1539, opened 2019-02-08
feat(channels): channel card list and view switcher

### LINKED ISSUE #1490: Implement new channel (card) list view
Implement the newly designed channel management (summary) list view.

Design can be found here: 
https://app.zeplin.io/project/5bacdbad33af886d419f1b34/screen/5c41f84f50e77937afe1181d

mrfelton:
## Description:

Add channel search and filter widgets to the new channels list page.

**NOTE: This PR builds on #1500 which should be merged first**

## Motivation and Context:

Resolves #1490

## How Has This Been Tested?

Manually

image

## Types of changes:

New features

## Checklist:


mrfelton:
> In my opinion the card view should be the default. Thoughts on this? That's a change I'd make personally.

The main consideration I make when deciding which view mode to use is based on the number of channels. IF I'm looking at a node with only a couple of channel the card view makes sense. If I'm looking at one with more channels, the list view makes more sense.

So - if as I suspect - we think that most people using Zap will not be users running routing nodes then the card view is probably a good option.

Actually, after using this UI with a node with 30 channels, I'd like to see another even more compressed view. Like the list, but with each row only being a single line. "Table view" or something like that. So that you can get a bigger snapshot of your overall current channel state on a single page. With the old sidebar I could see most of my channels together on the page and I miss that with these new views.

JimmyMow:
@mrfelton yep. instead of like a table view, i'm more interested in designing visual graphs and other tools to help digest your channel state. different charts, graphs, etc would be pretty cool and a great way to take in all that data

mrfelton:
@JimmyMow  I've updated the default view mode to use the card view.

> i'm interested in designing visual graphs and other tools to help digest your channel state. different charts, graphs, etc would be pretty cool and a great way to take in all that data

> I'd like to have a filter where I can view my channels that are heavy (more than half) on the remote side, and more than half on the local side.

Maybe you could create some stories for these additional features?

app/components/Channels/ChannelCardListItem.js
```diff
@@ -0,0 +1,114 @@
...
+      <Box as="header">
+        <Flex justifyContent="space-between">
+          <ClippedHeading my={1} opacity={opacity}>
+            {channelName}
+          </ClippedHeading>
+
+          <ChannelStatus status={status} mb="auto" />
+        </Flex>
+        <Flex justifyContent="space-between" opacity={opacity}>
+          <Text fontWeight="normal">Remote PubKey</Text>
```
korhaliv: Looks like `Remote PubKey` should be made intl compatible

app/components/Channels/ChannelCardListItem.js
```diff
@@ -0,0 +1,114 @@
...
+        localBalance={localBalance}
+        remoteBalance={remoteBalance}
+        opacity={opacity}
+        my={4}
+      />
+
+      <Box as="section" opacity={opacity}>
+        <Bar opacity={0.3} />
+        <DataRow
+          left="Funding Transaction"
```
korhaliv: Looks like this needs intl support

app/components/Channels/ChannelCardListItem.js
```diff
@@ -0,0 +1,114 @@
...
+                networkInfo && blockExplorer.showTransaction(networkInfo, channelFundingTxid)
+              }
+            >
+              <Truncate text={channelFundingTxid} maxlen={25} />
+            </Link>
+          }
+          py={2}
+        />
+        <Bar opacity={0.3} />
+        <DataRow left="Number of Updates" right={numUpdates} py={2} />
```
korhaliv: Looks like this needs intl support

app/components/Channels/ChannelCardListItem.js
```diff
@@ -0,0 +1,114 @@
...
+        <Bar opacity={0.3} />
+        <DataRow left="Block Time Lock" right={csvDelay} py={2} />
```
korhaliv: Looks like this needs intl support

app/components/Channels/ChannelCardListItem.js
```diff
@@ -0,0 +1,114 @@
...
+      </Box>
+
+      <Flex as="footer" justifyContent="center" height="40px">
+        {channelId && (
+          <ChannelMoreButton onClick={() => showChannelDetail(channelId)} mx="auto" my="auto" />
```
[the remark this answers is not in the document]
mrfelton: Yes - `alignItems="flex-end"` did it. Updated.

app/components/Channels/ChannelCardList.js
```diff
@@ -0,0 +1,54 @@
...
+        </Box>
+      )
+    })}
+  </Flex>
+)
+
+ChannelCardList.propTypes = {
+  channels: PropTypes.array,
+  showChannelDetail: PropTypes.func.isRequired,
+  networkInfo: PropTypes.shape({
```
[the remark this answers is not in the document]
mrfelton: It shouldn't have a default, since we don't make any assumptions about what network the user is connecting to. We figure this out by looking at the return data from `getInfo` lnd api call.

Higher up the chain, we prevent rendering the app until we know this information.

app/components/Channels/ChannelCardListItem.js
```diff
@@ -0,0 +1,114 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Box as BaseBox, Flex as BaseFlex } from 'rebass'
+import { Bar, Card, DataRow, Heading, Link, Text } from 'components/UI'
+import { ChannelCapacity, ChannelMoreButton, ChannelStatus } from 'components/Channels'
+import { blockExplorer } from 'lib/utils'
+import { Truncate } from 'components/Util'
+import withEllipsis from 'components/withEllipsis'
+import styled from 'styled-components'
+import { opacity, height } from 'styled-system'
```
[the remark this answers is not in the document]
mrfelton: Updated.

app/components/Channels/ChannelCardListItem.js
mrfelton: Completely forgot to go back and extract these. Updated, thanks!

app/components/Channels/ChannelCardListItem.js
mrfelton: updated.

app/components/Channels/ChannelCardListItem.js
mrfelton: updated.

app/components/Channels/ChannelCardListItem.js
mrfelton: updated.

PR #1551, opened 2019-02-10
fix(forms): more robust handling for amount inputs

### LINKED ISSUE #1547: Pay form goes blank when entering period in input
When trying to make an on-chain payment, if you enter a `.`, with the purpose of sending an amount less than `1`, like `.5` Zap goes blank.

Expected behavior is you are allowed to type `.5` into the payform input

mrfelton:
## Description:

- Improve input mask masking for crypto/fiat amount input fields when a period character is entered
- Extract duplicate code from `FiatAmountInput` and `CryptoAmountInput` components into a reusable HOC.

## Motivation and Context:

Fix #1547 in which entering `.` into the crypto amount input would cause the app to crash.

## How Has This Been Tested?

Manually

## Types of changes:
Bug fix

## Checklist:

### DIFF
--- a/app/components/UI/CryptoAmountInput.js
+++ b/app/components/UI/CryptoAmountInput.js
@@ -7,6 +7,7 @@ import * as yup from 'yup'
 import { convert } from 'lib/utils/btc'
 import { formatValue, parseNumber } from 'lib/utils/crypto'
 import Input from 'components/UI/Input'
+import withNumberInputMask from 'components/withNumberInputMask'
 
 /**
  * @render react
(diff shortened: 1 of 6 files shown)


app/components/UI/CryptoAmountInput.js
```diff
@@ -152,4 +124,4 @@ class WrappedCryptoAmountInputAsField extends React.Component {
   }
 }
 
-export default WrappedCryptoAmountInputAsField
+export default asField(withNumberInputMask(WrappedCryptoAmountInputAsField))
```
korhaliv: Could haved used `compose` from `redux`, but that's just a matter of preference really

Issue #1178, opened 2018-12-24
More helpful loading screen


JimmyMow:
Sometimes the app gives the user our animated loading screen for 20 seconds or so. Sometimes as a user I am confused whether something went wrong, and am often left checking logs, to see the app is acting totally fine and processing some data.

It would be nice if the Loading component was able to take some strings as an argument to be displayed. So for example the loading screen while the app is trying to connect to the last remote node used on start the Loading component can be streaming updates like:

`Loading - Connecting to custom lnd instance`
`Loading - Establishing connection to Lightning gRPC interface...`
`Loading - Connecting to Lightning gRPC service`

I think this would be much more helpful and improve the UX

mrfelton:
Closed via https://github.com/LN-Zap/zap-desktop/pull/1605

PR #1561, opened 2019-02-12
fix(ui): remote node loading screen

korhaliv:
## Description:
When trying to connect to a remote node that you previously had successfully connected to and that is currently offline, there is a blank screen for a long period of time (~18 secs) until you see a timeout error. This is caused by bolt loading screen being prematurely unmounted and creates an impression that app is stuck.

This is also a prerequisite for #1178

This PR also does an overhaul of initial route determination mechanism to make it more declarative and straightforward.

## How Has This Been Tested?
Manually

Tested:

1. Remote node that is unlocked
2. Remote node that is locked
3. Remote node that was successfully connected to and now is offline
4. Local node
5. Lndconnect link

## Types of changes:
Bugfix/refactor

## Checklist:

### DIFF
--- a/app/containers/Initializer.js
+++ b/app/containers/Initializer.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import { withRouter } from 'react-router'
+import { Redirect } from 'react-router'
 import { appSelectors } from 'reducers/app'
 import { startActiveWallet } from 'reducers/lnd'
 import { initCurrency, initLocale } from 'reducers/locale'
@@ -14,17 +14,12 @@ import { fetchSuggestedNodes } from 'reducers/channels'
  */
 class Initializer extends React.Component {
   static propTypes = {
-    history: PropTypes.object.isRequired,
     activeWallet: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     activeWalletSettings: PropTypes.object,
     hasWallets: PropTypes.bool.isRequired,
     isWalletOpen: PropTypes.bool.isRequired,
     isReady: PropTypes.bool.isRequired,
-    lightningGrpcActive: PropTypes.bool.isRequired,
     lndConnect: PropTypes.string,
-    walletUnlockerGrpcActive: PropTypes.bool.isRequired,
-    startLndError: PropTypes.object,
-    startActiveWallet: PropTypes.func.isRequired,
     fetchSuggestedNodes: PropTypes.func.isRequired,
     fetchTicker: PropTypes.func.isRequired,
     initLocale: PropTypes.func.isRequired,
@@ -32,10 +27,8 @@ class Initializer extends React.Component {
     initWallets: PropTypes.func.isRequired
   }
 
-  nextLocation = null
-
   /**
-   * Initialise app state.
+   * Initialize app state.
    */
   componentDidMount() {
     const { fetchSuggestedNodes, fetchTicker, initLocale, initCurrency, initWallets } = this.props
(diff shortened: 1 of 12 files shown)


app/containers/Initializer.js
```diff
@@ -48,74 +41,42 @@ class Initializer extends React.Component {
...
-    }
+    // If we have at least one wallet send the user to the homepage.
+    // Otherwise send them to the onboarding processes.
+    return hasWallets ? '/home' : '/onboarding'
   }
 
   render() {
-    return null
+    const location = this.getLocation()
+    return location && <Redirect to={location} />
```
mrfelton: I like the way you are doing this in the render function now. This simplifies things a lot and should also mean that it covers the case where the next location is already known when the component first mounts (previously, we only checked in the component update hook)

app/containers/Root.js
```diff
@@ -40,39 +38,19 @@ class Root extends React.Component {
...
-    // If this is the first time the app has mounted, initialise things.
+    // If this is the first time the app has mounted, initialize things.
     if (!isMounted) {
       setMounted(true)
       initTheme()
     }
-
-    // Hide the loading screen after a set time.
-    if (isLoading || !theme) {
-      const timer = setTimeout(() => setLoading(false), SPLASH_SCREEN_TIME)
```
mrfelton: Without this, when you start up the app that has no active/open wallet the loading graphic flickers onscreen for a split second before the user is shown the home screen.

I think we should try to keep the minimum splash screen visibility time if possible.

Ideally, the loading graphic should never show for less than a second, as it doesn't look good when it flickers on and off incredibly quickly.

app/containers/WalletStarter.js
```diff
@@ -0,0 +1,79 @@
...
+import PropTypes from 'prop-types'
+import { startActiveWallet } from 'reducers/lnd'
+import { Redirect } from 'react-router'
+import { walletSelectors } from 'reducers/wallet'
+
+/**
+ * Auxiliary router path that starts wallet or gRPC
+ * connection and redirects to a relevant path when complete
+ */
+class WalletStarter extends Component {
```
mrfelton: +1 for pulling this out into a separate class

app/reducers/lnd.js
```diff
@@ -87,11 +88,18 @@ export const lndSyncStatus = (event, status) => async (dispatch, getState) => {
   }
 }
 
+export const startLightningGrpc = () => dispatch => dispatch(setLoading(true))
+export const startLightningGrpcFinished = () => dispatch => dispatch(setLoading(false))
+
+export const startWalletUnlockerGrpc = () => dispatch => dispatch(setLoading(true))
```
mrfelton: Triggering the loading screen on and off like this causes a bit of a flicker effect.

For example:

On the home screen when clicking the "launch: button to launch a local wallet, the loading screen flickers on and off in the split second between when you click the button, and when the password field shows, whilst the wallet unlocker service starts up.

I don't think we really want to trigger the loading screen to show here. Better that the loading button just goes to the processing state until we are ready to show the password entry form.

korhaliv:
Done some updates as per the slack discussion

app/reducers/lnd.js
```diff
@@ -90,8 +90,9 @@ export const lndSyncStatus = (event, status) => async (dispatch, getState) => {
 // Connected to Lightning gRPC interface (lnd wallet is connected and unlocked)
 export const lightningGrpcActive = (event, lndConfig) => async dispatch => {
   dispatch({ type: SET_LIGHTNING_WALLET_ACTIVE })
-
-  // Once we we have established a connection, save the wallet settings.
+  // Once we we have established a connection, save the wallet settings
+  // after connection was successfully established. This is especially important
+  // for the first connection to be sure settings are correct  if (lndConfig.id !== 'tmp') {
```
mrfelton: I think you can remove `if (lndConfig.id !== 'tmp') {` from this comment. It doesn't seem relevant. That condition just ensures that we are excluding the special `tmp` wallet (the one that we use to generate new seeds)

korhaliv: Right. That's probably a merge artifact.

app/containers/Root.js
```diff
@@ -130,7 +110,7 @@ const mapStateToProps = state => ({
   hasWallets: walletSelectors.hasWallets(state),
   notifications: notificationSelectors.getNotificationState(state),
   theme: themeSelectors.currentThemeSettings(state),
-  isLoading: appSelectors.isLoading(state),
+  isLoading: appSelectors.isLoading(state) || lndSelectors.isStartingLnd(state),
```
mrfelton: I think here you may need to make an exclusion for if `onboarding` is true - since we don't want the loading screen to flicker on/off when starting lnd in the onboarding process (this happens clicking "launch" to start a local wallet, and also when generating a new seed in the "Create new wallet" onboarding process.

mrfelton: Actually, this probably won't work, because the state prop I was referring to (`state.onboarding.onboarding`) gets reset every time you restart the onboaridng process (ie, every time you enter the Onboarding Wizard).

Perhaps what we need is a new reducer for the home area so that we can track when it's in use, and then exclude triggering the loader based on this. Or, maybe exclude based on the current location? (though this feels a little off somehow).

Something like:

```js
  isLoading:
    (appSelectors.isLoading(state) || lndSelectors.isStartingLnd(state)) &&
    !onboardingSelectors.isOnboarding(state) &&
    !homeSelectors.isActive(state),
```

korhaliv: I think I've found a way. Check the update

mrfelton: Looks good @korhaliv

PR #1568, opened 2019-02-13
feat(channels): new create channel form

### LINKED ISSUE #1448: New create channel form UI
Implement the new channel form UI.

Screen designs can be found here:

https://app.zeplin.io/project/5bacdbad33af886d419f1b34/screen/5c519c6d1dfb92353b503ef9
https://app.zeplin.io/project/5bacdbad33af886d419f1b34/screen/5c519c6e0ddfba308d08c95e

mrfelton:
## Description:

Implement **basic version** of new create channel form.

## Motivation and Context:

This provides a base that we can use to start building out the additional bells and whistles of the new create channel flow.

Resolves #1448

## How Has This Been Tested?

Manually

image

## Types of changes:

New Features

## Checklist:


mrfelton:
@JimmyMow can you supply copy to replace the lorum ipsum plz.

app/components/Channels/ChannelCreateForm.js
```diff
@@ -0,0 +1,135 @@
...
+            <Panel.Footer>
+              <Flex alignItems="center">
+                <Box width={1 / 4}>
+                  {onCancel && <ChannelBackButton onClick={onCancel} mr="auto" />}
+                </Box>
+                <Flex width={2 / 4}>
+                  <Button
+                    type="submit"
+                    my={3}
+                    mx="auto"
```
korhaliv: Can the same effect be achieved with flex props?

app/components/Form/Form.js
```diff
@@ -4,6 +4,7 @@ import { Modal } from 'components/UI'
 import Pay from 'containers/Pay'
 import Request from 'containers/Request'
 import Channels from 'containers/Channels'
+import ChannelCreate from 'containers/Channels/ChannelCreate'
```
korhaliv: This is probably outside of the scope of this PR, but `Form` keeps extending with new containers and is effectively a container itself.

app/containers/Channels/ChannelCreate.js
```diff
@@ -0,0 +1,32 @@
...
+    setFormType: PropTypes.func.isRequired
+  }
+
+  render() {
+    const { setFormType, ...rest } = this.props
+
+    return (
+      <ChannelCreateForm
+        onCancel={() => setFormType('CHANNELS')}
+        onSubmit={() => setFormType('CHANNELS')}
```
korhaliv: this can be extracted and bound in constructor to avoid redundant re-renders
`this.setFormType = this.setFormType.bind(this, 'CHANNELS')`

app/components/Form/Form.js
mrfelton: Yes, this is very true but I don't want to address here. This so called `Form` component could probably be more like a `Modal` container that is used for the display of all modals. The name `Form` is misleading as the use case for this container is not really about displaying forms, but more about displaying stuff on top of the main app.

app/containers/Channels/ChannelCreate.js
mrfelton: Good call. Updated.

app/components/Channels/ChannelCreateForm.js
mrfelton: Yes. Updated!

mrfelton:
Thanks @korhaliv . I've updated this PR to address your comments.

mrfelton:
> Move funds onto the Lightning Network by ...

Or

> Make funds available for sending over the Lightning Network by ...

mrfelton:
@JimmyMow copy updated. As per slack convo, went with second option.

PR #1576, opened 2019-02-14
feat(ui): add close channel dialog

### LINKED ISSUE #1449: Implement force close modal warning
When a user force closes a channel, they should first be prompted to verify, as their funds will be time locked.

Use the existing warning modal within the code base that we use for deleting wallets, but just changed copy.

Design can be found here: https://app.zeplin.io/project/5bacdbad33af886d419f1b34/screen/5c3c88e822be21beb3e8c99e

korhaliv:
## Description:
Implement channel close confirmation dialog. 

## Motivation and Context:
fix #1449 

## How Has This Been Tested?
Manually

image

## Types of changes:
New feature

## Checklist:


app/components/Channels/CloseChannelDialog.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
```
mrfelton: Could we call this `ChannelCloseDialog` rather than `CloseChannelDialog` since all of the other related Channel components are prefixed `Channel`

app/components/Channels/CloseChannelDialog.js
```diff
@@ -0,0 +1,72 @@
...
+import messages from './messages'
+import Overlay from '../DialogOverlay'
+
+const DialogWrapper = ({ intl, isOpen, onClose, onCancel }) => {
+  if (!isOpen) {
+    return null
+  }
+  const checkboxFieldName = 'actionACK'
+
+  // bind button disabled state to a form field or ignore it if we are dealing with a remote
```
mrfelton: Reference to remote wallet can be removed from comment.

app/components/Channels/messages.js
```diff
@@ -43,5 +43,10 @@ export default defineMessages({
     'The amount to be paid in fees for the current set of commitment transactions.',
   base_fee_label: "Node's Fee Base",
   base_fee_description:
-    'The required number of {currencyName} per kilo-weight that the requester will pay at all times, for both the funding transaction and commitment transaction.'
+    'The required number of {currencyName} per kilo-weight that the requester will pay at all times, for both the funding transaction and commitment transaction.',
```
mrfelton: New translation strings need to be extracted with `yarn extract-messages`

app/components/DialogOverlay.js
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components'
```
mrfelton: Lets either place this into `app/components/Util` or `app/components/UI` so that we don't have dangling files in `app./components`

app/containers/Channels/CloseChannelDialog.js
```diff
@@ -0,0 +1,27 @@
...
+const mapStateToProps = state => {
+  const isOpen = Boolean(state.channels.isCloseDialogOpen)
+  return {
+    isOpen
+  }
+}
+
+const mapDispatchToProps = dispatch => ({
+  onClose() {
+    alert('close')
```
mrfelton: Do we really want to `alert` here?

korhaliv:
Thanks for the comments but this is still wip as per the title. So some things like messages are not yet done purposely

korhaliv:
Previously you said that you want to see that there is some work being done on this and i should push whatever i have.

app/containers/Channels/CloseChannelDialog.js
korhaliv: This is wip

app/components/Channels/messages.js
korhaliv: This is wip

app/components/Channels/CloseChannelDialog.js
korhaliv: This is still wip

app/components/DialogOverlay.js
korhaliv: done

app/components/Channels/CloseChannelDialog.js
korhaliv: done

app/components/Channels/CloseChannelDialog.js
korhaliv: removed

mrfelton:
> Thoughts on this copy for a force close?
Looks ok to me @JimmyMow . 

@korhaliv I think that would be the `csv_delay` channel property that needs inserting into the copy.

korhaliv:
Last revision includes the following updates:
1. Theme compatibility (visual glitches)
2. Updated copy
3. Intl support
4. Ack checkbox only in force close mode
5. Some refactor as requested by @mrfelton  (hook has been extracted and applied to Delete wallet dialog)

app/components/Channels/ChannelCloseDialog.js
```diff
@@ -0,0 +1,86 @@
...
+      </CloseChannelButton>
+      <Button variant="secondary" type="button" onClick={onCancel}>
+        <FormattedMessage {...messages.close_channel_dialog_cancel_text} />
+      </Button>
+    </>
+  )
+
+  const header = (
+    <Flex flexDirection="column" alignItems="center" alignSelf="center">
+      <Delete color="#e8383a" width={72} height={72} />
```
mrfelton: What is this hardcoded colour?

korhaliv: Strange stuff. Probably inherited from delete dialog. So has to be updated there too

korhaliv: updated

Issue #1559, opened 2019-02-11
Adjustments to lint rules


mrfelton:
## Detailed Description

Adjust lint rules to automate more things. Things that have been discussed include:

- Enable `react/jsx-sort-props` rule
- Enable `react/sort-prop-types` rule
- Enable `react/boolean-prop-naming` rule
- Enable `comma-dangle` rule
- Enable `no-negated-condition` rule
- Enable a rule that enforces import order such as `sort-imports` or `eslint-plugin-import-order`

## Context

Save time and rework by automating and enforcing things that we like to be consistent about.

mrfelton:
Closed via https://github.com/LN-Zap/zap-desktop/pull/1736

PR #1579, opened 2019-02-16
chore(devops): enable comma dangle rule

bolatovumar:
Address part of #1559

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -67,6 +67,7 @@ module.exports = {
     'no-mixed-operators': 'error',
     'no-new': 0,
     'no-tabs': 'error',
+    'comma-dangle': ['error', 'never'],
     'compat/compat': 'error',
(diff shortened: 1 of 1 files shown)


mrfelton:
Thanks for the patch @bolatovumar . However, I'm going to close this issues because https://github.com/LN-Zap/zap-desktop/issues/1559 requires some special handling and timing.

Firstly, the intent was to enable this specific rule, rather than disable it as done in this patch.

Secondly, after changing our lint rules we will need to re-lint and fix/adjust our entire codebase to confirm to the new rules. This will be quite a disruptive action and will likely cause conflicts across many PR's.

So, with that in mind, our plan is to make all of the changes to the lint rules listed in #1559 in one go, and re-lint and adjust the codebase in a single PR, which will be closely timed with our next release in order to reduce the amount of potential conflicts and disruption.

PR #1593, opened 2019-02-17
refactor(channels): remove old network sidebar

### LINKED ISSUE #1426: Remove current channel UI (Right side bar)
Once the new channel UI is implemented, the old channel UI needs to be removed from the UI and any old (now unused) code should be removed

mrfelton:
## Description:

Remove the old network sidebar from the UI. This PR does not fully remove all the old network related code as we have not yet finished reimplementing everything but instead focuses on removing the network list from the UI.

Will follow with a PR to remove all old/unused code at a later point.

Ideally, we should merge #1568 prior to this one since we need that in place to ensure that we can still do all basic channel management through the new UI.

## Motivation and Context:

Resolves #1426

## How Has This Been Tested?

Manually

image

## Types of changes:

Refactor

## Checklist:

### DIFF
--- a/app/components/App/App.js
+++ b/app/components/App/App.js
@@ -1,14 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import Form from 'components/Form'
-import ChannelForm from 'components/Contacts/ChannelForm'
-import Network from 'components/Contacts/Network'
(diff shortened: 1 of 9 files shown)


mrfelton:
> I'm wondering if we should decrease the default width to `900px`

Yes, I think this works. Have added another commit to this PR to adjust default width.

PR #1599, opened 2019-02-18
refactor(ui): multi-level modals

mrfelton:
## Description:

Support multiple modals in a stack.

## Motivation and Context:

Make for a better UX by allowing for multiple modals to be stacked on top of each other.

## How Has This Been Tested?

Try all modals.

## Types of changes:

UX improvement

## Checklist:


app/reducers/modal.js
```diff
@@ -0,0 +1,81 @@
...
+  [CLOSE_MODAL]: (state, { id }) => {
+    if (id) {
+      return {
+        ...state,
+        modals: state.modals.filter(item => item.id !== id)
+      }
+    }
+    return {
+      ...state,
+      modals: state.modals.filter((item, index) => index !== state.modals.length - 1)
```
korhaliv: This looks like comparing an index to a constant (last modal index). Can we avoid using filter here and just use that last element ?

app/reducers/modal.js
```diff
@@ -0,0 +1,81 @@
...
+// ------------------------------------
+// Constants
+// ------------------------------------
+export const OPEN_MODAL = 'OPEN_MODAL'
+export const CLOSE_MODAL = 'CLOSE_MODAL'
+
+/**
+ * Generates uniq id for modals
+ */
+const genId = () =>
```
korhaliv: There is a similar code in notifications. Could you please extract that into an util and reuse ?

app/containers/Root.js
```diff
@@ -81,18 +83,18 @@ class Root extends React.Component {
...
-                    </Modal>
+                    </ModalOverlay>
                   )}
                 />
                 <Route
                   exact
                   path="/syncing"
                   render={() => (
-                    <Modal withHeader onClose={() => history.push('/logout')} pb={0} px={0}>
+                    <ModalOverlay withHeader onClose={() => history.push('/logout')} pb={0} px={0}>
```
korhaliv: Is it necessary to use `pb={0} px={0}` ?  I suppose we are overriding some defaults here ?

app/containers/Root.js
```diff
@@ -81,18 +83,18 @@ class Root extends React.Component {
                   exact
                   path="/onboarding"
                   render={() => (
-                    <Modal withClose={hasWallets} onClose={() => history.push('/home')}>
+                    <ModalOverlay withClose={hasWallets} onClose={() => history.push('/home')}>
```
korhaliv: Currently modals are re-rendered because of this `onClose`, even if nothing else has changed. Could you extract this ?

app/containers/Modal.js
```diff
@@ -0,0 +1,67 @@
...
+
+ModalContent.propTypes = {
+  type: PropTypes.string.isRequired,
+  closeModal: PropTypes.func.isRequired
+}
+
+const Modal = ({ modals, closeModal }) =>
+  modals.map(modal => {
+    return (
+      <ModalOverlay onClose={() => closeModal(modal.id)} key={modal.type}>
```
korhaliv: Can we use `modal.id` as a key? Currently there is nothing in your code that prevents creation of modals with the same type within the stack other than this key. It may not be useful or desirable, but for code consistency sake I think it's the right thing to do.

app/containers/Modal.js
```diff
@@ -0,0 +1,67 @@
...
+      return <Pay width={9 / 16} mx="auto" />
+
+    case 'REQUEST_FORM':
+      return <Request width={9 / 16} mx="auto" />
+
+    case 'CHANNELS':
+      return <Channels mx={-4} />
+
+    case 'CHANNEL_CREATE_FORM':
+      return <ChannelCreateForm onSubmit={() => closeModal()} width={9 / 16} mx="auto" />
```
korhaliv: All modals except for the last one have `mx="auto"` or them. I wonder what's the reason for that

app/components/UI/Modal.js
```diff
@@ -60,7 +60,7 @@ class Modal extends React.Component {
         bg="primaryColor"
         color="primaryText"
       >
-        <Panel.Header p={3}>
+        <Panel.Header pt={3} px={3}>
           <Flex justifyContent="space-between">
             <Box
               css={{ height: '40px', cursor: 'pointer', opacity: 0.6, '&:hover': { opacity: 1 } }}
```
korhaliv: Since you are doing an overhaul, could you please extract this into a styled component please?

app/reducers/modal.js
mrfelton: Updated.

app/reducers/modal.js
mrfelton: Good call. Done.

app/containers/Root.js
mrfelton: Exactly, the modal has sensible defaults. in this case though, we want the content to stretch right the to bottom and sides.

app/containers/Root.js
mrfelton: Updated.

app/containers/Modal.js
mrfelton: Yep, done. I think I did this before I introduced the unique id.

app/containers/Modal.js
mrfelton: This is a bit of a strange case. For the, `CHANNEL_DETAIL` modal, the `ChannelDetail` component itself has the margin already applied because we want the content centered in the middle, but we want it's scrollbars to appear right at the edge of the modal rather than next to the centered content.

app/components/UI/Modal.js
mrfelton: I've cleaned up the Modal component a little.

mrfelton:
Made these updates @korhaliv - ready for re-review

korhaliv:
@mrfelton This is starting to look not bad at all. But there is an issue. If you hit `escape` it still closes all levels of the popups

mrfelton:
Yeah, re the escape issue - I'd like to address in a separate PR since there are already issues with that outside of this PR

PR #1611, opened 2019-02-20
feat(channels): new channel search form

### LINKED ISSUE #1571: I should see a summary of my final channel configuration before I submit the new channel form
## Detailed Description

After filling in the new channel form, I should see a summary of the channel config before finally submitting it so that I can verify everything is how I want it to be. Similar to what the user sees when making payments.

image

## Context

New Channel UI

mrfelton:
## Description:

Build new channel search form

## Motivation and Context:

Resolves #1571

## How Has This Been Tested?

Manually

image

## Types of changes:

New Feature

## Checklist:


app/reducers/contactsform.js
```diff
@@ -201,6 +201,17 @@ contactFormSelectors.filteredNetworkNodes = createSelector(
   }
 )
 
+contactFormSelectors.isSearchValidNodeAddress = createSelector(
+  searchQuerySelector,
+  searchQuery => {
+    if (!searchQuery) {
+      return false
+    }
+    const [pubkey, host] = searchQuery.split('@')
```
korhaliv: Could be a little more efficient if you first check for `searchQuery.length` to be equal 2

app/components/Channels/ChannelNodeSearch.js
```diff
@@ -0,0 +1,128 @@
...
+    {filteredNetworkNodes.map(node => {
+      return (
+        <React.Fragment key={node.pub_key}>
+          <Flex justifyContent="space-between" alignItems="center" py={2} pr={2}>
+            <Box mr={3}>
+              {node.alias.length > 0 ? (
+                <>
+                  <ClippedText>{node.alias.trim()}</ClippedText>
+                  <Text color="gray" fontSize="s">
+                    <Truncate text={node.pub_key} maxlen={25} />
```
korhaliv: What's the difference between `ClippedText` and `Truncate` ? Sounds similar judging from the names

app/reducers/contactsform.js
mrfelton: I don't think that would have the same effect. I specifically want to ensure that we have something before and after an `@` to ensure that what we have looks like a node address.

app/components/Channels/ChannelNodeSearch.js
mrfelton: `ClippedText` chips off long text with an ellipsis. `Truncate` truncates the middle of a string.

app/reducers/contactsform.js
korhaliv: Yes, that's why I said first. So it's an addition. You don't need to split if length != 2 right?

mrfelton: Got it. Updated.

PR #1616, opened 2019-02-21
fix(ui): node and activity modals overhaul

### LINKED ISSUE #1614: Address modal bug (next branch)
In the next branch when clicking on the `qrcode` to access my address, URI and lndconnect QRCode I see the following UI:

screen shot 2019-02-20 at 6 50 44 pm

Also, clicking on `Node URI` makes the screen go blank

korhaliv:
## Description:
Bring node details and activity modals under the common modal system (modal stack)

## Motivation and Context:
Currently node and activity modals are standalone and thus are not benefiting from the existing modal system (transition animation, proper escape button handling etc), as well as causing rendering issues.

Modals are now part of the common stack and are also bound closer to a redux state source resulting in a minor perf improvements.

This PR also does some cleanup and refactoring of the top level `App` container

Fix #1614 

## How Has This Been Tested?
Manually

## Types of changes:
Bugfix and refactoring

## Checklist:

### DIFF
--- a/app/components/Activity/ActivityModal/ActivityModal.js
+++ b/app/components/Activity/ActivityModal/ActivityModal.js
@@ -1,7 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Box } from 'rebass'
-import { Modal } from 'components/UI'
 import { InvoiceModal } from 'components/Activity/InvoiceModal'
 import { PaymentModal } from 'components/Activity/PaymentModal'
 import { TransactionModal } from 'components/Activity/TransactionModal'
@@ -13,12 +12,11 @@ export default class ActivityModal extends React.PureComponent {
       id: PropTypes.string,
       name: PropTypes.string
     }),
-    hideActivityModal: PropTypes.func.isRequired,
     showNotification: PropTypes.func.isRequired
   }
 
   render() {
-    const { item, networkInfo, hideActivityModal, showNotification } = this.props
+    const { item, networkInfo, showNotification, ...rest } = this.props
 
     if (!item) {
       return null
@@ -42,11 +40,9 @@ export default class ActivityModal extends React.PureComponent {
     const SpecificModal = MODAL_COMPONENTS[item.type].component
 
     return (
-      <Modal onClose={hideActivityModal}>
-        <Box width={9 / 16} mx="auto">
-          <SpecificModal {...MODAL_COMPONENTS[item.type].props} />
-        </Box>
-      </Modal>
+      <Box {...rest}>
+        <SpecificModal {...MODAL_COMPONENTS[item.type].props} />
+      </Box>
     )
   }
 }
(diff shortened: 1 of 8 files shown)


app/components/Wallet/ReceiveModal/ReceiveModal.js
```diff
@@ -123,61 +110,59 @@ class ReceiveModal extends React.PureComponent {
...
-            {qrCode && qrCodeType !== QRCODE_TYPE_LNDCONNECT && (
-              <QRCode value={qrCode} size="xlarge" />
-            )}
-            {qrCodeType === QRCODE_TYPE_LNDCONNECT && (
-              <>
-                <QRCode
-                  value={qrCode}
-                  size="xlarge"
-                  obfuscate={qrCodeType === QRCODE_TYPE_LNDCONNECT && !reveal}
+      <Box width={9 / 16} mx="auto">
```
mrfelton: Rather than hard coding the width and margin onto this component, it would be better to apply `{...rest}` to the wrapping `Box`, and have these props passed in by the caller.

app/components/Activity/ActivityModal/ActivityModal.js
```diff
@@ -42,11 +40,9 @@ export default class ActivityModal extends React.PureComponent {
     const SpecificModal = MODAL_COMPONENTS[item.type].component
 
     return (
-      <Modal onClose={hideActivityModal}>
-        <Box width={9 / 16} mx="auto">
-          <SpecificModal {...MODAL_COMPONENTS[item.type].props} />
-        </Box>
-      </Modal>
+      <Box width={9 / 16} mx="auto">
```
mrfelton: Same goes here. positioning props would be better passed in from the caller.

app/containers/ModalStack.js
```diff
@@ -25,6 +27,12 @@ const ModalContent = ({ type, closeModal }) => {
     case 'REQUEST_FORM':
       return <Request width={9 / 16} mx="auto" />
 
+    case 'RECEIVE_MODAL':
+      return <ReceiveModal width={9 / 16} mx="auto" />
+
+    case 'ACTIVITY_MODAL':
+      return <ActivityModal width={9 / 16} mx="auto" />
```
mrfelton: so the positioning props are already being passed in,. the just aren't being applied correctly.

app/components/Wallet/ReceiveModal/ReceiveModal.js
korhaliv: I didn't hardcode them (this is how it was), but your suggestion makes sense

korhaliv: updated

app/components/Activity/ActivityModal/ActivityModal.js
korhaliv: updated

app/containers/ModalStack.js
korhaliv: updated

PR #1620, opened 2019-02-22
 feat(ui): add ability to edit host cert and macaroon

### LINKED ISSUE #1182: Ability to edit cert & macaroon paths in wallet settings screen
## Detailed Description

I should be able to see and edit the cert and macaroon paths in the wallet settings screen so that I can more easily understand the wallet config.

## Context

Hard to know which cert and macaroon files are being used for a remote host.

## Possible Implementation

Add fields for cert and macaroon path into wallet settings form (remote wallets only)

korhaliv:
## Description:
resolves #1182 

## How Has This Been Tested?
Manually

image

## Types of changes:
New feature

## Checklist:


app/lib/utils/isBase64url.js
```diff
@@ -0,0 +1,15 @@
...
+ * Checks whether specified string is base64 url encoded
+ * @returns {boolean}
+ */
+const isBase64url = str => {
+  try {
+    //convert from base64 url to base64
+    window.atob(str.replace(/-/g, '+').replace(/_/g, '/'))
+    return true
+  } catch (e) {
+    // console.log(e)
```
mrfelton: console.log line can go

app/lib/utils/isBase64url.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Checks whether specified string is base64 url encoded
+ * @returns {boolean}
+ */
+const isBase64url = str => {
+  try {
+    //convert from base64 url to base64
```
mrfelton: I think this should be written as `base64url` (no space)

app/components/Home/WalletLauncher.js
```diff
@@ -132,19 +155,65 @@ class WalletLauncher extends React.Component {
     }
   }
 
-  onSubmit = values => {
-    const { startLnd } = this.props
+  onSubmit = async values => {
+    const { startLnd, wallet } = this.props
+    //for the remote wallets we need to initiate lndconnect URI re-generation before launching
```
mrfelton: Add space after `//`.

app/components/Home/WalletLauncher.js
```diff
@@ -132,19 +155,65 @@ class WalletLauncher extends React.Component {
...
+      const formState = formApi.getState()
       const { values } = formState
-      putWallet(formToWalletFormat(values))
+      let result = values
+
+      // for the remote wallets we need to re-generate lndconnectUri and QR using updated
+      // host, cert and macaroon values. This is done in the main process
+      // this process is skipped if original lndconnect uri contains raw cert or macaroon and not paths
+      if (wallet.type !== 'local' && !isEmbeddedLndConnectURI(wallet.lndconnectUri)) {
+        ipcRenderer.send(
```
mrfelton: Should we be using `dispatch` here rather than sending via `ipcRenderer` directly? I think there is only one other place in the codebase where we do this, and I'm not particularly happy about that. All other IPC messages are sent using `send` from `redux-electron-ipc`

app/components/Home/WalletLauncher.js
korhaliv: Ideally we need lint rule for this

app/components/Home/WalletLauncher.js
korhaliv: By dispatch you mean `redux dispatch` ? If so, then this doesn't update the redux state so doesn't seem appropriate here

mrfelton: It does in the sense that It logs the action in the redux history so that you have visibility into the fact that an IPC message was sent within the redux history. This is how all of our IPC messages are sent.

korhaliv: I get that it's how it's currently done in most places, but benefits are not clear to me. Especially when seeing code like

mrfelton: I would definitely agree with that. The whole process of using IPC is horrible IMO. Sending off events and hoping that certain other events come back in response... I'd much rather see a fully promise based approach to communicating with the main process. But short of that sort of wider change, I do feel there is some merit to taking a consistent approach across the board.

korhaliv: Updated

korhaliv:
@JimmyMow I understand the motivation, but there may be a situation when a user has installed the wallet some time ago and may not remember the details, so the hint will be quite handy here. After all we even support older versions of zap db formats(via migration scripts) to take care of longer time users

mrfelton:
Agreed - I think it's better to keep the help text on the page to help users fill out the form. IMO it's better that users are shown the expected format than them getting confused and posting support questions here or in slack.

mrfelton:
@korhaliv on the issue of the field labels...

At the default screen size they look a bit squashed up and run over to multiple lines with a break in a strange place. See screenshot:

**With examples**

image

**Compare without examples:**

image

It definitely looks a lot cleaner without the examples. So I'm a little torn on the best thing to do as I think the examples would be useful to people (actually, the reason  why we added the examples to the onboarding forms in the first place was because people kept asking us questions about how they should enter the paths)

One option could be for us to leave out the examples for now, and then consider adding them back in as tooltips once #1291 is in. Or, perhaps we could add the examples as a hover tooltip on the inputs themselves?

JimmyMow:
@mrfelton this is the motivation behind my comment above. I like the tooltip idea

korhaliv:
Ok, I've removed examples. I think for consistency sake it's better to do tooltips the way #1291 does it, so let's wait until it's merged and update settings screen once more

PR #1621, opened 2019-02-22
 feat(channels): private channel toggle

### LINKED ISSUE #1570: I should be able to select wether a channel will be private or not when creating it
## Detailed Description

Add private toggle to add channel form

image

## Context

New Channel UI

mrfelton:
## Description:

Add ability to specify wether a channel will be created as private or not. Defaults as true for local wallets.

## Motivation and Context:

Resolves #1570

## How Has This Been Tested?

Manually

## Types of changes:

New feature

## Checklist:


mrfelton:
@JimmyMow I've removed the on/off labels - I agree these aren't really needed.

I've also updated so that the field only shows for remote wallets. Part of me wants to say that we shouldn't be telling users what they can or can't do with their local nodes. But a bigger part of me says that it makes very little sense for a local wallet to be creating public channels in the vast majority of cases.

PR #1635, opened 2019-02-25
fix(lnd): defer fetching data until sync complete

mrfelton:
## Description:

Defer fetching activity/channel data from lnd until sync is complete.

## Motivation and Context:

Attempting to fetch channel data can cause neutrino to crash with an error like `unable to find arbitrator`.

There is no need for us to attempt to fetch this data until we mount the App, (after syncing has completed in the case of local neutrino wallets).

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/app/components/App/App.js
+++ b/app/components/App/App.js
@@ -34,17 +34,17 @@ class App extends React.Component {
     channelFormProps: PropTypes.object,
     setIsWalletOpen: PropTypes.func.isRequired,
     fetchPeers: PropTypes.func.isRequired,
-    fetchDescribeNetwork: PropTypes.func.isRequired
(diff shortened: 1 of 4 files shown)


korhaliv:
Concept ACK. @mrfelton I wonder if we should dedicate a special method(as opposed to componentDidMount of some component) where we place initial fetches ? This way it will be more pronounced and it will be more straightforward to add additional fetches there if future.

mrfelton:
@korhaliv I've updated this so that the calls are all now made in `fetchActivityHistory` which acts as a sort of meta fetch method that triggers fetching of all related data.

PR #1640, opened 2019-02-26
feat(channels): categorise opening/funding transactions

### LINKED ISSUE #902: Properly categorise funding and closing transactions in the activity list
## Description

All my channel closing txs resulted in "Sent 0 BTC" being showin in both "All" and "Sent" tabs. It is confusing because I didn't actually send any BTC to anyone.

## Expected Behavior

Either don't show these transactions at all or show them as "Received on chain"

## Your Environment

- Zap Desktop version: 0.2.2-beta.4225
- Operating System and version: OS X 10.14

mrfelton:
## Description:

Categorize channel funding and closing transactions by cross referencing transactions with channel data.

The PR removes opening/funding transactions from the `All` activity list, and moves them over to a separate 'Internal` tab.

The layout / organisation within the activity list could probably be improved, but I think changes to the filters and tabs is probably outside of the scope of this PR and should be addressed separately.

## Motivation and Context:

Fix #902

## How Has This Been Tested?

Manually

image

## Types of changes:

Enhancement

## Checklist:


app/lib/lnd/methods/index.js
```diff
@@ -107,12 +107,17 @@ export default function(lightning, log, event, msg, data) {
...
+          channelController.pendingChannels,
+          channelController.closedChannels
+        ].map(func => func(lnd))
       )
         .then(channelsData =>
           event.sender.send('receiveChannels', {
             channels: channelsData[0].channels,
-            pendingChannels: channelsData[1]
+            pendingChannels: channelsData[1],
+            closedChannels: channelsData[2].channels
```
korhaliv: is `channelsData[2]` always set?

app/components/Activity/Transaction/Transaction.js
```diff
@@ -6,67 +6,78 @@ import { Message, Text } from 'components/UI'
...
-          )}
-        </>
-      ) : (
-        <Text color="gray" fontSize="xs" fontWeight="normal">
-          <FormattedTime value={transaction.time_stamp * 1000} />
-        </Text>
-      )}
-    </Box>
+const Transaction = ({ transaction, showActivityModal, currencyName, intl }) => {
+  const type = transaction.isFunding
```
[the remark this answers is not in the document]
mrfelton: Couldn't really find a particularly nice way to do this, but I have switched from a multi-level ternary sequence to a single ternary plus some conditional overrides with `if/else if`. It's a little more readable at least.

app/lib/lnd/methods/index.js
mrfelton: Yes - `channelsData[2]` is the spread from `Promise.all` above

app/reducers/activity.js
```diff
@@ -336,11 +337,21 @@ const pendingActivity = createSelector(
   invoices => groupAll(invoices.filter(invoice => !invoice.settled && !invoiceExpired(invoice)))
 )
 
+const internalActivity = createSelector(
+  transactionsSelector,
+  transactions => {
+    return groupAll([
+      ...transactions.filter(transaction => transaction.isFunding || transaction.isClosing)
```
korhaliv: Is there a need to do a spread since `filter` already returns a new array?

app/reducers/channels.js
```diff
@@ -657,6 +662,15 @@ channelsSelectors.closingChannelIds = createSelector(
   ]
 )
 
+channelsSelectors.closedChannels = createSelector(
+  closedChannelsSelector,
+  nodesSelector,
+  closingChannelIdsSelector,
+  (closedChannels, nodes, closingChannelIds) => [
+    ...closedChannels.map(channelObj => decorateChannel(channelObj, nodes, closingChannelIds))
```
korhaliv: Is there a need to do a spread since `map` already creates a new array?

app/reducers/activity.js
mrfelton: Good catch. You;'re right. I have updated this.

app/reducers/channels.js
mrfelton: Same - updated here.

mrfelton:
>  Can we have more friendly copy? `COOPERATIVE_CLOSE`, `REMOTE_FORCE_CLOSE`, etc are a bit weird. Would be much better if we could map those to more friendlier descriptions like `Cooperative channel close`, `Remote party force close`, etc

For sure.

> I'm not sold on the `Internal` copy for the tab

Me neither. The primary purpose of this PR was to remove those confusing transactions from the main activity list. I just added the `internal` tab as I thought it made sense that there was still some way at least to see them.

I also barely ever use the tabs. Actually, I think Ole did some designs for an improved search/filter section that is more similar to the new one that we have in the new Channels section, in which the filters would be in a dropdown - so not right in your fave all the time.

I think for the purposes of this PR though, I really want to focus on the core functionality of being able to properly categorise and show/hide these confusing transactions.

What about - until we get round to redoing the search/filter section on the activity list - we add another button to the footer bar `Show internal transactions` (next to where we have `Show expired` for the invoices)?

JimmyMow:
@mrfelton I think that's a great compromise. So how do you envision the UI of the buttons? Stacking buttons like that may be weird, but I'm not sure I have an idea off the top of my head

mrfelton:
I've gone ahead and merged this as is, and will submit follow up PR's to adjust the tabs.

PR #1665, opened 2019-02-28
feat(channels): validate channel amount input

### LINKED ISSUE #1494: Channel funding dialog should prevent user from requesting a channel size that exceeds available wallet balance
## Expected Behavior

If wallet balance is 0.001 tBTC, Zap should constrain the balance in the channel-open dialog such that the channel balance + TX fees ≤ 0.001 tBTC.

## Actual Behavior

In this wallet with 0.001 tBTC, the dialog box still turns green when 0.005 tBTC is entered as the proposed channel amount:

image

Upon submission, Zap then displays the user-unfriendly error message `Error: not enough witness outputs to create funding transaction`:

image

## Possible Fix

Constrain the balance in the channel-open dialog such that the channel balance + TX fees ≤ confirmed wallet balance.  Confirmed wallet balance is readily available via https://api.lightning.community/#walletbalance

It doesn't look like there is a way to ask https://api.lightning.community/#openchannel to craft a transaction without actually _sending_ the transaction.  Is that the problem?  If so, I'll open something against LND.

## Steps to Reproduce

1. fund Zap Wallet
2. attempt to open channel
3. specify channel size equal or greater to the amount of available funds

## Context

This is especially confusing to new users when they attempt to create a channel with their entire balance, it fails because TX fees are not considered, and the error message isn't enlightening.

## Your Environment

- Zap Desktop version: v0.3.3-beta
- Operating System and version: Ubuntu Linux 18.04

mrfelton:
## Description:

Do not allow submission of the channel create form if the amount is greater than the total confirmed wallet balance.

**NOTE 1:** I have updated the footer text to show the confirmed wallet balance, rather than the total balance, since only confirmed UTXOs are relevant in this context.

**NOTE 2:** We only know the per-byte fee, so can not fully cater for the fee amounts. We need to wait for the new fee estimation api in lnd in order to know the full fee value.

## Motivation and Context:

Resolves #1494

## How Has This Been Tested?

Manually

## Screenshots:

image

## Types of changes:

Enhancement

## Checklist:

### DIFF
--- a/app/components/Channels/ChannelCreateForm.js
+++ b/app/components/Channels/ChannelCreateForm.js
@@ -257,6 +257,7 @@ class ChannelCreateForm extends React.Component {
           validateOnChange={formState.submits > 0}
           validateOnBlur={formState.submits > 0}
           required
+          css={{ height: '88px' }}
         />
 
         <Bar my={3} opacity={0.3} />
--- a/app/components/Channels/ChannelCreateForm.js
+++ b/app/components/Channels/ChannelCreateForm.js
@@ -170,7 +170,7 @@ class ChannelCreateForm extends React.Component {
     const [pubkey, host] = nodePubkey.split('@')
 
     // Determine the fee rate to use.
-    const speed = values.speed || defaultSpeed
+    const { speed = defaultSpeed } = values
     const satPerByte = this.getFeeRate(speed)
 
     // submit the channel to LND.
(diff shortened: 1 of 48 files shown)


mrfelton:
We can try it @JimmyMow , although I think it might look a bit odd with a large bit of blank space when there are no errors. Also, the height pf the form is a real issue. Actually, when the error is visible, the form is too tall for the page - look how the submit button is right up against the private toggle in my screenshot.

We need to loose some height somewhere, or make the default page height taller (which I'm not too keen on as I think the aspect ratio is already a bit weird since we made the default width smaller without altering the default height).

mrfelton:
@JimmyMow I pushed an an updated version with 2 changes:

1. Use fixed height for amount field group so that it doesn't jump when an error appears.
2. Moved the fee speed description under the estimated field amount on the right to save some height in that section

**No error:**

image


**With error:**

image

It's better  - but it is still a little too tall though unfortunately, although not by a huge amount (the button is still a little too close to the bottom of the form. I'm not sure where else we can loose height though.

mrfelton:
I guess we could loose the `Valid pubkey` green indicator? That would claw back a little height, and if you think about it, the only way to get to this form in the first place is with a valid pubkey so it's kinda redundant. Although, that green `Valid pubkey` check mark is consistent with the pay forms, so from that perspective it's right that it should be here too.

korhaliv:
@mrfelton Works good but I have some UI concerns/discussion point. When you switch between fee tires there is a significant jump in the area outline of the screenshot.

image

image

Maybe right-align this block ?

mrfelton:
@korhaliv looks like there was a typo on one of the flex properties. Have updated and resolved the alignment issue.

korhaliv:
Ok, much better now. Another thing is a little bit weird `Next` button behavior. As you can see it's enabled on the screenshot, but as soon as i click it just goes disabled without any additional feedback

mrfelton:
@korhaliv the issue with the submit button should be fixed now.

PR #1673, opened 2019-03-01
fix(ui): include suggested nodes in search results

korhaliv:
## Description:
Include suggested nodes in search results in create channel node search view

## Motivation and Context:
Currently we don't include suggested nodes in the search results

## How Has This Been Tested?
Manually

## Types of changes:
Enhancement

## Checklist:

### DIFF
--- a/app/reducers/contactsform.js
+++ b/app/reducers/contactsform.js
@@ -60,11 +60,22 @@ contactFormSelectors.suggestedNodes = createSelector(
   }
 )
 
+/**
+ * @param {array} suggestedNodes
+ * @return search results compatible version of suggested nodes array
+ */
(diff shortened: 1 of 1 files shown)


korhaliv:
@mrfelton @JimmyMow As a further improvement I'd recommend to rank suggested nodes higher in the search results and/or add `suggested` badge in some form. What are your thoughts on this?

mrfelton:
I'm going to merge this as is since it resolves the issue with suggested nodes not being included in the listing at all.

I think that ranking the suggested nodes higher in the search results is also a good suggestion. We already do the same thing with connected peers.

I'm not sure about showing a badge though

PR #1677, opened 2019-03-01
feat(activity): new activity search and filters

mrfelton:
## Description:

- Reimplement activity search and filters following the model provided by the new channel UI
- Replace show/hide expired footer button with new filter category
- Add new `Received` category to mirror existing `Sent` category
- Tidy up internal transactions list
- Fix search widget so that it correctly works on all filter categories

## Motivation and Context:

- Too many activity filter tabs looks very in your face and intimidating
- Be consistent in our UI (match new channels UI)
- Improve filters

## How Has This Been Tested?

Manually

## Screenshots:

image

## Types of changes:

Enhancements

## Checklist:


mrfelton:
Agreed. I've adjusted the length.

app/components/Activity/ActivityActions/ActivityRefresh.js
```diff
@@ -0,0 +1,66 @@
...
+    await onClick()
+    await delay(1000)
+    setStatus('done')
+  }
+
+  useEffect(() => {
+    if (status === 'fetching') {
+      fetchActivity()
+    }
+  }, [status])
```
korhaliv: Do you think it will be cleaner if combined into one `useEffect` since it basically analyzes the same `status` flag and decided what to do based on it's value?

app/containers/Activity/ActivityActions.js
```diff
@@ -0,0 +1,20 @@
...
+
+const mapDispatchToProps = {
+  changeFilter,
+  fetchActivityHistory,
+  updateActivitySearchQuery: updateSearchText
+}
+
+const mapStateToProps = state => ({
+  filter: state.activity.filter,
+  filters: state.activity.filters,
```
korhaliv: In most cases it's better to avoid variable names that differ in only one letter like `thing` and `thingS` because it's very error prone and harder to read

app/reducers/activity.js
```diff
@@ -266,91 +255,111 @@ function groupAll(data) {
...
+  invoicesSelector,
+  transactionsSelector,
+  (searchText, invoices, transactions) => {
+    const allData = [
+      ...invoices.filter(invoice => invoice.settled),
+      ...transactions.filter(
+        transaction => transaction.received && !transaction.isFunding && !transaction.isClosing
+      )
+    ]
+    return searchText ? groupAll(applySearch(allData, searchText)) : groupAll(allData)
```
korhaliv: I'm seeing this routine repeating multiple times in this module. Might be a good idea to extract it into a helper function

app/reducers/transaction.js
```diff
@@ -230,11 +232,17 @@ transactionsSelectors.transactionsSelector = createSelector(
         const channelData = getChannelData(channelObj)
         return [channelData.closing_tx_hash, channelObj.closing_txid].includes(transaction.tx_hash)
       })
+      const pendingChannels = [...closingPendingChannelsRaw, ...pendingOpenChannelsRaw].find(
```
korhaliv: the name of the constant is `pendingChannelS` but find returns a single value (or non at all), so probably should be named `pendingChannel`

app/components/Activity/ActivityActions/ActivityActions.js
```diff
@@ -0,0 +1,53 @@
...
+
+    <Flex as="section" alignItems="center" ml={3}>
+      <ActivityRefresh onClick={fetchActivityHistory} />
+    </Flex>
+  </Flex>
+)
+
+ActivityActions.propTypes = {
+  filter: PropTypes.string.isRequired,
+  filters: PropTypes.array.isRequired,
```
korhaliv: If possible,  i'd like to have something like `filter` and `allFilters` to make it more error prone

app/components/Activity/ActivityActions/ActivityFilter.js
```diff
@@ -0,0 +1,38 @@
...
+        highlightOnValid={false}
+      />
+    </Form>
+  )
+}
+
+ActivityFilter.propTypes = {
+  changeFilter: PropTypes.func.isRequired,
+  filter: PropTypes.string,
+  filters: PropTypes.array
```
korhaliv: Same comment here as above

app/components/Activity/ActivityActions/ActivitySearch.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import debounce from 'lodash.debounce'
+import { Form, Input } from 'components/UI'
+
+const ActivitySearch = ({ searchQuery, placeholder, updateActivitySearchQuery, ...rest }) => {
+  const debouncedUpdateActivitySearchQuery = debounce(updateActivitySearchQuery, 300)
```
korhaliv: In theory this will create debounced function on each render causing debounce to work incorrectly. I think it's better to create one with `useEffect` hook

app/components/Activity/Transaction/Transaction.js
```diff
@@ -29,8 +29,11 @@ const Transaction = ({ transaction, showActivityModal, currencyName, intl }) =>
         data-hint={intl.formatMessage({ ...messages.type })}
       >
         <Text mb={1}>
-          <FormattedMessage {...messages[type]} />
-          {transaction.closeType && <Text color="gray">{transaction.closeType}</Text>}
+          {!transaction.closeType ? (
```
korhaliv: Consider removing `!` and switching components

app/containers/Activity/ActivityActions.js
mrfelton: I think we touched on this before. I agree in principal, but the current code style of the app is to do it this way.

I'd like to keep things consistent across the codebase, so if we are going to change it (and I think we should), that should change should be made across the entire codebase as a standalone PR.

app/reducers/transaction.js
mrfelton: Updated

app/components/Activity/ActivityActions/ActivityActions.js
mrfelton: same as above

app/components/Activity/ActivityActions/ActivityFilter.js
mrfelton: here too

app/components/Activity/Transaction/Transaction.js
mrfelton: Updated.

There is actually a lint rule that I usually use to enforce no negatives in conditionals as it improves readability a lot IMO.

See https://eslint.org/docs/rules/no-negated-condition

I've added to the list in https://github.com/LN-Zap/zap-desktop/issues/1559

app/reducers/transaction.js
mrfelton: Updated.

app/components/Activity/ActivityActions/ActivitySearch.js
mrfelton: I took a stab at implementing this using hooks, but it wasn't really working as I wanted it to. Rather than spend too much time on this here I have converted this to a `PureComponent` and set up the debounce as a class method to remove the redundant renders. This the same approach as we take elsewhere in the app.

Converting all of our debounce to a standardised hook implementation would make a nice future refactor PR.

app/components/Activity/ActivityActions/ActivityRefresh.js
korhaliv: I think in the tip you are referring to they recommend to split `unrelated` code. In this case I'd say it's a regular action in redux terms: start fetching, fetch, some UI feedback when done. So usually a monolithic piece of code. I don't think it's critical by any means,  so up to you.

app/components/Activity/ActivityActions/ActivitySearch.js
korhaliv: I wonder what was the issue you were having and/or the particular hook you were using.

app/components/Activity/ActivityActions/ActivityRefresh.js
mrfelton: I've updated this as per your suggestion. It does read better when combined into a single hook in this case.

app/reducers/activity.js
mrfelton: Updated

app/components/Activity/ActivityActions/ActivitySearch.js
mrfelton: Got this working and updated the code top use new `useDebounce` hook

PR #1695, opened 2019-03-04
feat(channels): show closing transaction in channel detail

### LINKED ISSUE #1687: I should be able to see the closing transaction id for channels that are in the process of closing
## Detailed Description

When waiting for a channel to close, you want to be able to access at the details of the closing transaction so that you can track it's progress more easily.

Currently we show the funding transaction id, but not the closing transaction id.

## Context

Channel UI

## Possible Implementation

Show the closing transaction id on the channel detail page for transactions that are in the process of closing.

mrfelton:
## Description:

Show the closing transaction id for closing channels in the channel detail view and clean up the view to remove unknown properties.

## Motivation and Context:

Fix #1687

## How Has This Been Tested?

View channel details for closing transactions.

## Screenshots:

image

## Types of changes:

Enhancement

## Checklist:


app/components/Channels/ChannelCardList.js
```diff
@@ -38,19 +37,18 @@ const ChannelCardList = ({
...
-            networkInfo={networkInfo}
-          />
-        </Card>
+        <ChannelCardListItem
+          channel={channel}
+          currencyName={currencyName}
+          openModal={openModal}
+          setSelectedChannel={setSelectedChannel}
+          networkInfo={networkInfo}
+          css={{ height: '100%' }}
```
korhaliv: I wonder what's the reason behind this change ? We have a static row height of the list items

mrfelton: In the basic card view (shown on the list page), there are 3 rows of data. In some cases, we don't have values for all of these rows (eg, force closing channels).

Previously, we should show the rows but with values `unknown`, which looked a bit strange and raises the question of why the value is unknown.

In this PR, this is changed so that we only show data rows when we actually have values to show. This means that rather than always being 3 rows, we now display *up to* 3 rows.

In the case where there is only 1 row, we want to ensure that the card still takes up all available height.

PR #1696, opened 2019-03-04
feat(wallet): show confirmations in transaction detail

### LINKED ISSUE #1686: I should be able to see the number of confirmations a transaction has
## Detailed Description

Display the number of confirmations that a transaction has had, so that users can more easily understand the state of their onchain transactions.

## Context

Currently, we only display the block number that the transaction was mined in and users have to click through to the block explorer in order to see the number of confirmations.

Given that on chain transactions are only available for use within channels after 3 confirmations (by default), its pretty common that users want to know how many confirmations their transaction has had.

## Possible Implementation

Add a new row to the transaction detail data table

mrfelton:
## Description:

Show the number of blockchain confirmations in the transaction detail view so that users have this information at their fingertips rather than needing visit a block explorer in order to find it out.

## Motivation and Context:

Fix #1686

## How Has This Been Tested?

Manually - view transaction detail for unconfirmed transaction.

## Screenshots:

**Unconfirmed transaction**:

image

**Confirmed transaction:**

image

## Types of changes:

Enhancement

## Checklist:

### DIFF
--- a/app/components/Activity/TransactionModal/TransactionModal.js
+++ b/app/components/Activity/TransactionModal/TransactionModal.js
@@ -1,9 +1,10 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { FormattedDate, FormattedTime, FormattedMessage } from 'react-intl'
+import get from 'lodash.get'
+import { FormattedDate, FormattedTime, FormattedMessage, FormattedNumber } from 'react-intl'
 import { Flex } from 'rebass'
 import { blockExplorer } from 'lib/utils'
-import { Bar, DataRow, Header, Link, Panel } from 'components/UI'
+import { Bar, DataRow, Header, Link, Panel, Text } from 'components/UI'
 import { CryptoSelector, CryptoValue, FiatSelector, FiatValue } from 'containers/UI'
 import { Truncate } from 'components/Util'
 import Onchain from 'components/Icon/Onchain'
@@ -35,6 +36,7 @@ export default class TransactionModal extends React.PureComponent {
 
   render() {
     const { item, ...rest } = this.props
+    const destAddress = get(item, 'dest_addresses[0]')
 
     return (
       <Panel {...rest}>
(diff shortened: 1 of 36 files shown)


mrfelton:
Have updated the text there @JimmyMow

app/components/Activity/TransactionModal/TransactionModal.js
```diff
@@ -72,38 +72,53 @@ export default class TransactionModal extends React.PureComponent {
...
+                right={
+                  item.num_confirmations ? (
+                    <>
+                      <FormattedDate
+                        value={item.time_stamp * 1000}
+                        year="numeric"
+                        month="long"
+                        day="2-digit"
+                      />
+                      <br />
```
korhaliv: Is there a need for `<br/>` here ?

app/components/Activity/TransactionModal/TransactionModal.js
```diff
@@ -72,38 +72,53 @@ export default class TransactionModal extends React.PureComponent {
...
+              />
+
+              <Bar />
+
+              <DataRow
+                left={<FormattedMessage {...messages.address} />}
+                right={
+                  <Link
+                    className="hint--bottom-left"
+                    data-hint={item.dest_addresses[0]}
```
korhaliv: Consider extracting `item.dest_addresses[0]` into a const to improve readability

app/components/Activity/TransactionModal/TransactionModal.js
mrfelton: Yes - I want this to show over 2 lines

app/components/Activity/TransactionModal/TransactionModal.js
mrfelton: Updated

app/components/Activity/TransactionModal/TransactionModal.js
korhaliv: I mean it would probably be cleaner to do this with regular markup means. `brs` are usually not very flexible in terms of padding/margin etc

mrfelton: Ahh, yeah sorry - don't know why that didn't click. I've updated to use `Text` components.

PR #1737, opened 2019-03-08
feat(ui): add contextual help component

mrfelton:
## Description:

Rebase #1501 and make some minor adjustments to the style so that it's more inline with our existing tooltip styles and makes better use of rebass style props.

## Motivation and Context:

Addresses #1291

## How Has This Been Tested?

Manually

image

## Types of changes:

New Feature

## Checklist:


JimmyMow:
@mrfelton why does the copy there say it will be visible to you only? That isn't true. When the receiver decodes the request they will see the memo. The memo is intended to provide context to the user paying the request

mrfelton:
You're right @JimmyMow ! I think this was just taken directly from the visual in Zeplin. Could you provide some updated verbiage please?

JimmyMow:
Sure. How about:

`Add some describer text to your payment request for the recipient to see when paying.`

mrfelton:
@JimmyMow I've updated the text for this.

PR #1752, opened 2019-03-09
feat(wallet): ability to set on-chain transaction fees

mrfelton:
## Description:

Extract fee widget to TransactionFeeInput component and add to the Pay form.

## Motivation and Context:

Allow users to set their fee when making on-chain transactions.

Address #1754

## How Has This Been Tested?

Manually

## Types of changes:

Feature

## Checklist:


app/components/Pay/Pay.js
```diff
@@ -223,22 +262,18 @@ class Pay extends React.Component {
...
+        // Determine the fee rate to use.
+        const satPerByte = this.getFee()
+
+        // Send the coins.
         sendCoins({
           addr: values.payReq,
           value: values.amountCrypto,
           currency: cryptoCurrency,
-          satPerByte: onchainFees.fastestFee,
+          satPerByte: satPerByte,
```
korhaliv: Can be simplified to just `satPerByte`

app/components/UI/TransactionFeeInput.js
```diff
@@ -0,0 +1,85 @@
...
+  initialValue: PropTypes.string,
+  isQueryingFees: PropTypes.bool,
+  label: PropTypes.string,
+}
+
+TransactionFeeInput.defaultProps = {
+  initialValue: TRANSACTION_SPEED_SLOW,
+}
+
+export default withFormApi(TransactionFeeInput)
```
korhaliv: From what I see, `formApi` is only used to do `formApi.getValue(field)`. This can be achieved cleaner with `withFieldState` HOC which has narrower scope. So you can dynamically bind Input to a particular field by either exporting a binding function or binding it closer to where input is used (like `const MyInput = withFieldState('field')(TransactionFeeInput)`)

mrfelton: Doesn't `withFieldState` imply that you have already created a field outside of the scope of the component that you want to pass in so that the component can control that already existing field?

In the case here, the `TransactionFeeInput` component is responsible for creating it's own field and is fully self contained, so it doesn't relay on an externally created field.

PR #1812, opened 2019-03-15
[WIP] fix(channels): notification when channel enters pending state

JimmyMow:
## Description:

Give a notification when a freshly submitted channel is successfully broadcasted and enters a pending state

## Motivation and Context:

Right now we only notify when a user simply submits the channel form and have gotten feedback that this isn't too great.

There are a few things wrong with this pull request as it stands:

- [ ] The notification is a hard-coded string and needs to come from the channels `messages` file
- [ ] A user now gets back to back notifications on a successful channel submit.

Screen Shot 2019-03-14 at 9 32 56 PM

I think instead the form submit button should enter a `loading` state where the submit button displays our spinner until we get a response. On a success response, we can close the form and show the success notification. On an error, we can leave the loading state but keep the form open, and show an error notification.

@mrfelton @korhaliv feel free to pick this up. Given my schedule I'm not sure I will be able to finish this one

### DIFF
--- a/app/reducers/channels.js
+++ b/app/reducers/channels.js
@@ -2,7 +2,7 @@ import { createSelector } from 'reselect'
 import throttle from 'lodash.throttle'
 import { send } from 'redux-electron-ipc'
 import { requestSuggestedNodes } from 'lib/utils/api'
-import { showError } from './notification'
+import { showNotification, showError } from './notification'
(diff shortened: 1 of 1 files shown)


mrfelton:
Closing in favour of https://github.com/LN-Zap/zap-desktop/pull/1834

PR #1818, opened 2019-03-15
feat(channels): include loading channels in listings

mrfelton:
## Description:

Include loading channels in listings

## Motivation and Context:

Provide feedback to the user when opening channels

## How Has This Been Tested?

Manually

## Types of changes:

Enhancement

## Checklist:

### DIFF
--- a/app/components/Channels/ChannelData.js
+++ b/app/components/Channels/ChannelData.js
@@ -24,7 +24,7 @@ const ChannelData = ({ channel, currencyName, networkInfo, viewMode, ...rest }) 
     csv_delay,
     num_updates,
   } = channel
-  const [fundingTxid] = channel_point.split(':')
+  const [fundingTxid] = channel_point ? channel_point.split(':') : []
 
   const data = [
     {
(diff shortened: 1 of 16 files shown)


app/components/Channels/ChannelData.js
```diff
@@ -24,7 +24,7 @@ const ChannelData = ({ channel, currencyName, networkInfo, viewMode, ...rest })
     csv_delay,
     num_updates,
   } = channel
-  const [fundingTxid] = channel_point.split(':')
+  const [fundingTxid] = channel_point ? channel_point.split(':') : []
```
korhaliv: When `channel_point ` is null or undef `fundingTxid` will be `undefined`. Is this ok ? I see that we use down below to open block explorer and show text.

mrfelton: Yes, that's fine because we filter out the null values here

https://github.com/LN-Zap/zap-desktop/blob/67ffbd5733b5c15837c351ce167159cd695d1019/app/components/Channels/ChannelData.js#L130

korhaliv: Ok makes sense. Ideally i'd use `Boolean(channel[id])` instead of `channel[id] != null` since latter implies implicit conversion undefined to null.

mrfelton: `Boolean(channel[id])` would filter out things where the value was `0`, which we don't want here. We only want to filter out things that are `null` or `undefined`

mrfelton: I've refactored this as per slack convo, and added some comments.

PR #1849, opened 2019-03-19
feat(lnd): lnd crash recovery (show dialog)

### LINKED ISSUE #1824: Handle if LND process exits
## Description

Zap takes no action if LND crashes.

## Expected Behavior

Zap should take one of these actions when LND crashes:

1. restart LND (newest versions of LND can continue to sync the channel graph if the wallet is locked)
2. alert the user (I believe there is an alert, but I believe it disappears after only a few seconds. the alert should persist until it is dismissed)
3. send the user back to the wallet selection/login screen

## Actual Behavior

LND crashes.  The app is now totally broken, but there is no indication that LND has crashed.

## Steps to Reproduce

Leave Zap open and wait for LND to spontaneously crash, or `kill` it.

## Your Environment

- Zap version: v0.4.0 58188fb184621d23bd25f912da62360c1b33a1f5 (running self-packaged build)
- Operating System: Ubuntu Linux

mrfelton:
## Description:

If lnd crashes show a dialog advising what happened. Require user interaction to acknowledge the crash event. Clean up and return the user to the launchpad after closing the dialog.

## Motivation and Context:

Fix #1824

## How Has This Been Tested?

Start local wallet and try killing lnd at various different points in the cycle, such as:

 - on the wallet unlocker screen
 - on the sync screen
 - when the wallet is fully open

## Screenshots:
image

## Types of changes:

Enhancement

## Checklist:


app/components/Dialog/DialogLndCrashed.js
```diff
@@ -0,0 +1,92 @@
...
+      <Heading.h1>
+        <FormattedMessage {...messages.lnd_crashed_dialog_header} />
+      </Heading.h1>
+    </Flex>
+  )
+
+  const { code, signal, error } = lndCrashReason
+  const body = (
+    <>
+      <Flex>
```
korhaliv: To me it looks like
 ```
<Flex>
        <Text fontWeight="normal" mr={2}>
          <FormattedMessage {...messages.lnd_crashed_dialog_signal} />
        </Text>
        <Text>
          {signal == null ? <FormattedMessage {...messages.lnd_crashed_dialog_unknown} /> : signal}
        </Text>
      </Flex>
```
can be extracted into sub-component safely. It's repeated 3 times, so i'd say it's totally worth doing. What do you think?

File: app/components/Dialog/messages.js [hunk 2]
```diff
@@ -0,0 +1,11 @@
+import { defineMessages } from 'react-intl'
+
+/* eslint-disable max-len */
+export default defineMessages({
+  lnd_crashed_dialog_header: 'Lnd has unexpectedly quit!',
+  lnd_crashed_dialog_code: 'Error code:',
+  lnd_crashed_dialog_signal: 'Error signal:',
+  lnd_crashed_dialog_messgage: 'Error message:',
+  lnd_crashed_dialog_unknown: 'unknown',
```
korhaliv: First letter should be capitalized for consistency

app/components/UI/Dialog.js
```diff
@@ -35,7 +35,7 @@ const Dialog = ({ header, onClose, buttons, width, children }) => {
   const buttonsLayout = React.isValidElement(buttons) ? buttons : createButtons(buttons)
 
   // check if header is a string or a React renderable node
-  const headerLayout = React.isValidElement(buttons) ? (
```
korhaliv: Oops

app/components/Dialog/messages.js
mrfelton: Updated

app/components/Dialog/DialogLndCrashed.js
mrfelton: Updated

Issue #1779, opened 2019-03-12
When using a lightning invoice payment uri for enabled autopay merchants, the invoice should be paid automatically without any further user interaction


mrfelton:
## Detailed Description

When using a lightning invoice payment uri for enabled autopay merchants, the invoice should be paid automatically without any further user interaction

An initial implementation should focus on the case where the user already has the app open and is logged into a wallet.

## Context

Autopay

mrfelton:
in the case of an automatic payment, what should we set the fee limit as?

cc @JimmyMow

mrfelton:
Closed via https://github.com/LN-Zap/zap-desktop/pull/1890

PR #1890, opened 2019-03-25
feat(autopay): auto pay invoices based on autopay config

mrfelton:
## Description:

Auto pay invoices based on autopay config

## Motivation and Context:

Address #1779

## How Has This Been Tested?

- Try paying autopay and non-autopay invoices.
- Test with the app open, hidden, and closed

## Types of changes:

New feature

## Checklist:


mrfelton:
@JimmyMow @korhaliv This PR implements basic autopay support for the case where the user is already logged into one of their wallets.

I'd like some input on how we should handle the case where the user is not already logged into the wallet (in this PR, the behaviour is unchanged from before in that case)

There is a mockup in Zeplin for how this could work, but it's not clear to me from this exactly what the behaviour would be once the user clicks the `Got it` button in that case (assuming we go with something like this)

image

renderer/reducers/autopay.js
```diff
@@ -219,6 +222,28 @@ function setAutopayListFromArray(state, { list }) {
...
+ */
+export const showAutopayNotification = invoice => async (dispatch, getState) => {
+  const nodes = networkSelectors.nodes(getState())
+
+  const descriptionTag = invoice.tags.find(tag => tag.tagName === 'description') || {}
+  const node = nodes.find(n => n.pub_key === invoice.payeeNodeKey)
+  const nodeName = node ? getNodeDisplayName(node) : truncateNodePubkey(invoice.payeeNodeKey)
+
+  const title = `Autopay invoice for ${invoice.satoshis} satoshis`
+  let body = `Paying ${invoice.satoshis} satoshis to ${nodeName}`
```
korhaliv: add intl

renderer/reducers/pay.js
```diff
@@ -46,8 +52,33 @@ export const SET_REDIRECT_PAY_REQ = 'SET_REDIRECT_PAY_REQ'
...
+      const invoice = decodePayReq(address)
+      const { payeeNodeKey, satoshis, millisatoshis } = invoice
+      const amountInSats = satoshis || convert('msats', 'sats', millisatoshis)
+      const autopayEntry = autopayList[payeeNodeKey]
+
+      // If there autopay is enabled for the node pubkey we got from the invoice and the amount of the invoice is less
+      // than the autopay's configured limit, pay the invoice silently in the background.
+      if (autopayEntry && amountInSats <= autopayEntry.limit) {
+        dispatch(showAutopayNotification(invoice))
+        return dispatch(payInvoice({ payReq: address, amt: amountInSats }))
```
korhaliv: This needs consistent return

renderer/reducers/pay.js
```diff
@@ -46,8 +52,33 @@ export const SET_REDIRECT_PAY_REQ = 'SET_REDIRECT_PAY_REQ'
...
+      }
+    }
+
+    // If it wasn't handled with autopay, open in the pay form and focus the app.
+    dispatch(setRedirectPayReq({ address }))
+    dispatch(send('showMainWindow'))
```
korhaliv: consider using finally since this is used both in normal and catch routes

electron/main.js
```diff
@@ -76,7 +76,8 @@ const handleBitcoinLink = input => {
 const handleLightningLink = input => {
   const address = input.split(':')[1]
   zap.sendMessage('lightningPaymentUri', { address })
-  mainWindow.show()
+  // Hide the window. The app will bring itself into focus if/when needed.
+  mainWindow.hide()
```
korhaliv: This will also alter regular, non autopay behavior while `bitcoin:` links handling stays the same

renderer/reducers/pay.js
```diff
@@ -46,8 +52,33 @@ export const SET_REDIRECT_PAY_REQ = 'SET_REDIRECT_PAY_REQ'
...
+
+    // If the user is logged into a wallet, try to autopay the invoice.
+    if (walletSelectors.isWalletOpen(state)) {
+      const autopayList = autopaySelectors.autopayList(state)
+      const invoice = decodePayReq(address)
+      const { payeeNodeKey, satoshis, millisatoshis } = invoice
+      const amountInSats = satoshis || convert('msats', 'sats', millisatoshis)
+      const autopayEntry = autopayList[payeeNodeKey]
+
+      // If there autopay is enabled for the node pubkey we got from the invoice and the amount of the invoice is less
```
korhaliv: `If there is autopay enabled` maybe?

renderer/reducers/pay.js
mrfelton: This is only used in the normal route if not paying with autopay, since the methods aborts after paying with an early return.

electron/main.js
mrfelton: That's intended. There should be no change to `bitcoin` links. But for `lightning` links we want to let the app decide wether or not to focus the window based on wether its autopay or not, which happens in the `lightningPaymentUri` method.

renderer/reducers/pay.js
mrfelton: I reworked this to make it a little easier to follow

Issue #1997, opened 2019-04-10
Missing host string after creating new wallet connection with Node Launcher


mrfelton:
## Description

Missing host string after creating new wallet connection with Node Launcher

## Expected Behavior

Host should be set properly in the wallet config.

## Actual Behavior

Host value is empty.

## Steps to Reproduce

1. Launch the node launcher bitcoin and lnd nodes
2. Click the `Configure Zap Desktop` button to launch Zap
3. Confirm host as `127.0.0.1`
4. Log out of the wallet
5. Notice that the Host is blank.
6. Attempt to enter a host value and save
7. App crashes with `Uncaught TypeError: Cannot read property 'split' of undefined at split (webpack:/renderer/components/Util/WalletName.js:11)`

![image](https://user-images.githubusercontent.com/200251/55892782-410ca000-5bb7-11e9-9306-76370167e0f3.png)

## Context

lndconnect

## Your Environment

- Zap version: 0.4.0
- Operating System and version: Mac / Windows

mrfelton:
This looks to be because Node Launcher are using old style lndconnect links, eg:

```
lndconnect:?cert=~/.lnd/tls.cert&macaroon=~/.lnd/admin.macaroon&host=example.com:10009
```

New style links work correctly:

```
lndconnect://example.com:10009?cert=~/.lnd/tls.cert&macaroon=~/.lnd/admin.macaroon
```

JimmyMow:
@mrfelton let's create an issue within the Node Launcher repo and close this?

mrfelton:
@JimmyMow I put a fix in place for this: https://github.com/LN-Zap/zap-desktop/pull/2000

Also created issue for Node Launcher to update their link format: https://github.com/lightning-power-users/node-launcher/issues/272

Once other thing we should probably do is a migration script to delete any wallet configs with missing host info. I'll do another PR for that.

PR #2003, opened 2019-04-10
fix(launchpad): fallback wallet name incase host unknown

mrfelton:
## Description:

Show a fallback value in case we are unable to determine the wallet name, rather than crashing the app.

This PR sets the wallet name as `[unnamed]` in the case that we can't determine the wallet name.

## Motivation and Context:

In some situations (eg #1997) it's possible that a corrupt wallet config gets saved such that we are unable to determine the wallet name correctly. In this case, the app crashes as it assumes that a wallet config has a value for `host`.

## How Has This Been Tested?

Put an invalid wallet config into the db (one without a host prop) and verify that the app doesn't crash when trying to access the wallet's settings form.

image

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/renderer/components/Util/WalletName.js
+++ b/renderer/components/Util/WalletName.js
@@ -4,11 +4,18 @@ const WalletName = ({ wallet }) => {
   if (!wallet) {
     return null
   }
+  const { name, host, type, id } = wallet
 
-  if (wallet.type === 'local') {
-    return wallet.name || `Wallet #${wallet.id}`
+  // For local wallets, display the wallet name if set.
+  // Otherwise fallback to the wallet id.
+  if (type === 'local') {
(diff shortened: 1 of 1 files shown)


renderer/components/Util/WalletName.js
```diff
@@ -8,7 +8,8 @@ const WalletName = ({ wallet }) => {
   if (wallet.type === 'local') {
     return wallet.name || `Wallet #${wallet.id}`
   }
-  return wallet.name || wallet.host.split(':')[0]
+  const { name, host } = wallet
```
korhaliv: Could you please move destructuring up, since `wallet.name` is used 2 lines above ?

mrfelton: Updated

PR #2054, opened 2019-04-17
fix(ui): autofocus amount fields

### LINKED ISSUE #2046: Autofocus amount on channel form
When opening a channel, after I click a suggested node or enter a `pubkey@host` I am taken to the form to enter an amount. I think Zap should autofocus to the amount input as that's the only required field needed to submit the form

Screen Shot 2019-04-16 at 9 53 23 AM

mrfelton:
## Description:

Autofocus amount fields

## Motivation and Context:

Fix #2046

## How Has This Been Tested?

Ensure amount field autofocuses on channel form.

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/renderer/components/Channels/ChannelCreateForm.js
+++ b/renderer/components/Channels/ChannelCreateForm.js
@@ -33,9 +33,13 @@ import {
  * Animation to handle showing/hiding the amount fields.
  */
 const ShowHide = Keyframes.Spring({
-  show: async next => {
+  show: async (next, cancel, ownProps) => {
     await next({ pointerEvents: 'auto' })
     await next({ opacity: 1 })
+    // Focus the amount input if the current step is the form step.
+    if (ownProps.context.state.step === 'form') {
+      ownProps.context.focusAmountInput()
+    }
   },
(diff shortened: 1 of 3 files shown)


renderer/components/Channels/ChannelCreateForm.js
```diff
@@ -33,9 +33,13 @@ import {
  * Animation to handle showing/hiding the amount fields.
  */
 const ShowHide = Keyframes.Spring({
-  show: async next => {
+  show: async (next, cancel, ownProps) => {
     await next({ pointerEvents: 'auto' })
     await next({ opacity: 1 })
+    // Focus the amount input if the current step is the form step.
+    if (ownProps.context.state.step === 'form') {
+      ownProps.context.focusAmountInput()
```
korhaliv: Do you think `Spring keyframes` is an appropriate place for this? Maybe something like `cdu` would be better ?

mrfelton: The issue is that we can only focus it once it has been rendered, but it only gets rendered after the show animation has completed. This is the only place that we get notified of the animation having completed. ComponentDidUpdate does not get called.

PR #2028, opened 2019-04-13
refactor: rework neutrino / lnd / grpc handling to use Web Workers

mrfelton:
## Description:

Rework neutrino / lnd / grpc handling to use Web Workers

## Motivation and Context:

Clean up codebase, making it easier to follow
Improve performance by removing all potentially blocking code from the main process

## How Has This Been Tested?

Manually

## Types of changes:

Refactor

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have reviewed and updated the documentation accordingly.
- [x] I have read the _CONTRIBUTING_ document.
- [x] I have added tests to cover my changes where needed.
- [x] All new and existing tests passed.
- [x] My commits have been squashed into a concise set of changes.

PR #2055, opened 2019-04-17
refactor: combine grpc services

mrfelton:
## Description:

Builds on #2028 

Combines `WalletUnlocker` and `Lightning` services into a single `grpc` service.

## Motivation and Context:

Simplify code, consolidate multiple state machines, simplify connection logic

## How Has This Been Tested?

Manually - try connecting to, disconnecting from, unlocking, logging out of lots of wallets

## Types of changes:

Refactor

## Checklist:

### DIFF
--- a/renderer/reducers/channels.js
+++ b/renderer/reducers/channels.js
@@ -2,7 +2,7 @@ import { createSelector } from 'reselect'
 import throttle from 'lodash.throttle'
 import { proxyValue } from 'comlinkjs'
 import { requestSuggestedNodes } from '@zap/utils/api'
-import { lightningService } from 'workers'
+import { grpcService } from 'workers'
 import { updateNotification, showWarning, showError } from './notification'
 import { fetchBalance } from './balance'
 import { walletSelectors } from './wallet'
@@ -272,8 +272,8 @@ export const fetchSuggestedNodes = () => async dispatch => {
 // Send IPC event for peers
 export const fetchChannels = () => async dispatch => {
   dispatch(getChannels())
-  const lightning = await lightningService
-  const channels = await lightning.getChannels()
+  const grpc = await grpcService
+  const channels = await grpc.services.Lightning.getChannels()
   dispatch(receiveChannels(channels))
 }
 
@@ -308,9 +308,9 @@ export const openChannel = data => async (dispatch, getState) => {
   dispatch(showWarning('Channel opening initiated', { payload: { pubkey }, isProcessing: true }))
 
   // Attempt to open the channel.
-  const lightning = await lightningService
   try {
-    const data = await lightning.connectAndOpen({
+    const grpc = await grpcService
+    const data = await grpc.services.Lightning.connectAndOpen({
       pubkey,
       host,
       localamt,
@@ -319,7 +319,10 @@ export const openChannel = data => async (dispatch, getState) => {
       spendUnconfirmed,
     })
     dispatch(pushchannelupdated(data))
-    lightning.once('openChannel.data', proxyValue(data => dispatch(pushchannelupdated(data))))
+    grpc.services.Lightning.once(
+      'openChannel.data',
+      proxyValue(data => dispatch(pushchannelupdated(data)))
+    )
   } catch (e) {
     dispatch(
       pushchannelerror({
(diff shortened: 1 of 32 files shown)


renderer/reducers/app.js
```diff
@@ -63,14 +63,10 @@ export const initApp = () => async (dispatch, getState) => {
...
-  const walletUnlocker = await walletUnlockerService
-  if (await walletUnlocker.can('disconnect')) {
-    await walletUnlocker.disconnect()
-  }
-  const lightning = await lightningService
-  if (await lightning.can('disconnect')) {
-    await lightning.disconnect()
+  // Disconnect gRPC services.
+  const grpc = await grpcService
+  if (await await grpc.can('disconnect')) {
```
korhaliv: wait wait wait, are you saying we need all those awaits?

renderer/reducers/channels.js
```diff
@@ -308,9 +308,9 @@ export const openChannel = data => async (dispatch, getState) => {
   dispatch(showWarning('Channel opening initiated', { payload: { pubkey }, isProcessing: true }))
 
   // Attempt to open the channel.
-  const lightning = await lightningService
+  const grpc = await grpcService
```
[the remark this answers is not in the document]
mrfelton: Updated

renderer/reducers/app.js
mrfelton: lol. Updated

renderer/reducers/channels.js
```diff
@@ -378,9 +381,9 @@ export const closeChannel = () => async (dispatch, getState) => {
     const [funding_txid, output_index] = channel_point.split(':')
 
     // Attempt to open the channel.
-    const lightning = await lightningService
+    const grpc = await grpcService
```
korhaliv: can we move this inside try?

renderer/reducers/neutrino.js
```diff
@@ -108,17 +108,18 @@ export const startNeutrino = lndConfig => async dispatch => {
     // Initialise the Neutrino Web Worker..
     const neutrino = await neutrinoService
     await neutrino.init(lndConfig)
+    dispatch(initNeutrino())
 
     // Start the service and wait for one of the gRPC interfaces to become active.
     await new Promise(async resolve => {
-      neutrino.once(
+      neutrino.on(
```
korhaliv: Do we ensure this is cleared to prevent double subs?

renderer/reducers/peers.js
```diff
@@ -27,8 +27,8 @@ export function getPeers() {
 // Send IPC event for peers
 export const fetchPeers = () => async dispatch => {
   dispatch(getPeers())
-  const lightning = await lightningService
-  const peers = await lightning.listPeers()
+  const grpc = await grpcService
```
korhaliv: In other similar actions, worker calls are wrapper with `try catch` blocks

renderer/reducers/transaction.js
```diff
@@ -53,8 +53,8 @@ export function sendTransaction(data) {
 // Send IPC event for payments
 export const fetchTransactions = () => async dispatch => {
   dispatch(getTransactions())
-  const lightning = await lightningService
-  const transactions = await lightning.getTransactions()
+  const grpc = await grpcService
```
korhaliv: Same here

renderer/reducers/neutrino.js
mrfelton: I've added calls to remove these listeners in the `NEUTRINO_EXIT` event handler.

renderer/reducers/peers.js
mrfelton: As per slack, these should be wrapped with failure cases, but since we don't already have failure cases for these actions I don't want to add to the scope of this PR by introducing them here.

renderer/reducers/transaction.js
mrfelton: as above

services/grpc/grpc.js
```diff
@@ -0,0 +1,230 @@
...
+        case status.UNKNOWN:
+          return WALLET_STATE_LOCKED
+
+        /**
+          Bubble all other errors back to the caller and abort the connectiopn attempt.
+          Disconnect all services.
+        */
+        default:
+          grpcLog.warn('Unable to connect to lnd service', e)
+          await this.disconnectAll()
```
korhaliv: This theoretically could cause exception inside this catch, and `disconnectAll` doesn't handle exceptions itself, so this will result in an unhandled exception

PR #2062, opened 2019-04-18
feat(wallet): ability to log out from loader

### LINKED ISSUE #1989: Ability to cancel wallet connection attempt
When ZAP wallet is setup and connecting (or with wrong settings: not connecting); then there is no ability to change the connections settings; the only solution is to fully uninstall and manually remove data folders (which are left after uninstall).

## Detailed Description
Would be great to have a simple ability to change the connection settings on an already installed ZAP wallet.

mrfelton:
## Description:

Sometimes it happens that a wallet login is successful, but the wallet ui doesn't show due to missing dependencies such as ticker information or required information about the lnd node that we connected to. Provide a means to cancel the failed login attempt and return to the launchpad.

Depends on #1989 

## Motivation and Context:

Fix #1989

## How Has This Been Tested?

Manually.

## Screenshots:

(note X in upper right corner - this will trigger a logout)

image

## Types of changes:

Enhancement

## Checklist:


renderer/components/LoadingBolt/LoadingBolt.js
```diff
@@ -48,43 +51,79 @@ const FullPageGradient = styled(Flex)`
...
+const LoadingCloseButtonWrapper = styled(Box)`
+  height: 40px;
+  cursor: pointer;
+  opacity: 0.6;
+  &:hover: {
+    opacity: 1;
+  }
+`
+
+const LoadingCloseButton = ({ onClose }) => (
```
korhaliv: We have close button in a couple of places already (for instance in AutopayModalBody). Would be nice to finally extract this and avoid duplicating them.

renderer/components/LoadingBolt/LoadingBolt.js
```diff
@@ -48,43 +51,79 @@ const FullPageGradient = styled(Flex)`
...
+LoadingCloseButton.propTypes = {
+  onClose: PropTypes.func,
+}
+
 class LoadingBolt extends React.PureComponent {
   static propTypes = {
+    hasClose: PropTypes.bool,
     isLoading: PropTypes.bool.isRequired,
     message: PropTypes.object,
+    onClose: PropTypes.func,
```
korhaliv: Looks like this prop is not required, however it is directly passed into an onClick without a null check

renderer/reducers/app.js
```diff
@@ -49,6 +52,14 @@ export function resetApp() {
   }
 }
 
+export const logout = () => async dispatch => {
```
korhaliv: `await` is not used anywhere inside, nor does this function return a promise. consider removing `async` modifier

renderer/components/LoadingBolt/LoadingBolt.js
mrfelton: Have extracted this into a new `CloseButton` component.

renderer/components/LoadingBolt/LoadingBolt.js
mrfelton: Added null handling in `CloseButton`

renderer/reducers/app.js
mrfelton: Updated

mrfelton:
Superseded by https://github.com/LN-Zap/zap-desktop/pull/2069

PR #2066, opened 2019-04-18
perf(channels): optimize list rendering

korhaliv:
## Description:
Improve channels virtualized list rendering by reducing overscan for the card view, removing dynamic objects (css style) and introducing deep compare for the props. We have a complex `channel` object which is produced by a selector and is additionally decorated. This was causing redundant re-renders of the channel list items. Overscan count was reduced only  for the card view because are a lot larger than summary items.

## How Has This Been Tested?
Manually

## Types of changes:
Perf improvement

## Checklist:

### DIFF
--- a/renderer/components/Channels/ChannelCardListItem.js
+++ b/renderer/components/Channels/ChannelCardListItem.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import isEqual from 'lodash.isequal'
 import { FormattedMessage, injectIntl, intlShape } from 'react-intl'
 import styled from 'styled-components'
 import { opacity } from 'styled-system'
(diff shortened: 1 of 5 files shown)


renderer/components/Channels/ChannelCardListItem.js
```diff
@@ -29,8 +32,9 @@ const ChannelCardListItem = React.memo(
       active,
     } = channel
     const opacity = active ? 1 : 0.3
+
     return (
-      <Card {...rest}>
+      <Card>
```
mrfelton: Why remove `...rest` from here? This makes the component less configurable by the caller

korhaliv: I've removed it until it's actually needed. We can't blindly pass anything there because previously it was causing superfluous renders. This is a narrowly scoped component with perf focused code

PR #2086, opened 2019-04-23
fix(i18n): add intl support for selects

### LINKED ISSUE #2073: Filter dropdowns should be translatable
## Detailed Description

Filter dropdowns are currently not translatable as the strings are added in the reducers. They should be made translatable.

korhaliv:
## Description:
Resolves #2073 

## How Has This Been Tested?
Manually

## Types of changes:
Refactor/bugfix

## Checklist:


renderer/components/UI/Select.js
```diff
@@ -70,40 +70,45 @@ const StyledInput = styled(BasicInput)`
...
+function useIntl(items, messageMapper, intl) {
+  return useMemo(() => {
+    // if messageMapper is not set just original items array
+    // in this case items should already contain value prop for each key
+    if (!messageMapper) {
+      return items
     }
-  }
+    return items.map(({ key }) => {
+      return { key, value: intl.formatMessage({ ...messageMapper(key) }) }
```
mrfelton: It might be worth adding a check here to verify that `.messageMapper(key)` returns something and if not just return the item key as the value.

renderer/components/UI/Select.js
```diff
@@ -70,40 +70,45 @@ const StyledInput = styled(BasicInput)`
...
-    items: [],
-    iconSize: 8,
-  }
-
-  inputRef = React.createRef()
-
-  blurInput = () => {
-    if (this.inputRef.current) {
-      this.inputRef.current.blur()
+function useIntl(items, messageMapper, intl) {
```
mrfelton: Maybe move this to the `hooks` dir?

renderer/components/UI/Select.js
```diff
@@ -70,40 +70,45 @@ const StyledInput = styled(BasicInput)`
...
+    })
+  }, [items, messageMapper, intl])
+}
 
-  renderSelectOptions = (highlightedIndex, selectedItem, getItemProps) => {
-    let { items, theme } = this.props
+function Select(props) {
+  const { intl, messageMapper } = props
+  const inputRef = useRef(null)
+  // eslint-disable-next-line react/destructuring-assignment
```
mrfelton: Can you pull `items` out of props in line 104 to avoid this?

korhaliv: the idea was to override items prop so later there won't be a mistake because props.items will be used

renderer/components/UI/Select.js
korhaliv: This has pros and cons. Currently, mapper works as a "compile time" checker. so if the message is missing it will throw an error which would be easily detectable. Silently falling back to a key is ok, but may bring more problems than improvements. What do you think?

PR #2088, opened 2019-04-23
feat(onboarding): ability to import encrypted seed

### LINKED ISSUE #2071: I should be able to import an encrypted seed (recovery process hangs at 'Importing wallet' screen)
Have build a fresh BTC-LND and LTC-LND setup and successfully created new wallets with ZAP for both chains; now adding a 3rd wallet with using the recovery seed from a older (BTC only) setup/installation (thus different node LNDdb and bitcoindDB)...
while doing so, the wallet hangs on the msg "Importing wallet..."

when I use the cancel/return function...
I suddenly am seeing 2 extra lines in my ZAPwallet:
Under "Your Wallets" I see the imported (but failed) seed shows twice named
and the 2 wallets which I had initially setup moved down to "more"...

When selecting the added CoDaInLND entry, then I am asked for the password... which is not accepted (I am sure that I am entering it correctly ;-)
I am wondering if what I attempt can be done... importing a wallet by seed from a different installation; i hope so, as I assume it is loaded/retrieved from mainnet and not local ;-)

Using the existing wallets (BTC-LND and LTC-LND) still works.

Environment:
- Zap version: 0.4.1
- Operating System and version: WindowsServer2016

ZAPwallet

mrfelton:
## Description:

Provide users the ability to import a seed.

If the import fails with an error, check the error message to determine the reason for the failure. If the reason was that an invalid passphrase was provided, present the user with a password field where they can enter the cipher seed passphrase.

This does not change the standard import flow. It adds a new step to the end of the process only if we detect that the import failed due to a missing passphrase. 

## Motivation and Context:

Fix #2071

## How Has This Been Tested?

- test import process with encrypted seed
 - and with unencrypted seed

## Screenshots:

**initial submit:**
image

**password prompt (shown after faire to import due to missing seed):**
image

**incorrect password (shown after the user submits the password incorrectly:**
image

## Types of changes:

New feature

## Checklist:

- [x] My commits have been squashed into a concise set of changes


renderer/components/Onboarding/Steps/WalletRecover.js
```diff
@@ -20,18 +28,67 @@ class WalletRecover extends React.Component {
...
+
+    // Handle success case.
+    if (!walletRecoveryError && !isRecoveringWallet && prevProps.isRecoveringWallet) {
+      this.handleSuccess()
+    }
+
+    // Handle failure case.
+    if (walletRecoveryError && !isRecoveringWallet && prevProps.isRecoveringWallet) {
+      this.handleError()
+    }
```
korhaliv: Would be cleaner to code it like this imo:
```

    if( !isRecoveringWallet && prevProps.isRecoveringWallet){
        if(walletRecoveryError){
            this.handleError()
        }
        else {
            this.handleSuccess()
        }
    }
```

renderer/reducers/lnd.js
```diff
@@ -409,6 +406,10 @@ export const createNewWalletFailure = error => ({
  */
 export const recoverOldWallet = () => async (dispatch, getState) => {
   dispatch({ type: RECOVER_OLD_WALLET })
+
+  const grpc = await grpcService
```
korhaliv: Can you move this inside try please?

renderer/reducers/lnd.js
```diff
@@ -429,16 +430,40 @@ export const recoverOldWallet = () => async (dispatch, getState) => {
     // Start lnd with the provided wallet config.
     await dispatch(startLnd(wallet))
 
+    // Set up a listener that resolves once the lightning interface has become active.
+    // This is the point where we really know that the wallet has been created and we can connect to it.
+    const waitForLightning = new Promise(resolve => {
+      handleLightningActive = proxyValue(() => resolve())
+      grpc.prependListener('GRPC_LIGHTNING_SERVICE_ACTIVE', handleLightningActive)
+    })
```
korhaliv: This looks like a fairly low-level routine and also like a reusable one. Consider extracting this into util.

renderer/reducers/wallet.js
```diff
@@ -61,6 +61,29 @@ export const putWallet = wallet => async dispatch => {
...
+
+  // Delete the wallet from the database.
+  await window.db.wallets.delete(wallet.id)
+
+  // Refresh the wallets state data.
+  await dispatch(getWallets())
+}
+
+/**
+ * Handle wallet delete action.
```
korhaliv: Let's outline that `deleteWallet` is a response to a user input type of action, so it's clear what is the distinction between this and `removeWallet` 
Something like `Handle delete wallet confirmation by a user`

services/grpc/walletUnlocker/walletUnlocker.methods.js
```diff
@@ -27,19 +27,24 @@ async function unlockWallet(password) {
 
 /**
  * InitWallet is used when lnd is starting up for the first time to fully initialize the daemon and its internal wallet
- * @param  {String} wallet_password        wallet_password is the passphrase that should be used to encrypt the wallet
+ * @param  {String} wallet_password        Uused to encrypt the wallet
```
korhaliv: comment typo

renderer/reducers/lnd.js
```diff
@@ -429,16 +430,40 @@ export const recoverOldWallet = () => async (dispatch, getState) => {
...
-    dispatch(recoverOldWalletFailure(error))
+    // Attempt to clean up from the failed import attempt.
+    try {
+      const { lndConfig } = getState().lnd
+      await dispatch(stopLnd())
+      await dispatch(removeWallet(lndConfig))
+    } finally {
+      // Remove Lightning grpc activation listener.
+      if (handleLightningActive) {
+        grpc.off('GRPC_LIGHTNING_SERVICE_ACTIVE', handleLightningActive)
```
korhaliv: To me it looks like `GRPC_LIGHTNING_SERVICE_ACTIVE` listener will cleanup only if there is an error in the main code, otherwise this `finally` is unreachable

renderer/components/Onboarding/Steps/WalletRecover.js
mrfelton: This has been refactored

renderer/reducers/lnd.js
mrfelton: Updated

renderer/reducers/lnd.js
mrfelton: I'e extracted this out into a helper

renderer/reducers/wallet.js
mrfelton: Updated

services/grpc/walletUnlocker/walletUnlocker.methods.js
mrfelton: Fixed

renderer/reducers/lnd.js
mrfelton: This has been refactored, I now use `prependOnceListener` to ensure the callback will only be called once, so if the code block above is successful, there will be nothing to remove so it's fine that this finally block never runs.

mrfelton:
Closed via https://github.com/LN-Zap/zap-desktop/pull/2101

PR #2097, opened 2019-04-23
fix(onboarding): prevent duplicate wallet entries

mrfelton:
## Description:

Lock the create/import wallet form submit button whilst the wallet is in the process of being created to prevent duplicate submissions.    

NOTE: This builds on the work in #2088

## Motivation and Context:

Fix #2087

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix / enhancement

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have reviewed and updated the documentation accordingly.
- [x] I have read the _CONTRIBUTING_ document.
- [ ] I have added tests to cover my changes where needed.
- [ ] All new and existing tests passed.
- [x] My commits have been squashed into a concise set of changes.

PR #2101, opened 2019-04-23
refactor: combine create/import wallet actions

mrfelton:
## Description:

This PR combines #2088 and #2097 and then refactors the code to reduce duplication

## How Has This Been Tested?

See above 2 PR notes

## Types of changes:

Refactor / enhancement / feature

## Checklist:


renderer/reducers/lnd.js
```diff
@@ -50,6 +47,46 @@ export const DISCONNECT_GRPC = 'DISCONNECT_GRPC'
...
+ * Helper methods to allow waiting for the Lightning interface to become active.
+ * @return {Promise<Object>} Object with `isActive` and `cancel` properties.
+ */
+export const getGrpcHelpers = async () => {
+  let successHandler
+  const grpc = await grpcService
+
+  return {
+    /**
+     * Wait for lightning to becomee active.
```
korhaliv: comment typo

renderer/reducers/lnd.js
```diff
@@ -352,10 +385,16 @@ export const fetchSeedError = error => dispatch => {
...
+ * @param  {Object} options Options
+ * @param  {Object} options.recover Boolean indicating weather this is a recovery
+ * @return {Promise}
  */
-export const createNewWallet = () => async (dispatch, getState) => {
-  dispatch({ type: CREATE_NEW_WALLET })
+export const createWallet = ({ recover } = {}) => async (dispatch, getState) => {
+  dispatch({ type: CREATE_WALLET })
+
+  const grpcHelpers = await getGrpcHelpers()
```
korhaliv: can we move this inside try?

renderer/reducers/lnd.js
```diff
@@ -380,81 +419,51 @@ export const createNewWallet = () => async (dispatch, getState) => {
     const grpc = await grpcService
     await grpc.services.WalletUnlocker.initWallet({
       wallet_password: state.onboarding.password,
+      aezeed_passphrase: state.onboarding.passphrase,
       cipher_seed_mnemonic: state.onboarding.seed,
-      recovery_window: 0,
+      recovery_window: recover ? 2500 : 0,
```
korhaliv: Should this be made configurable?

renderer/reducers/lnd.js
mrfelton: Updated

renderer/reducers/lnd.js
mrfelton: No, because it's used in the catch block

renderer/reducers/lnd.js
mrfelton: extracted to the config file

services/grpc/grpc.js
```diff
@@ -205,6 +207,45 @@ class GrpcService extends EventEmitter {
...
+  waitForService(serviceName) {
+    let successHandler
+    const activationEventName = `GRPC_${snakecase(serviceName).toUpperCase()}_SERVICE_ACTIVE`
+
+    /**
+     * Promise that resolves when service is active.
+     */
+    const isActive = new Promise(async resolve => {
+      // If the service is already active, return immediatly.
+      if (this.services[serviceName].is('connected')) {
```
korhaliv: if we are requesting to wait for  a service that was not previously registered (and thus `this.services[serviceName]`) this will throw. is it expected?

services/grpc/grpc.js
```diff
@@ -205,6 +207,45 @@ class GrpcService extends EventEmitter {
...
+   */
+  waitForService(serviceName) {
+    let successHandler
+    const activationEventName = `GRPC_${snakecase(serviceName).toUpperCase()}_SERVICE_ACTIVE`
+
+    /**
+     * Promise that resolves when service is active.
+     */
+    const isActive = new Promise(async resolve => {
+      // If the service is already active, return immediatly.
```
korhaliv: comment typo `immediatly`

services/grpc/grpc.js
```diff
@@ -205,6 +207,45 @@ class GrpcService extends EventEmitter {
...
+      if (this.services[serviceName].is('connected')) {
+        return resolve()
+      }
+      // Otherwise, wait until we receive an activation event from the gRPC service.
+      successHandler = () => resolve()
```
korhaliv: To me it looks like if `isActive` is called multiple times `successHandler` will get overridden and event listener won't get cleared. Should we clear previous listener here?

services/grpc/grpc.js
mrfelton: I think it probably should throw in this case. The alternative is that we resolve immediately in the case that the caller asks about an unknown service, but that seems wrong since the service isn't active.

Services get registered in the `init` method, so if the user was asking about an unregistered service it would imply that they had either not initialised the services first, or that they are simply asking about an unknown service - in either case, this is a user error and we should probably let it throw.

services/grpc/grpc.js
mrfelton: Fixed

services/grpc/grpc.js
korhaliv: Ok, let's leave it then

services/grpc/grpc.js
mrfelton: `IsActive` is not a method. It is a promise. If the user does `await isActive` multiple times, they are waiting on the same promise, so no new listeners are being setup.

korhaliv: Right, my bad, was looking at `cancel` and automatically implied `isActive` is also a method

PR #2104, opened 2019-04-24
fix(deps): update lnd to v0.6.0-beta-31-g43f1290e

mrfelton:
## Description:

 - update lnd to v0.6.0-beta-31-g43f1290e
 - clean up disconnected streams
 - prevent duplicate lnd stop actions
 - add 2 second delay before connecting to grpc streams

## Motivation and Context:

Get latest neutrino updates.

## How Has This Been Tested?

Manually

## Checklist:

### DIFF
--- a/services/grpc/grpcService.js
+++ b/services/grpc/grpcService.js
@@ -145,6 +145,17 @@ class GrpcService extends EventEmitter {
    */
   subscribe() {
     // this.subscriptions['something'] = this.service.subscribeToSomething()
+    // super.subscribe()
+
+    // Close and clear subscriptions when they emit an end event.
+    const activeSubKeys = Object.keys(this.subscriptions)
+    activeSubKeys.forEach(key => {
+      const call = this.subscriptions[key]
+      call.on('end', () => {
+        grpcLog.info(`gRPC subscription "${this.serviceName}.${key}" ended.`)
+        delete this.subscriptions[key]
+      })
+    })
   }
 
   /**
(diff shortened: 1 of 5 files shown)


services/grpc/grpcService.js
```diff
@@ -77,8 +77,15 @@ class GrpcService extends EventEmitter {
   /**
    * Log successful connection.
    */
-  async onAfterConnect() {
-    await this.subscribe()
+  onAfterConnect() {
+    // Wait for 2 seconds before subscribing to grpc streams.
+    // This gives lnd a chance to start up it's services after unlocking a wallet.
+    // We add this delay as in recent versions fo lnd (post 0.6.0), the channel stream is not functional immediately.
+    setTimeout(() => {
```
korhaliv: can we use `await delay(2000)` here?

services/grpc/grpcService.js
```diff
@@ -145,32 +152,42 @@ class GrpcService extends EventEmitter {
    */
   subscribe() {
     // this.subscriptions['something'] = this.service.subscribeToSomething()
+    // super.subscribe()
+
+    // Close and clear subscriptions when they emit an end event.
+    const activeSubKeys = Object.keys(this.subscriptions)
+    activeSubKeys.map(key => {
```
korhaliv: While using `map` here works, semantically using `forEach` makes more sense here IMO

services/grpc/grpcService.js
mrfelton: The reason I didn't use await delay is because I want this to run in the background. If we await, it will hold up the progress of the state transition.

services/grpc/grpcService.js
mrfelton: Updated

services/grpc/grpcService.js
korhaliv: Fair enough. In that case, just for the sake of not using `setTimeout` and callbacks, how do you feel about this code then 
```
function onAfterConnect() {
     // non blocking delay
    (async function () {
        await delay(2000)
        if (this.is('connected')) {
            this.subscribe()
          }
    })()
    //do stuff, no block 
}
```

mrfelton: I'v adjusted this

services/grpc/grpcService.js
```diff
@@ -145,32 +155,42 @@ class GrpcService extends EventEmitter {
    */
   subscribe() {
     // this.subscriptions['something'] = this.service.subscribeToSomething()
+    // super.subscribe()
```
korhaliv: Is this a leftover?

mrfelton: No, this is meant as a sample to show what subclasses should do in their implementation of this method - add stuff to `this.subscriptions` and call `super.subscribe`

PR #2121, opened 2019-04-26
fix(grpc): comlink callback error

korhaliv:
## Description:
Fixes comlink callback error
```
comlink.js?bb96:118 Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'MessagePort': [object Function] could not be cloned.
    at MessagePort.eval (webpack-internal:///./node_modules/comlinkjs/comlink.js:124:25)
(anonymous) @ comlink.js?bb96:118
async function (async)
(anonymous) @ comlink.js?bb96:85
debug.js?774b:114 zap:grpc 
```

## How Has This Been Tested?
Manually

## Types of changes:
Bugfix

## Checklist:

### DIFF
--- a/services/grpc/grpc.js
+++ b/services/grpc/grpc.js
@@ -33,6 +33,7 @@ class Grpc extends EventEmitter {
         onAfterActivateLightning: this.onAfterActivateLightning.bind(this),
         onBeforeDisconnect: this.onBeforeDisconnect.bind(this),
         onAfterDisconnect: this.onAfterDisconnect.bind(this),
+        onLeaveActive: this.onLeaveActive.bind(this),
       },
     })
     this.supportedServices = [WalletUnlocker, Lightning]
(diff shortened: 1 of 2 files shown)


services/grpc/grpc.js
```diff
@@ -141,12 +142,25 @@ class Grpc extends EventEmitter {
...
+    Lightning.on('subscribeTransactions.data', forwardEvent('subscribeTransactions.data'))
+    Lightning.on('subscribeChannelGraph.data', forwardEvent('subscribeChannelGraph.data'))
   }
+
   async onAfterActivateLightning() {
     this.emit('GRPC_LIGHTNING_SERVICE_ACTIVE')
   }
 
+  onLeaveActive() {
+    this.removeAllListeners('subscribeInvoices.data')
```
mrfelton: Shouldn't this be removing listeners for ` this.services.Lightning` that we added here, rather than removing listeners that users of the `Grpc` class have added?

korhaliv: I was thinking about that too. But didn't want to introduce instance variables to track event handlers. My reasoning was that particular Lightning variable is an internal instance and we fully control those listeners and should not be having more than one handler per event. If you think this is confusing i can refactor. Also in a future update (the discussed refactor of grpc) Lightning won't get exposed at all, so there will be no chance client code adds its own listeners.

mrfelton: There is no need to track individual handlers and remove one by one. You can just use `removeAllListeners` as you are already:

```
this.removeAllListeners('subscribeInvoices.data')
this.removeAllListeners('subscribeTransactions.data')
this.removeAllListeners('subscribeChannelGraph.data')
```
->
```
this.services.Lightning.removeAllListeners('subscribeInvoices.data')
this.services.Lightning.removeAllListeners('subscribeTransactions.data')
this.services.Lightning.removeAllListeners('subscribeChannelGraph.data')
```

I'm just suggesting that you remove the event forwarding listeners that you added to `this.services.Lightning` in `onBeforeActivateLightning`, rather than removing listeners to `this` that external users of the class added.

korhaliv: Yeah, you are 100% right. My bad, I've misread your original comment

PR #2134, opened 2019-04-29
perf(app): prevent superfluous top level re-renders

korhaliv:
## Description:
Prevent unnecessary `App` re-renders by extracting peer poll timeout from the `state` into a ref

## Motivation and Context:
`nextFetchIn` doesn't  influence anything in render func so can be safely extracted into a ref

## How Has This Been Tested?
Manually by monitoring redux log

## Types of changes:
Refactor/perf improvement

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have reviewed and updated the documentation accordingly.
- [x] I have read the _CONTRIBUTING_ document.
- [ ] I have added tests to cover my changes where needed.
- [ ] All new and existing tests passed.
- [x] My commits have been squashed into a concise set of changes.

PR #2138, opened 2019-04-30
fix(grpc): subscribe for topology update only after sync is done

### LINKED ISSUE #2128: Error subscribing to channel graph gRPC stream during sync process
## Description

When creating a new wallet, or opening an existing one where some remaining sync is needed when opening the wallet, there is the following error in the logs:
```
zap:grpc [ERR]  CHANNELGRAPH ERROR: Error: router not started
    at Http2CallStream.call.on (call.js?4dae:52)
    at Http2CallStream.emit (events.js:187)
    at Http2CallStream.endCall (call-stream.js?85fb:58)
    at handlingTrailers (call-stream.js?85fb:147)
    at process._tickCallback (internal/process/next_tick.js:68) +19ms
```

The channel graph subscription doesn't get set up as a result.

This is a new problem after updating to the latest lnd.

In [this commit](https://github.com/LN-Zap/zap-desktop/pull/2104/commits/21fc54c6f237357953f65ed4f686fdbf78dae843) some adjustments were made to delay the setting up of the subscriptions by 2 seconds after connecting to the api which resolves a similar problem when starting wallets that don't need to do any sync when opening.

The 2 second delay chosen there was completely arbitrary.

I'm not sure if there is a way to actually know when it is possible to connect to the stream.

## Expected Behavior

Should not produce error and should set up subscription once lnd is ready to accept subscriptions.

## Actual Behavior

Produces error and no active channel graph subscription

## Possible Fix

- Remove 2 second delay previously added.
- If there is an error when trying to set up a subscription, set up a retry attempt on a backoff schedule.

## Your Environment

- Zap version: next
- Operating System and version: all

korhaliv:
## Description:
Resolves #2128 
This PR builds on top of #2134 `scheduler` functionality. So ideally that one should be merged first. The idea of the PR is to create a poll service (streamify util) for grpc commands which allows outer code to work with them as with regular LND subscriptions.  To solve #2128 we need to subscribe to topology update only after sync process is complete. To achieve this  `getInfo` is streamified and channel graph stream is set only after `sycned_to_chain` is true  

## How Has This Been Tested?
Manually

## Types of changes:
Refactor/bugfix

## Checklist:

### DIFF
--- a/services/grpc/grpc.js
+++ b/services/grpc/grpc.js
@@ -144,11 +144,22 @@ class Grpc extends EventEmitter {
   async onBeforeActivateLightning() {
     const { Lightning } = this.services
     await Lightning.connect()
-    // creates listener that re-emits specified event
-    const forwardEvent = event => data => this.emit(event, data)
-    Lightning.on('subscribeInvoices.data', forwardEvent('subscribeInvoices.data'))
-    Lightning.on('subscribeTransactions.data', forwardEvent('subscribeTransactions.data'))
-    Lightning.on('subscribeChannelGraph.data', forwardEvent('subscribeChannelGraph.data'))
+
+    // setups listener that re-emits specified event
+    const forwardEvent = event => {
+      Lightning.on(event, data => this.emit(event, data))
+    }
+
+    // forwards `data` and `error` events of the specified `base` subscription
+    const forwardAll = baseEvent => {
+      forwardEvent(`${baseEvent}.data`)
+      forwardEvent(`${baseEvent}.error`)
+    }
+
+    forwardAll('subscribeInvoices')
+    forwardAll('subscribeChannelGraph')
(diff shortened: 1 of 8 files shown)


services/grpc/grpc.js
```diff
@@ -160,6 +162,9 @@ class Grpc extends EventEmitter {
     Lightning.removeAllListeners('subscribeInvoices.data')
     Lightning.removeAllListeners('subscribeTransactions.data')
     Lightning.removeAllListeners('subscribeChannelGraph.data')
+    Lightning.removeAllListeners('subscribeChannelGraph.error')
```
mrfelton: No listeners were added for `subscribeChannelGraph.error`

korhaliv: updated. added forwarding of data/error events for all subscriptions

PR #2147, opened 2019-05-02
feat(wallet): apply BOS node scores to lnd autopilot

mrfelton:
## Description:

Apply BOS node scores and to lnd autopilot.

Note: Builds on #2143

## Motivation and Context:

Fix #2143

## How Has This Been Tested?

Manually

## Types of changes:

Feature

## Checklist:

### DIFF
--- a/renderer/reducers/autopilot.js
+++ b/renderer/reducers/autopilot.js
@@ -1,0 +1,74 @@
+import { grpcService } from 'workers'
+import { requestNodeScores } from '@zap/utils/api'
+import { infoSelectors } from './info'
+
+// ------------------------------------
+// Constants
+// ------------------------------------
+
+export const UPDATE_AUTOPILOT_NODE_SCORES = 'UPDATE_AUTOPILOT_NODE_SCORES'
+export const UPDATE_AUTOPILOT_NODE_SCORES_SUCCESS = 'UPDATE_AUTOPILOT_NODE_SCORES_SUCCESS'
+export const UPDATE_AUTOPILOT_NODE_SCORES_FAILURE = 'UPDATE_AUTOPILOT_NODE_SCORES_FAILURE'
+
+// ------------------------------------
+// Actions
+// ------------------------------------
+
+// Fetch autopilot node scores.
+export const updateAutopilotNodeScores = () => async (dispatch, getState) => {
+  const state = getState()
+
+  // Only apply node scores to local nodes where autopilot is enabled.
+  const { type, autopilot } = state.lnd.lndConfig
+  if (type !== 'local' || !autopilot) {
+    return
+  }
+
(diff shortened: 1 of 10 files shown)


utils/api.js
```diff
@@ -38,6 +38,35 @@ export function requestSuggestedNodes() {
...
+  const code = chainMap[chain]
+  const suffix = networkMap[network]
+  const chainCode = `${code}${suffix}`
+  const BASE_URL = `https://nodes.lightning.computer/availability/v1/${chainCode}.json`
+  return axios({
+    method: 'get',
+    url: BASE_URL,
+  }).then(response => {
+    const result = {}
+    response.data.scores.reduce((map, { public_key, score }) => {
```
korhaliv: this reduce could be made more readable if used in convertional form IMO.
Something like
```
  const result =  response.data.scores.reduce((map, { public_key, score }) => {
      if (typeof public_key !== 'string' || !Number.isInteger(score)) {
        throw new Error('Invalid node score format!')
      }
      map[public_key] = score / 100000000.0
      return map
}, {})```

File: utils/api.js [hunk 2]
```diff
@@ -38,6 +38,35 @@ export function requestSuggestedNodes() {
...
+      if (typeof public_key !== 'string' || !Number.isInteger(score)) {
+        throw new Error('Invalid node score format!')
```
korhaliv: Do we really want to discard the entire process if one entry is invalid? Maybe just skip the entry ?

renderer/reducers/autopilot.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  try {
+    dispatch({ type: UPDATE_AUTOPILOT_NODE_SCORES })
+
+    const chain = infoSelectors.chainSelector(state)
+    const network = infoSelectors.networkSelector(state)
+    const scores = await requestNodeScores(chain, network)
+    if (!scores) {
+      throw new Error('Node scores empty')
```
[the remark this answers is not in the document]
mrfelton: I'm not sure there is any need. If this throws an error because there are no scores, the `UPDATE_AUTOPILOT_NODE_SCORES_FAILURE` action will be fired and the error detail will be included in that. So you can see the details in the logs already.

In the future, we could consider doing something other than just storing the error in the redux store such adding an entry to a global notification system (which we don't have yet).

As per your other comment below, I have added a warning message when preprocessing the scores in the case that we encounter invalid data.

utils/api.js
mrfelton: agreed. Updated.

utils/api.js
mrfelton: Good point. I've changed this so that we log a warning message and continue with the reducer.

PR #2158, opened 2019-05-04
fix(onboarding): improve handling of empty form values

### LINKED ISSUE #2146: Error when trying to save a wallet without a name or alias
## Description

Getting an error when trying to save a wallet without a name or alias.

image

## Expected Behavior

Should save without error

## Steps to Reproduce

Try to save a wallet without a name and/or alias.

mrfelton:
## Description:

- improve handling of empty form values
- strip unknown properties in db.wallet

## Motivation and Context:

Fix #2146 

Additionally, be more strict about which wallet properties we save into the db. Do not save things that are supposed to be dynamically generated at runtime. Instead, generate them after fetching the wallets from the database. This ensures that if we change our underlaying setup at a later date users do not end up with stale wallet configs in their database.

**before:**

image

**after:**

image

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix / enhancement

## Checklist:

### DIFF
--- a/electron/preload.js
+++ b/electron/preload.js
@@ -94,7 +94,11 @@ async function generateLndConfigFromWallet(wallet) {
   // Wait for the config to become fully initialized.
   await lndConfig.isReady
 
-  return lndConfig
+  // lndConfig is an LndConfig class instance that contains methods and properties that are promises such as the
+  // `isReady` prop used above. We convert this to a plain object so that we can store in Redux and freely pass
+  // between the renderer and Web Worker processes without needing to proxy. This also flattens getter method
+  // properties to plan props.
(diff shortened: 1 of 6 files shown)


electron/preload.js
```diff
@@ -94,7 +94,7 @@ async function generateLndConfigFromWallet(wallet) {
   // Wait for the config to become fully initialized.
   await lndConfig.isReady
 
-  return lndConfig
+  return Object.assign({}, lndConfig)
```
korhaliv: Could you add a comment here, please? I assume we just create an independent copy of config here, so the original one stays untouched no matter what?

mrfelton: Have added some comments to this.

PR #2177, opened 2019-05-08
refactor: consolidate/optimise grpc subscription setup code

### LINKED ISSUE #2166: Reduce frequency of getInfo pollng
## Description

We poll getInfo in order to determine the current sync state of lnd. We poll every 5 seconds. Once we know that the chain has been synced we can reduce this frequency down to something more like once per minute.

mrfelton:
## Description:

- disconnect getinfo stream on chain sync complete
- consolidate grpc subscription setup code
- unsubscribe streams before disconnecting
- dynamic gRPC subscription registration

## Motivation and Context:

Fix #2166 

Since we are not making use of the info subscription after the initial sync has completed I have opted to disable it once the sync has completed and we have activated the channel subscription. 

## How Has This Been Tested?

Manually

## Types of changes:

Refactor / fix

## Checklist:

### DIFF
--- a/services/grpc/grpc.js
+++ b/services/grpc/grpc.js
@@ -127,6 +127,7 @@ class ZapGrpc extends EventEmitter {
         const { Lightning } = this.services
         this.subscriptions['channelGraph'] = Lightning.subscribeChannelGraph()
         this.subscribe('channelGraph')
+        this.unsubscribe('getinfo')
       }
     })
   }
--- a/services/grpc/grpc.js
+++ b/services/grpc/grpc.js
@@ -16,22 +16,26 @@ const GRPC_LIGHTNING_SERVICE_ACTIVE = 'GRPC_LIGHTNING_SERVICE_ACTIVE'
  * @extends EventEmitter
  */
 class ZapGrpc extends EventEmitter {
-  static INITIAL_STATE = {
+  /**
+   * State properties that should be reset after a disconnect.
+   * @type {Object}
+   */
+  static VOLATILE_STATE = {
     options: {},
     services: {},
-    subscriptions: {},
-  }
-
-  static SUBSCRIPTIONS = {
-    invoices: 'subscribeInvoices',
-    transactions: 'subscribeChannelGraph',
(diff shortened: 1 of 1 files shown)


services/grpc/grpc.js
```diff
@@ -19,15 +19,15 @@ class ZapGrpc extends EventEmitter {
...
-    'subscribeChannelGraph',
-    'subscribeTransactions',
-    'subscribeGetInfo',
-  ]
+  static SUBSCRIPTIONS = {
+    invoices: 'subscribeInvoices',
+    transactions: 'subscribeChannelGraph',
+    channelgraph: 'subscribeTransactions',
+    info: 'subscribeGetInfo',
+  }
```
korhaliv: Just a thought. Do we really need this to be hardcoded? I'm thinking we can use a more dynamic approach. like `registerService` and pre-register those 4. This eventually should work better with LND API versioning. Like for instances soon we will be adding `scb` to a list of subscriptions.

services/grpc/grpc.js
```diff
@@ -112,47 +98,62 @@ class ZapGrpc extends EventEmitter {
...
     // this is needed because LND chanRouter waits for chain sync
-    // to complete before accepting subscriptions
+    // to complete before accepting subscriptions.
     this.on('subscribeGetInfo.data', data => {
       const { synced_to_chain } = data
-      if (synced_to_chain && !this.subscriptions['channelGraph']) {
-        const { Lightning } = this.services
-        this.subscriptions['channelGraph'] = Lightning.subscribeChannelGraph()
-        this.subscribe('channelGraph')
+      if (synced_to_chain && !this.subscriptions['channelgraph']) {
```
korhaliv: let's use `this.subscriptions.channelgraph` ?

mrfelton: Updated

services/grpc/grpc.js
mrfelton: I've added a basic `registerSubscription` method that can be used to setup the mapping in a slightly more dynamic way.

PR #2204, opened 2019-05-15
feat(wallet): additional activity search criteria

### LINKED ISSUE #2174: Activity search should allow filtering on node alias
## Detailed Description

Activity search should allow filtering on node alias, especially since we display node alias in the list view.

mrfelton:
## Description:

Enhance activity search to support searching on destination addresses, node pubkey, alias, and preimage.

## Motivation and Context:

Fix #2174

## How Has This Been Tested?

Manually

## Types of changes:

Enhancement

## Checklist:

### DIFF
--- a/renderer/reducers/activity.js
+++ b/renderer/reducers/activity.js
@@ -1,6 +1,5 @@
 import { createSelector } from 'reselect'
-import { decodePayReq } from '@zap/utils/crypto'
-
+import { decodePayReq, getNodeAlias } from '@zap/utils/crypto'
 import { openModal, closeModal } from './modal'
 import { fetchDescribeNetwork } from './network'
 import { fetchTransactions, transactionsSelectors } from './transaction'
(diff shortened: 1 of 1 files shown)


renderer/reducers/activity.js
```diff
@@ -141,12 +140,28 @@ const transactionsSelector = state => transactionsSelectors.transactionsSelector
...
 
+/**
+ * Augment payments with additional data.
+ */
+const decoratedPaymentsSelector = createSelector(
+  paymentsSelector,
+  nodesSelector,
+  (payments, nodes) =>
+    payments.map(payment => {
+      const destPubKey = payment.path[payment.path.length - 1]
```
korhaliv: Is `payment.path` always set?

renderer/reducers/activity.js
```diff
@@ -279,7 +294,11 @@ const applySearch = (data, searchText) => {
         item =>
           (item.tx_hash && item.tx_hash.includes(searchText)) ||
           (item.payment_hash && item.payment_hash.includes(searchText)) ||
-          (item.payment_request && item.payment_request.includes(searchText))
+          (item.payment_preimage && item.payment_preimage.includes(searchText)) ||
```
korhaliv: consider extracting this into a helper to reduce potential typo errors. Smth like
`const matches = (prop) => item[props] &&  item[prop].includes(searchText)`
and then 
`['text_has','payment_hash','payment_preimage',...].some(matches)`

renderer/reducers/activity.js
mrfelton: It should be, but I'ave added some safeguards around this just incase.

renderer/reducers/activity.js
mrfelton: Yes, good idea. Have refactored this.

PR #2235, opened 2019-05-17
fix(grpc): setup channel graph stream after sync

### LINKED ISSUE #2226: Channel graph subscription sometimes fails to get setup after local wallet sync completion
## Detailed Description

Sometimes we fail to set up the channel graph subscription after logging into a local wallet with the following error in the console:

```
zap:grpc [ERR]  CHANNELGRAPH ERROR: Error: router not started
    at Http2CallStream.call.on (call.js?4dae:68)
    at Http2CallStream.emit (events.js:187)
    at Http2CallStream.endCall (call-stream.js?85fb:74)
    at handlingTrailers (call-stream.js?85fb:163)
    at process._tickCallback (internal/process/next_tick.js:68) +3ms
```

This is an issue since updating to `0.6.1-beta-rc1-1-g83adaaa0`. Steps were taken already to try to resolve this by waiting until the `sy8nced_to_chain` flag has flipped to true before trying to set up the subscription. However, this doesn't seem to be enough as there is no guarantee that the lnd router has been started at that point.

## Context

Logging in to a local wallet

## Possible Implementation

**Basic fix**
Add delay between `sync_to_chain` detection and setting up channel graph subscription.

**Better fix**
If the subscription fails to get set up, keep automatically retrying, ideally on some kind of backoff schedule.

mrfelton:
## Description:

If we fail to subscribe to the channel graph after initial sync has completed keep retrying until we get a successful connection.

## Motivation and Context:

Fix #2226

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/services/grpc/grpc.js
+++ b/services/grpc/grpc.js
@@ -4,6 +4,7 @@ import { proxyValue } from 'comlinkjs'
 import { status } from '@grpc/grpc-js'
 import LndGrpc from 'lnd-grpc'
 import { grpcLog } from '@zap/utils/log'
+import delay from '@zap/utils/delay'
 import lightningMethods from './lightning.methods'
 import lightningSubscriptions from './lightning.subscriptions'
 import { forwardAll, unforwardAll } from './helpers'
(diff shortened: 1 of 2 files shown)


services/grpc/grpc.js
```diff
@@ -248,6 +248,42 @@ class ZapGrpc extends EventEmitter {
...
+
+      this.subscribe('channelgraph')
+
+      // If the channel graph subscription fails to start, try again in a bit.
+      if (this.activeSubscriptions.channelgraph) {
+        this.activeSubscriptions.channelgraph.once('error', async error => {
+          if (error.message === 'router not started') {
+            grpcLog.warn('Unable to subscribe to channelgraph. Will try again in %sms', timeout)
+            await delay(timeout)
+            initSubscription()
```
korhaliv: Is it possible to use `scheduler` instead of managing backoff manually?

mrfelton: I originally tried to do it that way, but the issue is that this isn't quite a straightforward as retrying something on a backoff schedule as here there is an actually trigger other than the time based trigger - we only want to schedule a retry if we were unsuccessful in our previous subscribe attempt. ie, we are using the error event to trigger scheduling of future retry.

korhaliv: Makes sense. What do you think if we extend scheduler and add the ability to pause task execution? You will pause on success and resume on error. This way the backoff schedule will be maintained. The initial version of scheduler was retryOnError util, which has transformed into scheduler. This pattern will probably prove useful in future cases too.

mrfelton: I'm not sure that would work either, since there is no notable success case here - success is just the absence of an error event.

PR #2289, opened 2019-05-27
build(icons): add ability to generate specific icons

korhaliv:
## Description:
Add ability to generate a specific SVG icon instead of re-generating all of them

## Motivation and Context:
Currently `genereate-icons` re-generates all icons every time, which is redundant since icons are added one by one to the project

## How Has This Been Tested?
Manually
`yarn generate-icon icons/icon.svg`

## Types of changes:
Devops change

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -343,6 +343,7 @@
     "jstimezonedetect": "1.0.6",
     "lnd-grpc": "0.2.9",
     "lndconnect": "0.2.7",
+    "lodash.camelcase": "4.3.0",
     "lodash.debounce": "4.0.8",
     "lodash.get": "4.4.2",
     "lodash.intersection": "4.4.0",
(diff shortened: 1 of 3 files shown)


genIcons.js
```diff
@@ -0,0 +1,8 @@
+const { execSync } = require('child_process')
+
+const arg = process.argv[2] || ''
+const icon = `icons/${arg}`
+execSync(
+  `npx @svgr/cli --icon -d renderer/components/Icon ${icon} && npm run lint-fix -- renderer/components/Icon`,
```
mrfelton: Can you update the lint command so that it only runs against the file in question too?

korhaliv: good catch

korhaliv:
@mrfelton updated

genIcons.js
```diff
@@ -0,0 +1,15 @@
...
+const icon = process.argv[2] || ''
+// converts svg icon name to generated js file name (icon-name.svg->IconName.js)
+function getGenerateIconName() {
+  const pascalCaseName = camelCase(icon)[0].toUpperCase() + icon.slice(1)
+  const [convertedIconName] = pascalCaseName.split('.')
+  return `${convertedIconName}.js`
+}
+execSync(
+  // eslint-disable-next-line max-len
+  `npx @svgr/cli --icon -d renderer/components/Icon ${icon} && npm run lint-fix -- renderer/components/Icon/${getGenerateIconName()}`,
```
mrfelton: This should use `lint-fix-base` rather than `lint-fix` as the former allows you to pass your lint paths to it whilst the later applies a default lint path of `"./**/*.{js,json,md}"`

PR #2298, opened 2019-05-28
fix(app): wait for settings on app init

mrfelton:
## Description:

Wait for settings on app init

## Motivation and Context:

Prevents us from redirecting to the Initialiser before settings have been loaded from the database and ensures that if the user was logged into a wallet when they quit the app, they remain logged into it when then restart the app.

This preserves exist5ing functionality from master which got broken as part of the settings revamp.

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/renderer/reducers/settings.js
+++ b/renderer/reducers/settings.js
@@ -3,12 +3,15 @@ import merge from 'lodash.merge'
 import { createSelector } from 'reselect'
 import set from 'lodash.set'
 import difference from '@zap/utils/difference'
+import { showError } from './notification'
 
 // ------------------------------------
 // Constants
 // ------------------------------------
+export const INIT_SETTINGS = 'INIT_SETTINGS'
+export const INIT_SETTINGS_SUCCESS = 'INIT_SETTINGS_SUCCESS'
+export const INIT_SETTINGS_FAILURE = 'INIT_SETTINGS_FAILURE'
 
-export const SET_SETTINGS = 'SET_SETTINGS'
 export const SET_SETTING = 'SET_SETTING'
 export const UPDATE_CONFIG = 'UPDATE_CONFIG'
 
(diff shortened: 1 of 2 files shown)


renderer/reducers/settings.js
```diff
@@ -20,12 +23,23 @@ export const UPDATE_CONFIG = 'UPDATE_CONFIG'
...
+  try {
+    const allSettings = await window.db.settings.toArray()
+    const settings = allSettings.reduce((obj, item) => {
+      obj[item.key] = item.value
+      return obj
+    }, {})
+    dispatch({ type: INIT_SETTINGS_SUCCESS, settings })
+  } catch (e) {
+    dispatch({ type: INIT_SETTINGS_FAILURE, initSettingsError: e })
+    dispatch(showError(`Unable to load settings: ${e.message}`, { timeout: 0 }))
```
korhaliv: This error is not intl compatible. we have a similar problem in a couple of places. Theoretically, we could create a middleware or some other approach to be able to use intl in actions. I'd suggest at least to add `TODO` item here, so these places are easily searchable in future

renderer/reducers/settings.js
```diff
@@ -60,25 +76,23 @@ export const saveConfigOverrides = values => async (dispatch, getState) => {
...
-  }),
-  [SET_SETTING]: (state, { key, value }) => ({
+  [INIT_SETTINGS]: state => ({ ...state }),
+  [INIT_SETTINGS_SUCCESS]: (state, { settings }) => ({
     ...state,
-    ...{ ...state.settings, [key]: value },
+    ...settings,
+    isSettingsLoaded: true,
   }),
+  [INIT_SETTINGS_FAILURE]: (state, { initSettingsError }) => ({ ...state, initSettingsError }),
```
korhaliv: are init settings called only once per APP lifecycle? if not, then we need to clear error somewhere

mrfelton: Yes, this is expected to be called once and once only, when the root component mounts.

mrfelton: Have added a comment to note that.

renderer/reducers/settings.js
mrfelton: Good call. Have added a `TODO`

PR #2310, opened 2019-05-29
Implement new settings page

### LINKED ISSUE #2285: Implement new settings page design
## Detailed Description

Implement new settings page design.

## Possible Implementation

image

**Preferences**
```
 - General
   - theme
   - language
   - fiat currency
   - autoupdate (enabled, frequency, check now button)
 - Wallet
   - receiving address format
   - sending target confs
   - invoice expire time
   - block exporer
 - Advanced
   - ports
   - neutrino backends
```

**Wallet settings:**
```
 - Basic
   - chain (readonly)
   - network (readonly)
 - General
   - name
   - alias
   - autopilot
 - Backups
   - Google
   - Dropbox
   - Local
 - Actions
   - Delete
   - Resync
   - Backup
```

mrfelton:
## Description:

Implement new settings page

## Motivation and Context:

Fixes #2285

## How Has This Been Tested?

Manually

## Types of changes:

Enhancement

## Checklist:


renderer/components/Settings/SettingsPage.js
```diff
@@ -0,0 +1,125 @@
...
+import messages from './messages'
+
+const items = [
+  {
+    id: 'general',
+    title: 'General',
+  },
+  {
+    id: 'wallet',
+    title: 'Wallet',
```
korhaliv: Does this need to be intl-ed?

renderer/components/UI/Modal.js
```diff
@@ -14,49 +15,60 @@ export const ModalOverlayStyles = () => `
...
 
-ModalHeader.defaultProps = {
-  hasHeader: true,
+ModalLogo.defaultProps = {
+  hasLogo: true,
 }
 
+const ModalHeader = props => (
+  <Panel.Header
+    css={{ position: 'absolute', top: 0, left: 0, right: 0, 'z-index': '1' }}
```
korhaliv: Could you please extract this to styled

renderer/containers/ModalStack.js
```diff
@@ -75,9 +117,7 @@ function ModalStack(props) {
         /* eslint-disable react/display-name */
         (styles => (
           <Container style={styles}>
-            <Modal onClose={() => closeModal(modal.id)}>
-              <ModalContent closeModal={closeModal} type={modal.type} />
-            </Modal>
+            <ModalContent closeModal={() => closeModal()} type={modal.type} />
```
korhaliv: can be just `closeModal={closeModal}` now

renderer/reducers/app.js
```diff
@@ -105,6 +106,16 @@ export const terminateApp = () => async dispatch => {
   dispatch(send('terminateAppSuccess'))
 }
 
+/**
+ * openPreferences - IPC handler for 'openPreferences' message.
+ */
+export const openPreferences = () => (dispatch, getState) => {
+  const modals = modalSelectors.getModalState(getState())
+  if (!modals.find(m => m.type === 'SETTINGS')) {
```
korhaliv: could be a good idea to add this kind of guard to `openModal` itself to prevent this same issue for other modals

renderer/components/Settings/SettingsFieldsWallet.js
```diff
@@ -0,0 +1,75 @@
...
+            initialSelectedItem={currentConfig.address}
+            items={addressItems}
+            messageMapper={addressMessageMapper}
+          />
+        }
+      />
+
+      <Bar variant="light" />
+
+      <DataRow
```
korhaliv: Could be a good extraction, since it's basically `fn(targetConf)`

renderer/components/Settings/SettingsPage.js
```diff
@@ -0,0 +1,125 @@
+import React, { useState } from 'react'
+import PropTypes from 'prop-types'
+import { FormattedMessage } from 'react-intl'
+import { useFormState, useFormApi } from 'informed'
+import styled from 'styled-components'
+import { Box } from 'rebass'
+import difference from '@zap/utils/difference'
+import merge from 'lodash.merge'
```
korhaliv: Needs to be moved above

renderer/components/UI/Button.js
```diff
@@ -10,6 +10,7 @@ const Wrapper = styled(BaseButton)`
   font-weight: 300;
   line-height: '18px';
   white-space: nowrap;
+  text-align: ${({ justify }) => justify};
```
korhaliv: maybe call this prop `textAlign` ? sounds a bit confusing since `justify` is a prop name for `text-align` in css

renderer/components/UI/Sidebar.js
```diff
@@ -8,7 +8,12 @@ import BackgroundTertiary from './BackgroundTertiary'
  * <Sidebar>Some content</Sidebar>
  */
 const Sidebar = ({ ...props }) => (
-  <BackgroundTertiary as="aside" css={{ overflow: 'hidden' }} width={3 / 12} {...props} />
+  <BackgroundPrimary
+    as="aside"
+    css={{ overflow: 'hidden', 'box-shadow': '0 2px 24px 0 rgba(0, 0, 0, 0.5)' }}
```
korhaliv: I'd say extract this into styled too. generally, if it's more than one prop I try to extract. what do you think?

renderer/containers/ModalStack.js
```diff
@@ -23,32 +23,68 @@ const Container = styled(animated.div)`
 const ModalContent = ({ type, closeModal }) => {
   const doCloseModal = () => closeModal()
   switch (type) {
-    case 'SETTINGS_FORM':
-      return <Settings onSubmit={doCloseModal} />
+    case 'SETTINGS':
+      return (
+        <Modal onClose={closeModal} p={0}>
```
korhaliv: Not super clean that we have a lot of `<Modal onClose={closeModal}>` now, with only some of them having `p={0}`
Maybe it would be cleaner to have 
```
<Modal onClose={closeModal} {...getParams(modalype)}>
{getContent(modalype)}
</Modal>
```
but i don't have a strong opinion on this

renderer/components/UI/Sidebar.js
mrfelton: I'm a little in two minds about this. I tend to prefer only extracting these into separate components when it actually adds something in terms of reusability or readability. If it's just two or three one-off style props that aren't being reused elsewhere then I don't have a problem using the styled-components built in `css` prop. That is what it is for after all.

I've updated these instances in any case.

renderer/containers/ModalStack.js
mrfelton: Agreed. I actually thought I changed this but seems the commit where I did got lost somewhere. I've updated this so that the Modal is fullscreen by default and it's up to users of the `Modal` component to add any padding that they want. So no more `padding={-4}` to remove default padding.

renderer/containers/ModalStack.js
mrfelton: Updated.

renderer/reducers/app.js
mrfelton: Good idea. I've moved this guard back to `openModal`

renderer/components/App/App.js
```diff
@@ -3,7 +3,7 @@ import createScheduler from '@zap/utils/scheduler'
 import PropTypes from 'prop-types'
 import Activity from 'containers/Activity'
 import Wallet from 'containers/Wallet'
-import { MainContent } from 'components/UI'
+import { Box } from 'rebass'
```
korhaliv: Needs to be moved upwards

renderer/components/Settings/SettingsFieldsWallet.js
korhaliv: I was referring to ones that use `NumberField`. Is see 3 similar ones. Could be `NumberDataRow`

renderer/components/Settings/SettingsForm.js
```diff
@@ -0,0 +1,50 @@
...
+  configureAutoUpdater,
+  onSubmit,
+  setLocale,
+  saveConfigOverrides,
+  showNotification,
+  children,
+}) => {
+  const handleSubmit = async values => {
+    // Save the updated settings.
+    await saveConfigOverrides(values)
```
korhaliv: Maybe it's worth to also handle submit exception and show a notif in that case?

renderer/components/UI/Button.js
korhaliv: Ok got it

renderer/components/UI/Modal.js
```diff
@@ -14,49 +16,57 @@ export const ModalOverlayStyles = () => `
...
-      <Panel.Header pt={3} px={3}>
-        {hasClose && <CloseButton onClick={onClose} />}
-        {hasHeader && <ModalHeader hasHeader={hasHeader} />}
-      </Panel.Header>
-      <Panel.Body pb={4} px={4} {...rest} css={{ 'overflow-y': 'overlay', 'overflow-x': 'hidden' }}>
-        {children}
-      </Panel.Body>
+      {(hasClose || hasLogo) && (
+        <ModalHeader pt={3} px={3}>
+          {hasClose && <CloseButton onClick={onClose} />}
```
korhaliv: I wonder if `hasClose` can actually be removed at all and `onClose` could be used to determine the visibility. It's probably not worth displaying X if the handler is not set anyways. For the sake of keeping props count smaller

renderer/containers/ModalStack.js
```diff
@@ -14,41 +14,76 @@ import ChannelDetailModal from 'containers/Channels/ChannelDetailModal'
...
+      return (
+        <Modal onClose={closeModal}>
+          <SettingsPage />
+        </Modal>
+      )
 
     case 'AUTOPAY':
-      return <Autopay mx={-4} />
+      return (
+        <Modal onClose={closeModal} p={4}>
```
korhaliv: Now that you are using `p={4}` (except settings, but maybe it can be fine-tuned there too) can we move  `<Modal onClose={closeModal} p={4}>` higher and basicall have
```
<Modal onClose={closeModal} p={4}>
   switch
</Modal>
```
?

renderer/components/Settings/SettingsFieldsWallet.js
mrfelton: Ok yes that actually reads much better. Have updated.

renderer/components/Settings/SettingsForm.js
mrfelton: good call. Have added errors case

renderer/components/UI/Modal.js
mrfelton: I've actually done some further refactoring of this in a more recent PR so don't want to mess with this anymore here. I'll consider this change in that PR.

renderer/containers/ModalStack.js
mrfelton: Updated.

PR #2333, opened 2019-06-03
fix(wallet): handle case when payment path is missing

mrfelton:
## Description:

Handle case when payment path is missing

## Motivation and Context:

`payment.path` missing in recent versions of lnd.

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix

## Checklist:


renderer/components/Activity/Payment/Payment.js
```diff
@@ -1,18 +1,34 @@
...
+      pubkey = paymentRequest.payeeNodeKey
+    }
+
+    // If we know the pubkey, try to lookup its alias.
+    if (pubkey) {
+      const node = nodes.find(n => pubkey === n.pub_key)
+      if (node && node.alias.length) {
+        return node.alias
+      }
+      return pubkey.substring(0, 10)
```
korhaliv: should we use `truncateNodePubkey` from `network.js` ?

mrfelton: Good call. I've extracted the `truncateNodePubkey` method into a standalone utility method and then updated the codebase to use this.

mrfelton:
Should be good now @JimmyMow

PR #2335, opened 2019-06-03
feat(ui): implement new channel header

### LINKED ISSUE #2332: Implement new channel header
## Detailed Description

Implement new channel header

image

mrfelton:
## Description:

Implement new channel header

## Motivation and Context:

Fix #2332

## How Has This Been Tested?

Manually

## Types of changes:

UI enhancement

## Checklist:


JimmyMow:
Screen Shot 2019-06-03 at 8 04 09 PM

Nice @mrfelton!

Few comments:

I feel like the header is too close to the 3 Window buttons in the top left and adding some top padding would be good. Thoughts?

The refresh icon is thicker and not the icon that is in Zeplin. We also should maybe change the icon in the Activity search header thing if we are going to change it here? (For the record, I like the thinner refresh button)

Screen Shot 2019-06-03 at 8 48 48 PM

Is the order arrow icon separated from the other icons on purpose? If so, I wonder what everyone thinks about that. Grouping all the icons would probably be cleaner in my opinion

Lastly, we discussed in Slack the need to create space between the subheader/channel cards and the search bar thingy. Do we still want to do that? I think so personally

PR #2390, opened 2019-06-10
feat(ui): make wallet connection indicator more informative

korhaliv:
## Description:
Use `busy` state of the wallet connectivity indicator if `synced_to_chain` is false to give the user more info on the current state of the wallet. This is especially useful when used with remote nodes. This PR also re-instates `getInfo` polling after initial sync is done, but does it in a less resource-consuming manner by reducing the polling interval

## How Has This Been Tested?
Manually. Connected to a node that was in a process of sync

image

## Types of changes:
Enhancement

## Checklist:

### DIFF
--- a/services/grpc/grpc.js
+++ b/services/grpc/grpc.js
@@ -5,6 +5,7 @@ import { status } from '@grpc/grpc-js'
 import LndGrpc from 'lnd-grpc'
 import { grpcLog } from '@zap/utils/log'
 import delay from '@zap/utils/delay'
+import isObject from '@zap/utils/isObject'
 import { forwardAll, unforwardAll } from '@zap/utils/events'
 import lightningMethods from './lightning.methods'
 import lightningSubscriptions from './lightning.subscriptions'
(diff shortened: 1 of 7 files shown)


services/grpc/grpc.js
```diff
@@ -110,20 +111,43 @@ class ZapGrpc extends EventEmitter {
     this.on('subscribeGetInfo.data', async data => {
       const { synced_to_chain } = data
       if (synced_to_chain && !this.activeSubscriptions.channelgraph) {
-        this.unsubscribe('info')
+        // reduce poll interval to once a minute after synced_to_chain is true
+        await this.unsubscribe('info')
+        this.subscribe({ name: 'info', params: { pollInterval: 5000 } })
```
mrfelton: comment says once per minute, but this is actually running every 5 seconds. I think once per minute is probably enough.

korhaliv: sorry, reduced for testing and forgot to roll-back before force-pushing

PR #2399, opened 2019-06-11
 perf: debounce updateNodeData

### LINKED ISSUE #2225: Reduce frequency of updates to node data
## Detailed Description

The `UPDATE_NODE_DATA` action gets called a lot in quick succession after creating a new wallet. We call this every time receive an update from the channel graph subscription with updated node data.

## Context

excessive calls to `UPDATE_NODE_DATA` noticed in console after new wallet creation.

## Possible Implementation

We should batch/throttle these in order to prevent excessive re-renders.

korhaliv:
## Description:
Resolve #2225
This PR also introduces easily reusable `debounceMiddleware`.  See code comments for usage example

## How Has This Been Tested?
Manually

## Types of changes:
Perf optimization

## Checklist:

### DIFF
--- a/utils/debouncedFuncQueue.js
+++ b/utils/debouncedFuncQueue.js
@@ -1,0 +1,31 @@
+import debounce from 'lodash/debounce'
+/**
+ * createDebouncedQueue - Creates a debounced function that delays invoking func until after wait milliseconds have
+ * elapsed since the last time the debounced function was invoked
+ * when function is finally invoked it's passed an array of accumulated  since the last call arguments
+ *
+ * @export
+ * @param {Function} func The function to debounce
+ * @param {number} wait The number of milliseconds to delay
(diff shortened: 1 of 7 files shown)


utils/debouncedFuncQueue.js
```diff
@@ -0,0 +1,32 @@
+import debounce from 'lodash.debounce'
```
mrfelton: This should now be in the form:
```js
import { debounce } from 'lodash/debounce'
```

korhaliv: updated

korhaliv: should be
`import debounce from 'lodash/debounce'`
:)

mrfelton: doh!

Issue #2377, opened 2019-06-07
Update lnd to latest master build


mrfelton:
## Detailed Description

Update lnd to incorporate recent improvement

- Cut new release
- Source in new release
- Add related `rpc.proto` file to codebase
- Use `rpc.proto` override when connecting to local lnd
- Ensure backwards compatibility

JimmyMow:
this can be closed right? #2406

PR #2406, opened 2019-06-12
fix(deps): update lnd to v0.6.1-beta-255-gb3fd7cc8

mrfelton:
## Description:

This PR updates lnd to a recent build from master (0.6.1-beta-255). I reviewed the `rpc.proto` and the only changes vs the version of the proto files that we are currently using from 0.6.1-beta are very minor and not anything that would affect us, so I have not updated them as part of this.

In my limited testing, everything appears to be working normally.

## Motivation and Context:

https://github.com/LN-Zap/zap-desktop/issues/2377

## How Has This Been Tested?

Manually

## Types of changes:

Lnd update

## Checklist:

### DIFF
--- a/electron/migrations/delete-tmp-wallets.js
+++ b/electron/migrations/delete-tmp-wallets.js
@@ -1,0 +1,22 @@
+import config from 'config'
+import { mainLog } from '@zap/utils/log'
+import { getAllLocalWallets, deleteLocalWallet } from '@zap/utils/localWallets'
+
+/**
+ * migration - Migration script to delete all tmp wallets from the filesystem.
+ *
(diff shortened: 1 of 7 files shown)


mrfelton:
Actually, there is an issue with this that I'm working on resolving.

mrfelton:
Fixed. The problem was that the updated lnd includes a migration script which expects the channel db to be present. This is not the case for our tmp wallets (the ones we use for generating seeds) so this caused our onboarding process to break with a fatal error from lnd when trying to start these wallets. Worked around the issue by adding a migration script to delete the tmp wallets.

PR #2424, opened 2019-06-13
feat(settings): add ability to select exchange rate provider

korhaliv:
## Description:
Add ability to select rate provider(exchange) that is used to calculate fiat values

## How Has This Been Tested?
Manually

image

## Types of changes:
New feature

## Checklist:


renderer/components/Settings/SettingsPage.js
```diff
@@ -105,9 +105,14 @@ const SettingsPage = ({ currentConfig, ...rest }) => {
           <FormattedMessage {...messages.settings_title} />
         </Heading.h1>
 
-        <SettingsForm>
+        <SettingsForm onSubmit={fetchTickers}>
```
mrfelton: I'd probably move this to `SettingsForm::handleSubmit` which is where we handle all the other special submit behaviour for the form.

renderer/components/Settings/messages.js
```diff
@@ -37,4 +37,10 @@ export default defineMessages({
   lndTargetConfirmations_fast_description: 'Number of blocks to target for "fast" sending speed.',
   fieldgroup_general: 'General',
   fieldgroup_wallet: 'Wallet',
+  rateProvider_label: 'Rate provider',
+  rateProvider_option_coinbase: 'Coinbase',
```
mrfelton: Do we actually want the company names to be translatable? I did it this was with the block explorers but now question wether there is really any point. Would these company names ever be written differently in different languages?

korhaliv: I think we don't. I've used that to make it consistent with how block explorers are handled mostly

korhaliv:
@mrfelton updated

korhaliv:
> Screen Shot 2019-06-13 at 4 23 57 PM
> 
> @korhaliv it seems that conversions aren't working. Everything in my activity list is `$0.00`

I think that's default rate provider config issue that @mrfelton has also pointed out. Should be fine now

renderer/components/Settings/messages.js
korhaliv: I've updated rate provider Select to use names directly from the config (which is probably the right thing to do in the first place because previously those weren't used at all)
Also, update blockexplorer to do the same for consistency

PR #2461, opened 2019-06-19
fix(ui): use Select widget for invoice expiry

mrfelton:
## Description:

Use Select widget for invoice expiry

## Motivation and Context:

Improved usability

## How Has This Been Tested?

Manually

**Before:**
image

**After:**
image

## Types of changes:

UI enhancement

## Checklist:


mrfelton:
Issue with far the future date should be resolved @JimmyMow

renderer/reducers/invoice.js
```diff
@@ -115,14 +115,14 @@ export const createInvoice = (amount, cryptoUnit, memo, isPrivate) => async (
...
+    dispatch(createInvoiceSuccess(invoice))
+  } catch (error) {
+    dispatch(createInvoiceFailure(error))
   }
 }
 
 // Receive IPC event for newly created invoice
-export const createdInvoice = invoice => dispatch => {
+export const createInvoiceSuccess = invoice => dispatch => {
   decorateInvoice(invoice)
```
korhaliv: Ideally, we shouldn't mutate invoice here (I know it was like this before this PR)

renderer/components/Settings/SettingsFieldsWallet.js
```diff
@@ -30,7 +31,17 @@ const blockExplorerItems = [
   { key: 'smartbit', value: 'Smartbit' },
 ]
 
-const SettingsFieldsWallet = ({ currentConfig }) => {
+const SettingsFieldsWallet = ({ currentConfig, intl }) => {
+  const invoicesExpireItems = [
```
korhaliv: since this is quite a big map, could be a good idea to move it one level above to avoid re-calculation on every render. I know that we use this style in other places. Their maps are usually smaller (2-3 items) but probably we shouldn't place them inside mappers as well.

korhaliv:
image
Copy needs update

renderer/components/Settings/SettingsFieldsWallet.js
mrfelton: It's here because it relies on `intl` so it can't be generated statically outside of the render.

renderer/reducers/invoice.js
mrfelton: Agreed. Have added to my list and will submit a separate PR to fix this.

renderer/components/Settings/SettingsFieldsWallet.js
mrfelton: I've extracted this out into a helper method and assed the intl module to it. It doesn't avoid recalculation on each render but it is a little cleaner.

PR #2484, opened 2019-06-21
fix(ui): ensure lndconnect string is trimmed

korhaliv:
## Description:
Prevent app from crashing when using a valid LND connect with leading or trailing spaces

## How Has This Been Tested?
Manually

## Types of changes:
Bugfix

## Checklist:

### DIFF
--- a/renderer/components/Onboarding/Steps/ConnectionConfirm.js
+++ b/renderer/components/Onboarding/Steps/ConnectionConfirm.js
@@ -46,7 +46,7 @@ class ConnectionConfirm extends React.Component {
       connectionString,
       startLnd,
     } = this.props
-
+    connectionString = connectionString && connectionString.trim()
     // If we have a hostname, assume we are using the custom form in which host, cer and macaroon paths are supplied.
     if (connectionHost) {
       return startLnd({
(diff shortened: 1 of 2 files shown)


renderer/components/Onboarding/Steps/ConnectionConfirm.js
```diff
@@ -46,7 +46,7 @@ class ConnectionConfirm extends React.Component {
       connectionString,
       startLnd,
     } = this.props
-
+    connectionString = connectionString && connectionString.trim()
```
mrfelton: Instead of doing the trim here at point of usage, perhaps we could do it at the source in `LndConnectionStringInput` using `Informed::mask`?

korhaliv: This sounds like a better approach. Will work on it

korhaliv:
@mrfelton I've added trimming to LndConnectionStringInput but left the trimming code inside `ConnectionConfirm` too for the sake of future-proofness
The reason is - lndconnect string comes to `ConnectionConfirm` from the outside and it can't control whether it's trimmed or not. So if in future we add, say, paste from clipboard - there will still be a crash

PR #2513, opened 2019-06-26
refactor: use immer in network and payment reducers

korhaliv:
## Description:
Currently, we are using vanilla JS in our reducers which makes code boilerplatty, harder to write, read and prone to immutability errors. One option is to use `immutableJS` lib, but it requires a full rewrite of our reducers and a lot of changes to our components. It's also not compatible with libraries like `lodash` which we currently use a lot. This PR proposes to start using `immer` as a solution for the immutability. Benefits of immer:
1. No need to use additional container libs
2. Can be used with vanilla js or with `lodash`
3. No need to change our components
4. Migration can be done iteratively - state produced by immer is fully compatible with the old one
5. State auto-freeze

More info:
https://github.com/immerjs/immer
https://www.netlify.com/blog/2018/09/12/the-rise-of-immer-in-react/
https://hackernoon.com/introducing-immer-immutability-the-easy-way-9d73d8f71cb3

## How Has This Been Tested?
So far manually

## Types of changes:
Refactor/ dx improvement

## Checklist:

### DIFF
--- a/renderer/reducers/network.js
+++ b/renderer/reducers/network.js
@@ -1,6 +1,7 @@
 import { createSelector } from 'reselect'
 import { grpcService } from 'workers'
 import truncateNodePubkey from '@zap/utils/truncateNodePubkey'
+import createReducer from './utils/createReducer'
 
 // ------------------------------------
 // Initial State
@@ -31,29 +32,19 @@ export const UPDATE_NODE_DATA = 'UPDATE_NODE_DATA'
  * @param  {object} nodeData Node data
  * @returns {object} Updated state
  */
-const mergeNodeUpdate = (state, nodeData) => {
-  const { nodes: originalNodes } = state
+const mergeNodeUpdates = (state, nodeData) => {
+  const { nodes } = state
   // Check if this is an existing node
-  const index = originalNodes.findIndex(item => item.pub_key === nodeData.identity_key)
+  const index = nodes.findIndex(item => item.pub_key === nodeData.identity_key)
   // If we didn't find the node, add it to the end of the nodes list.
   // Otherwise update existing.
-  const nodes =
-    index < 0
-      ? [...originalNodes, nodeData]
-      : [
-          ...originalNodes.slice(0, index),
-          {
-            ...originalNodes[index],
-            ...nodeData,
-            last_update: Math.round(new Date() / 1000),
-          },
-          ...originalNodes.slice(index + 1),
-        ]
-
-  return {
-    ...state,
-    nodes,
+  if (index < 0) {
+    nodes.push(nodeData)
+  } else {
+    nodes[index] = { ...nodes[index], ...nodeData, last_update: Math.round(new Date() / 1000) }
   }
+
+  return state
 }
 
 /**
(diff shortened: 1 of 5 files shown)


renderer/reducers/createReducer.js
```diff
@@ -0,0 +1,10 @@
+import produce from 'immer'
+
+const createReducer = (initialState, actionHandlers) => (state = initialState, action) => {
```
mrfelton: Lets move this into `reducers/utils` or something like that

korhaliv: The reason why I went with a separate file was to avoid cyclical deps, since `utils` use reducers a lot.

mrfelton: I agree that it should be in a separate file - but thought it made sense to place that file in a subdir so that it's separate from all of the actual reducers

korhaliv: got it. i'll update

mrfelton: Oh lol, I forgot we actually have `reducers/utils.js` already

korhaliv: yeah..

korhaliv: @mrfelton i've moved createReducer to `reducers/utils/createReducer.js`

korhaliv: oops, rebased against master by mistake. fixed

renderer/reducers/network.js
```diff
@@ -1,7 +1,7 @@
 import { createSelector } from 'reselect'
 import { grpcService } from 'workers'
 import truncateNodePubkey from '@zap/utils/truncateNodePubkey'
-
+import createReducer from './utils/createReducer'
```
mrfelton: Add newline below imports

renderer/reducers/network.js
```diff
@@ -31,31 +31,20 @@ export const UPDATE_NODE_DATA = 'UPDATE_NODE_DATA'
...
-    nodes,
+  if (index < 0) {
+    nodes.push(nodeData)
+  } else {
+    nodes[index] = { ...nodes[index], ...nodeData, last_update: Math.round(new Date() / 1000) }
   }
-}
 
+  return state
+}
```
mrfelton: Add newline below codeblock

renderer/reducers/network.js
```diff
@@ -123,23 +112,34 @@ export const fetchDescribeNetwork = () => async dispatch => {
...
  * @returns {object} Action
  */
-export const receiveDescribeNetwork = ({ nodes }) => ({
-  type: RECEIVE_DESCRIBE_NETWORK,
-  nodes,
-})
+export const receiveDescribeNetwork = ({ nodes }) => dispatch =>
+  dispatch({ type: RECEIVE_DESCRIBE_NETWORK, nodes })
+
+// ------------------------------------
```
mrfelton: This whole comment block can go. Looks like a merge related issue

renderer/reducers/network.js
```diff
@@ -159,16 +159,11 @@ export { networkSelectors }
 // ------------------------------------
 // Reducer
 // ------------------------------------
-
 /**
  * networkReducer - Network reducer.
  *
  * @param  {object} state = initialState Initial state
```
mrfelton: This whole docblock can probably go too since we aren't defining anything new here.

renderer/reducers/payment.js
```diff
@@ -427,8 +402,4 @@ export { paymentSelectors }
  * @param  {object} action Action
  * @returns {object} Next state
```
mrfelton: Same here. docblock can probably go

renderer/reducers/utils/createReducer.js
```diff
@@ -0,0 +1,17 @@
+import produce from 'immer'
+
+/**
+ * createReducer - Creates immer compatible reducer.
+ *
+ * @param {object} initialState
+ * @param {object} actionHandlers
+ * @returns {Function}
```
mrfelton: Needs description of return value

renderer/reducers/payment.js
mrfelton: The documented  params are not correct here. I would probably just remove this docblock completely since we are no longer defining a method here.

mrfelton: This docblock still needs to be removed @korhaliv . Same goes for `network.js`

PR #2550, opened 2019-07-09
refactor: use immer in a number of reducers

korhaliv:
## Description:
Continue migration to `immer` 

## How Has This Been Tested?
Manually

## Types of changes:
Refactor

## Checklist:

### DIFF
--- a/renderer/reducers/info.js
+++ b/renderer/reducers/info.js
@@ -5,6 +5,7 @@ import { grpcService } from 'workers'
 import { initAddresses } from './address'
 import { putWallet, walletSelectors } from './wallet'
 import { settingsSelectors } from './settings'
+import createReducer from './utils/createReducer'
 
 // ------------------------------------
 // Initial State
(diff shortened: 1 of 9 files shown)


renderer/reducers/info.js
```diff
@@ -186,29 +187,21 @@ export const receiveInfo = data => async (dispatch, getState) => {
...
-    }
+    state.infoLoading = false
+    state.infoLoaded = true
+    state.data = data
+    state.chain = get(data, 'chains[0].chain')
+    state.network = get(data, 'chains[0].network')
   },
 }
 
-// ------------------------------------
```
mrfelton: This section heading should remain for consistency

korhaliv: hm. i have removed those in #2513 too, should I reinstate them too?

mrfelton: You removed the one above the reducer declaration:
```
// Reducer
```
Which I think is fine since we are no longer declaring a reducer but simply exporting the reducer as defined by createReducer.

But I think we should keep the heading above the selectors, which is what I was asking you to bring back here:
```
// Selectors
```

korhaliv: ok got it

korhaliv: updated

PR #2557, opened 2019-07-11
fix(wallet): improve handling of invalid invoices

mrfelton:
## Description:

Improve validation and input handling for invoice amounts

## Motivation and Context:

Fix some issues with invoice amount input validation

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix

## Checklist:


renderer/components/UI/FiatAmountInput.js
```diff
@@ -56,6 +55,7 @@ class FiatAmountInput extends React.Component {
   getRules() {
     return {
       precision: 2,
+      step: '.01',
```
korhaliv: should it be `'0.01'` like in other places?

renderer/components/UI/LightningInvoiceInput.js
```diff
@@ -30,11 +30,31 @@ class LightningInvoiceInput extends React.Component {
...
+
+      // If we have a LN invoice, ensure the invoice has an amount.
+      if (invoiceIsLn) {
+        try {
+          const invoice = decodePayReq(value)
+          if (!invoice || (!invoice.satoshis || !invoice.millisatoshis)) {
+            throw new Error('Invalid invoice')
+          }
+        } catch (e) {
+          return intl.formatMessage({ ...messages.invalid_request }, { chain: chainName })
```
korhaliv: if `decodePayReq` throws, won't this cause some UX confusion ? maybe it's better to just explicitly return 
```intl.formatMessage({ ...messages```
 from 
```if (!invoice || (!invoice.satoshis || !invoice.millisatoshis)) {``` 
block

File: renderer/hocs/withNumberValidation.js [hunk 3]
```diff
@@ -0,0 +1,72 @@
...
+        }
+        if (!isEmptyAmount(moreThan)) {
+          validator = validator.moreThan(moreThan)
+        }
+        if (!isEmptyAmount(lessThan)) {
+          validator = validator.lessThan(lessThan)
+        }
+        validator.validateSync(value)
+      } catch (error) {
+        return error.message
```
korhaliv: do we handle/check for this case somewhere below?

renderer/hocs/withNumberValidation.js
```diff
@@ -0,0 +1,72 @@
...
+const isEmptyAmount = amount => amount === undefined || amount === null || amount === ''
+
+/**
+ * withNumberValidation - A HOC that will add validation of a `required` property to a field.
+ *
+ * @param {React.Component} Component Component to wrap
+ * @returns {React.Component} Wrapped component
+ */
+const withNumberValidation = Component =>
+  class extends React.Component {
```
korhaliv: ideally, i'd use functional components for new components (since classes are deprecated sorts of)

renderer/components/UI/FiatAmountInput.js
mrfelton: Yeah, makes sense.

renderer/components/UI/LightningInvoiceInput.js
mrfelton: > if decodePayReq throws, won't this cause some UX confusion

The Error that gets thrown is caught below and handled by returning the appropriate error message, so the UI will behave as expected (show error if the invoice or invoice amount is invalid)

renderer/hocs/withNumberValidation.js
mrfelton: This runs as an informed validate callback where it is expected that you return an error message string in the case of an error. So, it's handled by informed.

renderer/hocs/withNumberValidation.js
mrfelton: Good call. I've converted this.

PR #2585, opened 2019-07-15
fix(ui): display correct channel count when search or filter is applied

korhaliv:
## Description:
Currently, channel count is tied to `allChannels.length` which seem incorrect UX wise because the placement of that text implies that it shows a number of currently displayed channels. This PR makes it to use `currentChannels` instead so it respects search query and/or filters

## How Has This Been Tested?
Manually

image

## Checklist:


JimmyMow:
@korhaliv I like this change. Do you think the copy should reflect what is shown as well? Instead of having `Channels` hardcoded it's `All channels`, `Online channels`, etc depending on the filter?

korhaliv:
I'm torn here. From one point your suggestion adds more context, from the other it's less minimalistic because we will have 2 equal texts on the same page - one that you propose and current select filter name (which is obviously the same). I'd say leave generic one for now. But maybe we can get input from @mrfelton too and decide

mrfelton:
I do think it would be good to include some detail about what the channel count is. It could be a little unclear since it’s not obvious when looking at the channel count what it is counting or that the count is relative to the active filter.

What about appending the name of the active filter after the channel count in the case it’s filtered?

Channels: 35

or if the pending filter is active:

Channels: 7 (pending)

or like:

Channels (pending): 7

Or:

Channels: 7 (show all)

With the ‘show all’ text acting as a link to clear the current filter and revert to showing all channels

korhaliv:
@mrfelton i've fixed some bugs (see the latest commit) and cleaned up the PR. it's read for the re-review

PR #2624, opened 2019-07-26
fix(ui): ensure fiat and crypto inputs in agroup are always updated

### LINKED ISSUE #2602: USD amount not populated on send all
When toggling the `Send all of your funds` toggle in the pay form, the USD amount is not populated:

Screen Shot 2019-07-23 at 10 22 34 AM

korhaliv:
## Description:
fix  #2602

## How Has This Been Tested?
Manually

## Types of changes:
Bugfix

## Checklist:

### DIFF
--- a/renderer/components/UI/CurrencyFieldGroup.js
+++ b/renderer/components/UI/CurrencyFieldGroup.js
@@ -35,51 +35,57 @@ const CurrencyFieldGroup = React.forwardRef(
     },
     ref
   ) => {
-    // whether update of a linked input (crypto<->fiat) is enabled
-    // this ref is used in a process of linked value calculations to
-    // circular updates
-    const shouldUpdateLinked = useRef(true)
+    // Whether update of a linked input (crypto<->fiat) is enabled.
+    // This ref is used in a process of linked value calculations to prevent circular updates.
+    const blockUpdates = useRef(false)
+
+    // Prevent updates of linked form values.
+    const blockLinkedUpdates = () => {
+      blockUpdates.current = true
+    }
+
+    // informed calls onValueChange multiple time during value updates
+    // because of masks and patterns applied on top of UI elements
+    // give value a chance to settle before enabling updates again
+    const unblockLinkedUpdates = async () => {
+      await Promise.resolve()
+      blockUpdates.current = false
+    }
+
+    // Get current block status of linked form elements.
+    const isLinkedUpdatesBlocked = () => {
+      return blockUpdates.current
+    }
 
     /**
      * handleAmountCryptoChange - Set the amountFiat field whenever the crypto amount changes.
      *
-     * @param {Event} e Event
+     * @param {string} value Value
      */
     const handleAmountCryptoChange = async value => {
-      const lastPrice = currentTicker[fiatCurrency]
-      const fiatValue = convert(cryptoUnit, 'fiat', value, lastPrice)
-      // temporarily disable updates to prevent dead loop
-      const doUpdate = shouldUpdateLinked.current
-      shouldUpdateLinked.current = false
-      if (doUpdate) {
+      if (!isLinkedUpdatesBlocked()) {
+        blockLinkedUpdates()
+        const lastPrice = currentTicker[fiatCurrency]
+        const fiatValue = convert(cryptoUnit, 'fiat', value, lastPrice)
         formApi.setValue('amountFiat', fiatValue)
+        await unblockLinkedUpdates()
       }
-      // informed calls onValueChange multiple time during value updates
-      // because of masks and patterns applied on top of UI elements
-      // give value a chance to settle before enabling updates again
-      await Promise.resolve()
-      shouldUpdateLinked.current = true
       onChange && onChange()
     }
 
     /**
      * handleAmountFiatChange - Set the amountCrypto field whenever the fiat amount changes.
      *
-     * @param {Event} e Event
+     * @param {string} value Value
      */
     const handleAmountFiatChange = async value => {
-      const lastPrice = currentTicker[fiatCurrency]
-      const cryptoValue = convert('fiat', cryptoUnit, value, lastPrice)
-      const doUpdate = shouldUpdateLinked.current
-      shouldUpdateLinked.current = false
-      if (doUpdate) {
+      if (!isLinkedUpdatesBlocked()) {
+        blockLinkedUpdates()
+        const lastPrice = currentTicker[fiatCurrency]
+        const cryptoValue = convert('fiat', cryptoUnit, value, lastPrice)
         formApi.setValue('amountCrypto', cryptoValue)
+        await unblockLinkedUpdates()
       }
-      // informed calls onValueChange multiple time during value updates
-      // because of masks and patterns applied on top of UI elements
-      // give value a chance to settle before enabling updates again
-      await Promise.resolve()
-      shouldUpdateLinked.current = true
       onChange && onChange()
     }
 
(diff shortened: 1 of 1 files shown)


renderer/components/UI/CurrencyFieldGroup.js
```diff
@@ -88,7 +99,7 @@ const CurrencyFieldGroup = React.forwardRef(
                 isRequired={isRequired}
                 label={cryptoLabel || <FormattedMessage {...messages.amount} />}
                 name="amountCrypto"
-                onChange={handleAmountCryptoChange}
+                onValueChange={handleAmountCryptoChange}
```
mrfelton: It's no longer possible to enter a number into the crypto amount field on the pay form. I have a feeling it may be related to this change. I think it causes some kind conflict with other onchange handlers that kick in as part of the number field handling.

renderer/components/UI/CurrencyFieldGroup.js
```diff
@@ -34,26 +34,37 @@ const CurrencyFieldGroup = React.forwardRef(
...
+      }
     }
 
     /**
      * handleAmountFiatChange - Set the amountCrypto field whenever the fiat amount changes.
      *
      * @param {Event} e Event
      */
-    const handleAmountFiatChange = e => {
+    const handleAmountFiatChange = async value => {
```
mrfelton: Async not needed here.

korhaliv:
@mrfelton are you sure you don't have "send all" enabled? In that case fields are not editable (same behavior before this PR)
Nevertheless, this issue is really tricky. I've identified some more corner cases where a circular update was happening as well as fees were not updating when pressing enabling send all (same in current next).  I've re-done impl from scratch, so re-view is required.

mrfelton:
I didn’t have send all enabled. I could type in the fiat box, but not the crypto one

korhaliv:
> I didn’t have send all enabled. I could type in the fiat box, but not the crypto one

ok, please try updated version then and let me know if the issue persists

renderer/components/UI/CurrencyFieldGroup.js
```diff
@@ -30,30 +30,57 @@ const CurrencyFieldGroup = React.forwardRef(
...
   ) => {
+    // whether update of a linked input (crypto<->fiat) is enabled
+    // this ref is used in a process of linked value calculations to
+    // circular updates
+    const shouldUpdateLinked = useRef(true)
+
     /**
      * handleAmountCryptoChange - Set the amountFiat field whenever the crypto amount changes.
      *
      * @param {Event} e Event
```
mrfelton: param doc needs updating (e -> value)

mrfelton:
Updated version does work @korhaliv 

I'm wondering though wether we can reduce the duplication here a little. What do you think about something like this:

```js
    // Whether update of a linked input (crypto<->fiat) is enabled.
    // This ref is used in a process of linked value calculations to prevent circular updates.
    const blockUpdates = useRef(true)

    // Prevent updates of linked form values.
    const blockLinkedUpdates = () => {
      blockUpdates.current = false
    }

    // Reenable updates of linked form elements.
    const unblockLinkedUpdates = async () => {
      await Promise.resolve()
      blockUpdates.current = true
    }

    // Get current block status of linked form elements.
    const isLinkedUpdatesBlocked = () => {
      return blockUpdates.current
    }

    /**
     * handleAmountCryptoChange - Set the amountFiat field whenever the crypto amount changes.
     *
     * @param {Event} value Value
     */
    const handleAmountCryptoChange = async value => {
      const lastPrice = currentTicker[fiatCurrency]
      const fiatValue = convert(cryptoUnit, 'fiat', value, lastPrice)
      if (!isLinkedUpdatesBlocked()) {
        blockLinkedUpdates()
        formApi.setValue('amountFiat', fiatValue)
      }
      unblockLinkedUpdates()
      onChange && onChange()
    }

    /**
     * handleAmountFiatChange - Set the amountCrypto field whenever the fiat amount changes.
     *
     * @param {Event} value Value
     */
    const handleAmountFiatChange = async value => {
      const lastPrice = currentTicker[fiatCurrency]
      const cryptoValue = convert('fiat', cryptoUnit, value, lastPrice)
      if (!isLinkedUpdatesBlocked()) {
        blockLinkedUpdates()
        formApi.setValue('amountCrypto', cryptoValue)
      }
      unblockLinkedUpdates()
      onChange && onChange()
    }
```

It's essentially the same code that you had but with the special logic extracted to some helper methods. I feel like this makes the change handlers themselves a little easier to read and with less duplication.

renderer/components/UI/CurrencyFieldGroup.js
```diff
@@ -30,30 +30,66 @@ const CurrencyFieldGroup = React.forwardRef(
...
-      const lastPrice = currentTicker[fiatCurrency]
-      const value = convert(cryptoUnit, 'fiat', e.target.value, lastPrice)
-      formApi.setValue('amountFiat', value)
+    const handleAmountCryptoChange = async value => {
+      if (!isLinkedUpdatesBlocked()) {
+        const lastPrice = currentTicker[fiatCurrency]
+        const fiatValue = convert(cryptoUnit, 'fiat', value, lastPrice)
+        blockLinkedUpdates()
+        formApi.setValue('amountFiat', fiatValue)
+        // informed calls onValueChange multiple time during value updates
```
mrfelton: Can we move this comment into the `unblockLinkedUpdates` method?

renderer/components/UI/CurrencyFieldGroup.js
```diff
@@ -30,30 +30,66 @@ const CurrencyFieldGroup = React.forwardRef(
...
-      const lastPrice = currentTicker[fiatCurrency]
-      const value = convert('fiat', cryptoUnit, e.target.value, lastPrice)
-      formApi.setValue('amountCrypto', value)
+    const handleAmountFiatChange = async value => {
+      if (!isLinkedUpdatesBlocked()) {
+        const lastPrice = currentTicker[fiatCurrency]
+        const cryptoValue = convert('fiat', cryptoUnit, value, lastPrice)
+        blockLinkedUpdates()
+        formApi.setValue('amountCrypto', cryptoValue)
+        // informed calls onValueChange multiple time during value updates
```
mrfelton: and then remove from here too

renderer/components/UI/CurrencyFieldGroup.js
```diff
@@ -30,30 +30,66 @@ const CurrencyFieldGroup = React.forwardRef(
...
      */
-    const handleAmountCryptoChange = e => {
-      const lastPrice = currentTicker[fiatCurrency]
-      const value = convert(cryptoUnit, 'fiat', e.target.value, lastPrice)
-      formApi.setValue('amountFiat', value)
+    const handleAmountCryptoChange = async value => {
+      if (!isLinkedUpdatesBlocked()) {
+        const lastPrice = currentTicker[fiatCurrency]
+        const fiatValue = convert(cryptoUnit, 'fiat', value, lastPrice)
+        blockLinkedUpdates()
```
mrfelton: Could I suggest moving this line up to the top of this block so that it reads more like:

```
if (!isLinkedUpdatesBlocked()) {
  blockLinkedUpdates()
  // do stuff ...
  await unblockLinkedUpdates()
}
```

korhaliv:
@mrfelton updated. i've also moved 
```
 onChange && onChange
``` 
out of `unblockLinkedUpdates`. Imo that code doesn't belong there semantically. Even though we have to call it twice now

PR #2682, opened 2019-08-03
refactor: proxymise grpc calls

korhaliv:
## Description:
Simplify interaction with grpc comlink proxy. 
Instead of
```
import { grpcService } from 'workers'
const grpc = await grpcService
const result = await grpc.service.Lightning.someStuff()
```
We can now just do
```
import { grpc } from 'workers'
const result = await grpc.service.Lightning.someStuff()
```

## Motivation and Context:
Reduce boilerplate and improve code readability

## How Has This Been Tested?
Manually

## Types of changes:
Refactor

## Checklist:

### DIFF
--- a/renderer/reducers/lnd.js
+++ b/renderer/reducers/lnd.js
@@ -1,7 +1,7 @@
 import config from 'config'
 import { createSelector } from 'reselect'
 import { proxyValue } from 'comlinkjs'
-import { grpcService } from 'workers'
+import { grpc } from 'workers'
 import { fetchInfo, setInfo } from './info'
 import { startNeutrino, stopNeutrino } from './neutrino'
 import { putWallet, removeWallet, setActiveWallet, walletSelectors } from './wallet'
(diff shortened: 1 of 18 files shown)


renderer/reducers/lnd.js
```diff
@@ -168,7 +166,7 @@ export const connectGrpcService = lndConfig => async dispatch => {
   } catch (error) {
     dispatch({ type: CONNECT_GRPC_FAILURE, error })
     // Disconnect event listeners.
-    unsubFromGrpcEvents(grpc)
+    // unsubFromGrpcEvents(grpc)
```
popenkomaksim: is that commented for purpose?

renderer/reducers/network.js
```diff
@@ -103,7 +103,7 @@ export function getDescribeNetwork() {
  */
 export const fetchDescribeNetwork = () => async dispatch => {
   dispatch(getDescribeNetwork())
-  const grpc = await grpcService
+
```
popenkomaksim: Do you need empty line here?

renderer/reducers/lnd.js
korhaliv: actually no. that's merge issue, so good catch thanks!

Issue #2100, opened 2019-04-23
Refactor: pay form


korhaliv:
## Description
Refactor pay form component. Decompose it into smaller components. Use functional component + hooks instead of class

bolatovumar:
@korhaliv I can take a stab at this. I was wondering about the "Use functional component + hooks instead of class" bit because function components don't support proptypes AFAIK. Do we want to get rid of the existing proptypes on the Pay component and whatever components will be split out of it?

korhaliv:
Functional components actually support proptypes without any issues. We have many instances of that in our codebase. For example
https://github.com/LN-Zap/zap-desktop/blob/next/renderer/components/UI/ActionButton.js
Also, be advised that this kind of PR should go against the `next` branch

bolatovumar:
@korhaliv right, I see. I don't usually use proptypes and React docs show them being used with class components only so I assumed there weren't supported with function components.

PR #2594, opened 2019-07-23
refactor: onboarding

korhaliv:
## Description:
resolves #2292

## How Has This Been Tested?
Manually

## Types of changes:
Refactor

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have reviewed and updated the documentation accordingly.
- [x] I have read the _CONTRIBUTING_ document.
- [ ] I have added tests to cover my changes where needed.
- [ ] All new and existing tests passed.
- [x] My commits have been squashed into a concise set of changes.

PR #2688, opened 2019-08-05
refactor(pay-form): decompose into smaller components

bolatovumar:
Address #2100

## Description:

Decomposed the pay form into smaller components and added unit tests and additional e2e tests.

## How Has This Been Tested?

Unit tests, e2e tests and manual testing.

## Checklist:


korhaliv:
Hey @bolatovumar . Haven't looked at the code yet, but tests are failing. Could you please look into that first?

mrfelton:
There was an issue with our testnet4 lnd node that could have been causing the failure here. I have restarted that node and I will rerun these tests on CI now

bolatovumar:
@mrfelton thanks. Tests were passing on my end but I will re-run them again to confirm locally.

bolatovumar:
@mrfelton I started going to down the path of using hooks and stuff but ran into some difficulties with how the react-spring library does hooks and since it's used quite a bit I left it as is.

You are definitely right about proptypes. I made sure proptypes that are defined are actually used but it's not obvious when you have a large list of them and they are being passed down. Would you rather have props passed down explicitly instead of doing something like `<Components {...props} />`?

mrfelton:
> I started going to down the path of using hooks and stuff but ran into some difficulties with how the react-spring library does hooks and since it's used quite a bit I left it as is.

@bolatovumar Yes, that's fair enough. I don't want to complicate this PR with a mass conversion to hooks.

> I made sure proptypes that are defined are actually used but it's not obvious when you have a large list of them and they are being passed down. Would you rather have props passed down explicitly instead of doing something like <Components {...props} />?

We generally do this more explicitly since it makes the code a little easier to read being able to see what props are being passed into a component. The way you have it now you are passing all props to all child components (in `PayPanelBody` for example). It'd be better to explicitly pass through only the props that are needed for each child component so that there are no unused props being passed into a component.

A general note regarding prop passing - generally speaking, we are trying to refactor things such that we create multiple containers and connect components to the store as low as possible rather than connecting one single high level component and passing props down from there to everything else. Obviously that may not make sense in all situations.

bolatovumar:
@mrfelton re:prop passing I did notice the pattern with multiple containers but also saw quite a few cases where the top level component accepted quite a few props and just passed them down to some child components and just went with that since most of the child components of the pay form are quite simple and have some tiny bit of top-level logic.

korhaliv:
> @mrfelton re:prop passing I did notice the pattern with multiple containers but also saw quite a few cases where the top level component accepted quite a few props and just passed them down to some child components and just went with that since most of the child components of the pay form are quite simple and have some tiny bit of top-level logic.

Yes, there are instances of such a pattern in the current codebase. However, our goal is to get rid of that as much as possible. For reference see #2594 as one of the most recent moves in this direction.

bolatovumar:
> > @mrfelton re:prop passing I did notice the pattern with multiple containers but also saw quite a few cases where the top level component accepted quite a few props and just passed them down to some child components and just went with that since most of the child components of the pay form are quite simple and have some tiny bit of top-level logic.
> 
> Yes, there are instances of such a pattern in the current codebase. However, our goal is to get rid of that as much as possible. For reference see #2594 as one of the most recent moves in this direction.

Thanks, I will take a look at that PR.

bolatovumar:
Appveyor build seems to have failed for unrelated reasons.

renderer/components/Pay/PayAddressField.js
```diff
@@ -0,0 +1,101 @@
...
+      currentStep,
+      handlePayReqChange,
+      intl,
+      isLn,
+      network,
+      redirectPayReq,
+    } = this.props
+
+    const payReq_label =
+      currentStep === PAY_FORM_STEPS.address
```
korhaliv: Can we re-write this in a more understandable way? These nested ternary operators look confusing

renderer/components/Pay/PayAmountFields.js
```diff
@@ -0,0 +1,173 @@
...
+      initialAmountFiat,
+      isOnchain,
+      isQueryingFees,
+      lndTargetConfirmations,
+    } = this.props
+
+    const formState = formApi.getState()
+    const { isCoinSweep } = formState.values
+    const amountFieldsDisplayState =
+      currentStep === PAY_FORM_STEPS.amount
```
korhaliv: Same here

renderer/components/Pay/PayPanelHeader.js
```diff
@@ -0,0 +1,38 @@
...
+
+  return (
+    <>
+      <PayHeader
+        title={
+          <>
+            <FormattedMessage {...messages.send} /> {chainName} ({cryptoUnitName})
+          </>
+        }
+        type={isLn ? 'offchain' : isOnchain ? 'onchain' : null}
```
korhaliv: and here

renderer/components/Pay/PaySummary.js
```diff
@@ -0,0 +1,105 @@
...
+          address={payReq}
+          amount={amountInSats}
+          fee={getFee()}
+          isCoinSweep={isCoinSweep}
+          lndTargetConfirmations={lndTargetConfirmations}
+          mt={-3}
+          speed={speed}
+        />
+      )
+    } else if (isLn) {
```
korhaliv: else is not needed

renderer/components/Pay/PayPanelFooter.js
```diff
@@ -0,0 +1,135 @@
...
+  if (isLn && invoice) {
+    hasEnoughFunds = amountInSats <= channelBalance
+  } else if (isOnchain) {
+    hasEnoughFunds = amountInSats <= walletBalanceConfirmed
+  }
+
+  return hasEnoughFunds
+}
+
+const nextButtonText = (formState, props) => {
```
korhaliv: can we rename this to `getNextButtonText` ? there is a local variable with the same name and in general `getX` is probably a better name for this kind of routine

renderer/components/Pay/PayAddressField.js
bolatovumar: will do

korhaliv: also please rename `payReq_label ` to a camel case `payReqLabel`

bolatovumar:
@korhaliv all your comments should be addressed now in the latest commit

PR #2696, opened 2019-08-06
feat: add ability to sign and verify messages

### LINKED ISSUE #2023: Sign messages
Zap should provide a UI for users to sign messages. This can be used to authenticate their node's public key when users use their node as an online identity

korhaliv:
## Description:
Resolves #2023 
Also adds the ability to verify messages

## How Has This Been Tested?
Manually

image
image

## Types of changes:
New feature

## Checklist:


renderer/components/Profile/ProfilePaneSignMessage/ProfilePaneSignMessage.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const formApiRef = useRef(null)
+  const notifyOfCopy = () =>
+    showNotification(intl.formatMessage({ ...messages.sig_copied_notification_description }))
+
+  const onSignMessage = async () => {
+    try {
+      const { current: formApi } = formApiRef
+      const message = formApi.getValue('message')
+      const { signature } = await grpc.services.Lightning.signMessage({ msg: Buffer.from(message) })
```
mrfelton: Should we extract this to a redux thunk rather than using the service directly in the component to abstract the service api specifics out of the component and be more consistent with all other service invocations?

renderer/components/Profile/ProfilePaneVerifyMessage/ProfilePaneVerifyMessage.js
```diff
@@ -0,0 +1,116 @@
...
+    showNotification(intl.formatMessage({ ...messages.pubkey_copied_notification_description }))
+
+  const reset = () => setInfo(null)
+
+  const onSignMessage = async () => {
+    try {
+      const { current: formApi } = formApiRef
+      const message = formApi.getValue('message')
+      const signature = formApi.getValue('signature')
+      const { valid, pubkey } = await grpc.services.Lightning.verifyMessage({
```
mrfelton: Same here

renderer/components/Profile/ProfilePaneSignMessage/ProfilePaneSignMessage.js
```diff
@@ -0,0 +1,83 @@
...
+          </Text>
+          <Bar mb={4} mt={2} />
+          <TextArea
+            css={`
+              word-break: break-all;
+            `}
+            description={intl.formatMessage({ ...messages.sign_message_desc })}
+            isRequired
+            label={intl.formatMessage({ ...messages.sign_message_label })}
+            {...rest}
```
mrfelton: How come we are applying remaining props here in addition to the container `Box`?

renderer/components/Profile/ProfilePaneSignMessage/ProfilePaneSignMessage.js
korhaliv: I'm a bit skeptical of this. That action won't do anything. Not even dispatch and IPC event or do anything that can affect something in another part of the app. Feels kinda boilerplaty or consistency for the sake of consistency. But if you insist I can add that.

mrfelton: There are a couple of advantages to doing it this way.

1) Better testability by leveraging dependency injection
2) Better component reusability for the same reason
3) Easier debugging by ensuring a consistent approach to logging (all actions logged the same way)
4) Hiding API complexity from our components (using Buffers in this case)
5) General code consistency

I don't see that there is lots of boilerplate code to do this - it's just defining and exporting a wrapper thunk and injecting it into the component with connect.

korhaliv: 1. Agree
2. Partly agree. Theoretically maybe practically in this concrete case - not so much
3. Agree but with some remarks because of pt4
4. Kinda. But leveraging redux here doesn't feel right. One can hide complexity by adding some sort of a util/wrapper. There are other ways to do dep injections too.
5. Consistency is good, but not for _just_ for the sake of it.

mrfelton: > Partly agree. Theoretically maybe practically in this concrete case - not so much

I was thinking specifically about wanting to display this component in the storybook without having to build a general mock for the grpc worker. Mocking a redux action that is injected via a prop is probably simpler.

> Kinda. But leveraging redux here doesn't feel right. 

In what way does redux not feel right? Because we may not be mutating state off the back of the action?

> One can hide complexity by adding some sort of a util/wrapper. There are other ways to do dep injections too.

Such as what? Only way I know to inject things into components is via props. Perhaps the prop doesn't have to be a redux action. It could be some other wrapper. But, in the context of our codebase I don't know of a more suitable place to wrap things like api calls than our reducers.

> Consistency is good, but not just for the sake of it.

I agree. If there were no other gains to be had I would be less inclined to change the implementation just for the sake of consistency.

korhaliv: >Mocking a redux action that is injected via a prop is probably simpler.
You can mock a prop
>In what way does redux not feel right? Because we may not be mutating state off the back of the action?
It doesn't affect the state in any way and we know that before and are not planning that it will. 
>Such as what?

korhaliv: >Mocking a redux action that is injected via a prop is probably simpler.

You can mock a prop

>In what way does redux not feel right? Because we may not be mutating state off the back of the action?

It doesn't affect the state in any way and we know that before and are not planning that it will. 

>Such as what?

Injecting a prop doesn't mean it has to be a redux action. You can use HOC, hook, render props, pass props directly after all.

As far as logging is concerned - all GRPC calls are also logged to a console

PS: as i said in the first post of this chain - if you want I can make this change, it's not something that i consider a no go no matter what, just thinking that it may be sub optimal in certain ways. This convo is more to discuss what are the pros and the cons and what are the other ways

mrfelton: > It doesn't affect the state in any way and we know that before and are not planning that it will.

We use the redux store to describe the current state of the app so that components or any other part of the app has the ability to act on that in one way or another. IMO The fact that we are choosing not to act directly on that at this point is neither here nor there - it's still good to have the store accurately reflect what's happening in the app at any given time.

> PS: as i said in the first post of this chain - if you want I can make this change, it's not something that i consider a no go no matter what, just thinking that it may be sub optimal in certain ways. This convo is more to discuss what are the pros and the cons and what are the other ways

I know, I get that. I'm just not keen on the idea of components importing and using services directly as it feels like a rather tight coupling of a component to the mechanics of a specific API call, and I'm trying to understand the objection to using redux for this as we do everywhere, and wether there is another approach that we could be taking.

korhaliv:
@mrfelton updated

renderer/components/Profile/ProfilePaneSignMessage/messages.js
```diff
@@ -0,0 +1,13 @@
+import { defineMessages } from 'react-intl'
+
+/* eslint-disable max-len */
+export default defineMessages({
+  sign_message_pane_title: 'Sign Message',
+  feature_desc: `This feature allows you to sign arbitrary messages to prove that you are in control of the node's public key. Signed message can be independently verified by the people you hand the message to.`,
```
mrfelton: I'd suggest writing this in the imperative.

ie `Sign arbitrary messages to prove that you are in control of the node's public key....`

renderer/components/Profile/ProfilePaneVerifyMessage/messages.js
```diff
@@ -0,0 +1,21 @@
+import { defineMessages } from 'react-intl'
+
+/* eslint-disable max-len */
+export default defineMessages({
+  verify_pane_title: 'Verify Message',
+  feature_desc: `Verifies a signature over a messsage. The signature must be zbase32 encoded and signed
```
mrfelton: Maybe also use imperative form here too.

ie `Verify a signature over a message...`

renderer/components/Profile/ProfilePaneSignMessage/ProfilePaneSignMessage.js
```diff
@@ -0,0 +1,83 @@
...
+  const notifyOfCopy = () =>
+    showNotification(intl.formatMessage({ ...messages.sig_copied_notification_description }))
+
+  const onSignMessage = async () => {
+    try {
+      const { current: formApi } = formApiRef
+      const message = formApi.getValue('message')
+      const { signature } = await signMessage(message)
+      setSig(signature)
+    } catch {
```
mrfelton: I think this is what's causing the test failure @korhaliv - needs to be `catch (e) {`

alevchuk:
Hi, is Sign Message also included in the iOS app and going to be available in App Store?

bolatovumar:
@alevchuk this is the desktop app repo. Feel free to ask here (iOS repo): https://github.com/LN-Zap/zap-iOS

PR #2697, opened 2019-08-06
feat(ui): add ability to specify fallback address in LN invoices

### LINKED ISSUE #2021: Add fallback address to LN invoices
Our LN invoices don't include a fallback address. The future UX for LN is to opt to pay on-chain if there are capacity issues with paying over Lightning. We should support this ourselves and encourage others to do the same. For starters, we should pass a fallback address to our newly generated LN invoices for users.

This can be accomplished by passing `fallback_addr` to the `addinvoice` API call: https://api.lightning.community/#addinvoice

korhaliv:
## Description:
Resolves #2021 

## How Has This Been Tested?
Manually

image

image

## Types of changes:
New feature/enhancement

## Checklist:


JimmyMow:
@korhaliv Why wouldn't we just input the address for them @korhaliv? I am always very cautious in complicating the UI even more than we already have to

korhaliv:
> @korhaliv Why wouldn't we just input the address for them @korhaliv? I am always very cautious in complicating the UI even more than we already have to

In that case, should it be always on? What if some doesn't want fallback and prefers to postpone payment?

JimmyMow:
@korhaliv hmm, interesting point. Why would someone want it disabled? Maybe to enforce the timeout or for privacy reasons. Maybe we enable them by default and allow a setting in preferences to turn them off? But I don't think you should have to copy and paste an address in the form. We can do that for you

korhaliv:
@JimmyMow one may not want on-chain payment for economical (future fees when you need to spend coins)  or privacy reasons and in that scenario, one may want to choose to wait until LN payment is available. There is a reason why `fallback_addr` is an optional field. It's not a problem to pre-populate that input (or use some other form. maybe readonly address field + "use fallback" toggle). The main question should imo be whether we are explicit about using fallback(e.g not adding it without the user knowing about this at all) and allow user to switch it off if he so desires.

mrfelton:
I think you have both raised good points.

My 2 stas:

- The feature should be optional - there are valid reasons why a user might not want to provide a fallback address.
- The user should not have to type/paste a bitcoin address in order to use the feature.
- The default should be to have fallback enabled and to have it populated with an address from the local wallet.
- In a separate PR, we should build the UX to handle the other side of this - enabling a payee to make use of the fallback address in the event that an LN payment fails.

Can we make this less intrusive. I feel like this takes away from the simplicity of the payment form somewhat, for a feature that most people probably shouldn't have to think about.

Perhaps we could consider adding an expandable/collapsible `advanced` section to the bottom of the form that would be collapsed by default, and move this into it (as well as the "Include routing hints"). That would enable us to provide the feature to users that want to use it, but in a way that allows us to keep a more streamlined simple interface for the majority use case. Advances users can expand the advanced section and fiddle with this if they want (I think this is what @korhaliv was suggesting too)

Alternatively, this could be done via a simple global setting in the preferences page `Enable fallback address` but in some ways this may be less flexible and harder to use than if if was available directly on the payment form.

korhaliv:
@JimmyMow @mrfelton gents re-review plz

korhaliv:
@JimmyMow @mrfelton sirs re-review plz

config/default.js
```diff
@@ -104,7 +104,7 @@ module.exports = {
     baseRetryDelay: 1000,
     retryCount: 3, // Number of retries for pay invoice failure
   },
-
+  useAddressFallback: false,
```
mrfelton: Should this be under invoices, at `invoices.useAddressFallback`?

renderer/reducers/invoice.js
```diff
@@ -168,9 +168,10 @@ export const receiveInvoices = ({ invoices }) => dispatch => {
  * @param {string} cryptoUnit Crypto unit (sats, bits, btc)
  * @param {string} memo Memo
  * @param {boolean} isPrivate Set to true to include routing hints
+ * @param {string} fallbackAddress on-chain address fallback
  * @returns {Function} Thunk
  */
-export const createInvoice = (amount, cryptoUnit, memo, isPrivate) => async (
+export const createInvoice = (amount, cryptoUnit, memo, isPrivate, fallbackAddress) => async (
```
mrfelton: Consider refactoring to accept an `options` object. This is starting to make for a hard to read caller signature.

korhaliv: good call

config/default.js
korhaliv: absolutely

renderer/components/Settings/messages.js
```diff
@@ -32,6 +32,8 @@ export default defineMessages({
   address_option_np2wkh: 'Segwit',
   blockExplorer_label: 'Block explorer',
   blockExplorer_description: 'Your preferred block explorer.',
+  fallbackAddress_label: 'Enable on-chain address fallback',
+  fallbackAddress_description: 'Add fallback on-chain address to LN invoices.',
```
mrfelton: Can we reword this to `Add on-chain fallback address to LN invoices.`

"on-chain fallback address"
vs
"fallback on-chain address" 

We know it's a fallback address, the question is what type of fallback address. The type qualifier should come first.

renderer/containers/Request.js
```diff
@@ -14,12 +16,15 @@ const mapStateToProps = state => ({
   isProcessing: state.invoice.invoiceLoading,
   payReq: state.invoice.invoice,
   invoice: invoiceSelectors.invoice(state),
+  willUseFallback: settingsSelectors.currentConfig(state).useAddressFallback,
```
mrfelton: Shouldn't this be `.invoices.useAddressFallback`?

korhaliv: it should

PR #2706, opened 2019-08-07
fix: ensure timeout is cleared asap in promiseTimeout

korhaliv:
## Description:
Fixes corner case with `promiseTimeout` which prevents node.js process from exiting because of hanging timeout. This PR ensures timeout is cleared as soon as promise race resolves

## How Has This Been Tested?
Manually

## Types of changes:
Bugfix

## Checklist:

### DIFF
--- a/utils/promiseTimeout.js
+++ b/utils/promiseTimeout.js
@@ -3,19 +3,20 @@
  *
  * @param {number} ms Timeout (ms)
  * @param {Promise} promise Promise to timeout
+ * @param {string} message messaged passed to the reject promise
  * @returns {Promise} Promise that rejects in <ms> milliseconds
  */
-const promiseTimeout = function(ms, promise) {
+export default function(ms, promise, message = 'Timed out') {
   // Create a promise that rejects in <ms> milliseconds
-  let timeout = new Promise((resolve, reject) => {
-    let id = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


utils/promiseTimeout.js
```diff
@@ -5,17 +5,19 @@
...
-const promiseTimeout = function(ms, promise) {
+export default function(ms, promise) {
+  let timerId
   // Create a promise that rejects in <ms> milliseconds
   let timeout = new Promise((resolve, reject) => {
-    let id = setTimeout(() => {
-      clearTimeout(id)
+    timerId = setTimeout(() => {
+      clearTimeout(timerId)
       reject(`Timed out.`)
```
mrfelton: Let's also fix this to reject with an Error rather than a string.

utils/promiseTimeout.js
```diff
@@ -5,17 +5,19 @@
...
     }, ms)
   })
 
   // Returns a race between our timeout and the passed in promise
-  return Promise.race([promise, timeout])
+  return Promise.race([promise, timeout]).then(value => {
+    timerId && clearTimeout(timerId)
```
mrfelton: Should we also clear this in the event that the promise rejects? (in catch block)

korhaliv: absolutely

PR #2720, opened 2019-08-09
 feat(wallet): add ability to export on-chain tx invoices

### LINKED ISSUE #1290: I should be able to download a PDF version of a payment/transaction receipt
## Detailed Description

Add ability to download/print transaction details as a PDF

## Possible Implementation

Add `Download PDF` icon to ActivityModal's that allow for downloading details of a transaction in PDF format

image

korhaliv:
## Description:
Resolves #1290 

## How Has This Been Tested?
Manually

image
image
image

## Types of changes:
New feature

## Checklist:


mrfelton:
I'd go with `Download complete`.

Also @korhaliv can we set the default download location to the user's home directory or documents directory?

korhaliv:
> I'd go with `Download complete`.
> 
> Also @korhaliv can we set the default download location to the user's home directory or documents directory?

This location differs for macOS, Win and Linux.  So we will need to use something like this https://github.com/kingjan1999/platform-folders to get access to it

mrfelton:
@korhaliv  I'd probably be more inclined to just default to the home directory -which we can get quite easily - rather than adding an additional dependency that relies on a native binary to get the path to the documents dir

korhaliv:
> @korhaliv I'd probably be more inclined to just default to the home directory -which we can get quite easily - rather than adding an additional dependency that relies on a native binary to get the path to the documents dir

i don't think this is actually useful for windows users. Documents folder is accessible in 1 click from any location when using system dialog. And users folder itself isn't a popular place amongst win users. So either it should be mac specific or i suggest leaving it as it is now.

mrfelton:
The reason I suggested altering it is because when I first tested it I wasn't sure where it put the file and couldn't find it - it was my fault for not checking what the download location was set to.

I just retested it again now and in fact it does seem to default to the home directory on the mac. Do you know where it defaults to on Windows / Linux?

korhaliv:
> The reason I suggested altering it is because when I first tested it I wasn't sure where it put the file and couldn't find it - it was my fault for not checking what the download location was set to.
> 
> I just retested it again now and in fact it does seem to default to the home directory on the mac. Do you know where it defaults to on Windows / Linux?

I believe the last location. It's shared between system save dialogs

mrfelton:
When I retested, I moved my Electron dir out of the way and started over fresh - and it did default to the home dir. So unless it stores that history elsewhere then it does seem to default to the home dir (on mac at least)

korhaliv:
@mrfelton ready for re-test

PR #2727, opened 2019-08-12
refactor: migrate to react-intl v3

korhaliv:
## Description:
Migrate to react-intl v3. In our case main breaking changes were:
Removal of `addLocaleData` 
Removal of `intlShape`
Removal of `FormattedRelative`

## How Has This Been Tested?
Manually 

## Types of changes:
Refactor

## Checklist:


i18n/intjShape.js
```diff
@@ -0,0 +1,4 @@
+import PropTypes from 'prop-types'
```
mrfelton: The name of this file is `intjShape` - should probably be `intlShape`

renderer/components/Activity/ActivityActions/ActivityActions.js
```diff
@@ -1,12 +1,13 @@
...
 import PropTypes from 'prop-types'
-import { injectIntl, intlShape } from 'react-intl'
+import { injectIntl } from 'react-intl'
 import { Flex } from 'rebass'
 import { Card } from 'components/UI'
 import ActivityFilter from './ActivityFilter'
 import ActivityRefresh from './ActivityRefresh'
 import ActivitySearch from './ActivitySearch'
 import messages from './messages'
+import { intlShape } from '@zap/i18n'
```
mrfelton: import should go before relative imports (this comment applies to the whole PR)

i18n/index.js
```diff
@@ -2,6 +2,9 @@ import get from 'lodash/get'
 import { lookup } from 'country-data-lookup'
 import createDebug from 'debug'
 import isRenderer from 'is-electron-renderer'
+import PropTypes from 'prop-types'
+
+export const intlShape = PropTypes.instanceOf(Object)
```
mrfelton: Why is this defined as a simple `Object` type here? Shouldn't it be `IntlShape` as per the seemingly unused `i18n/intjShape.js` file?

stories/components/countdown.stories.js
```diff
@@ -8,22 +8,16 @@ storiesOf('Components', module).addWithChapters('Countdown', {
       sections: [
         {
           title: 'Default',
-          sectionFn: () => <Countdown date={Math.round(new Date() / 1000) + 5000} />,
+          sectionFn: () => <Countdown date={5} />,
```
mrfelton: `date={5}` seems quite non-explicit. 5 what? seconds? 5 months? I think if we want to change the way that this works then we should rename the prop to something that describes what is expected. `5` is not a `date` which is what I would expect to pass in to a prop named `date`

korhaliv: updated

i18n/index.js
korhaliv: resolved

renderer/components/Activity/ActivityActions/ActivityActions.js
korhaliv: updated

renderer/components/Request/RequestSummary.js
```diff
@@ -1,52 +1,29 @@
...
-      schedulerInstance.addTask({
-        task: refreshIsExpired,
-        baseDelay: 1000,
-      })
-    }
-    // Clear scheduler on unmount.
+    setExpiryDelta(decodedInvoice.timeExpireDate - getUnixTime() / 1000)
     return () => {
-      schedulerInstance.removeAllTasks()
+      setIsExpired(false)
```
mrfelton: Why the need to update the state on unmount?

PR #2735, opened 2019-08-13
fix(autopay): correctly apply blue theme overrides

mrfelton:
## Description:

Correctly apply blue theme overrides

## Motivation and Context:

App crashes when trying to access autopay feature.

## How Has This Been Tested?

Enable autopay feature and try to navigate to that part of the wallet ui.

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/renderer/components/Autopay/Autopay.js
+++ b/renderer/components/Autopay/Autopay.js
@@ -12,7 +12,7 @@ import AutopayActions from './AutopayActions'
 import messages from './messages'
 
 const customiseTheme = theme => {
-  return createThemeVariant('autopilot', {
+  return createThemeVariant('autopay', {
     colors: {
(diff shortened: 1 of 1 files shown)


renderer/components/Autopay/Autopay.js
```diff
@@ -13,8 +13,10 @@ import messages from './messages'
 
 const customiseTheme = theme => {
   return createThemeVariant('autopilot', {
```
korhaliv: Not related to this PR obviously, but i wonder why do we call theme `autopilot`. Should be `autopay`

mrfelton: Good catch. This isn't the first time this typo has slipped its way in somewhere. Fixed.

PR #2767, opened 2019-08-16
fix(ui): form input name and autofocus

mrfelton:
## Description:

Set missing input props
Ensure autofocus works
Add autofocus examples in storybook

## Motivation and Context:

We set TextArea `name` based on field but fail to do so in the `Input` component.

## How Has This Been Tested?

Manually

## Types of changes:

Refactor

## Checklist:

### DIFF
--- a/renderer/components/Form/TextArea.js
+++ b/renderer/components/Form/TextArea.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react'
+import React, { useLayoutEffect, useState } from 'react'
 import PropTypes from 'prop-types'
 import { compose } from 'redux'
 import { asField } from 'informed'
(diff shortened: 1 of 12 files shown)


renderer/components/Form/TextArea.js
```diff
@@ -13,6 +13,12 @@ import { createSystemInput } from './util'
 // Create an html input element that accepts all style props from styled-system.
 export const SystemTextArea = createSystemInput('textarea')
 
+const maybeFocusRef = ref => {
+  if (ref.current && document.activeElement !== ref.current) {
```
korhaliv: Looks like the code from `Input.js` consider extracting as an util

renderer/components/Form/TextArea.js
```diff
@@ -37,16 +43,13 @@ const TextArea = ({
...
   const [spaceProps, otherProps] = extractSpaceProps(rest)
   const { setValue, setTouched } = fieldApi
   const { value, maskedValue, error, asyncError } = fieldState
 
-  useEffect(() => {
-    if (willAutoFocus) {
-      inputRef.current.focus()
-    }
-  }, [willAutoFocus])
+  useLayoutEffect(() => {
```
korhaliv: Could be an even better idea to extract the hook

mrfelton: Good call. Updated.

PR #2811, opened 2019-08-27
 feat(ui): warn of potential liquidity issues when creating/paying invoices

korhaliv:
## Description:
Currently, we show the user's total inbound/outgoing capacity, which is slightly misleading since the real max for send/receive depends on the user's channel with the largest local or remote balance respectively. This PR adds non-blocking validation to the corresponding forms

## How Has This Been Tested?
Manually

image
image

## Types of changes:
Enhancement

## Checklist:


renderer/components/Pay/PayPanelFooter.js
```diff
@@ -49,6 +50,22 @@ const getNextButtonText = (formState, props) => {
   return nextButtonText
 }
 
+const getLiquidityWarningMessage = (hasEnoughFunds, capacity, unit) => {
```
mrfelton: I find the logic of this method a little confusing...

If the user doesn't have enough funds it returns a "not enough funds error". Otherwise it returns an "one time send capacity" error.

The reason for the first error is obvious since its handled by `if (!hasEnoughFunds) {` but it's not obvious why it defaults to the "one time send capacity" error otherwise.

I'd suggest breaking this out into 2 separate methods and calling slightly more explicitly:

```js
if (!hasEnoughFunds) {
  getNotEnoughFundsError()
} else if (isAboveMax) {
  getAboveMaxSendCapacityError(capacity, unit))
}
```

renderer/components/Request/Request.js
```diff
@@ -30,6 +31,7 @@ class Request extends React.Component {
     invoice: PropTypes.object,
     isAnimating: PropTypes.bool,
     isProcessing: PropTypes.bool,
+    maxOneTimeReceive: PropTypes.number.isRequired,
```
mrfelton: Since this prop has been marked as required we're going to get warnings in the storybook for not setting it. Consider making non-required. Alternatively ensure that we set it everywhere where `Receive` is used (storybook).

renderer/components/Request/Request.js
```diff
@@ -46,6 +48,17 @@ class Request extends React.Component {
...
 
+  /**
+   * getMaxOneTimeReceive - User's max one time receive capacity in satoshis.
+   *
+   * @returns {number} user's max one time receive capacity in satoshis
+   * @memberof Request
+   */
+  getMaxOneTimeReceive() {
+    const { cryptoUnit, maxOneTimeReceive } = this.props
+    return convert('sats', cryptoUnit, maxOneTimeReceive)
```
mrfelton: Comment says that it returns amount in satoshi, but this looks to be returning in `cryptoUnit`

renderer/components/Request/Request.js
mrfelton: Actually, this is not an issue since we use the connected container in the storybook and not the raw component.

renderer/components/Pay/PayPanelFooter.js
korhaliv: i've redone that entire part from scratch

renderer/components/Request/Request.js
korhaliv: updated

PR #2842, opened 2019-09-03
fix(ui): ensure select value from items list

mrfelton:
## Description:

Ensure select value from items list

## Motivation and Context:

Fixes an issue in which it's possible to type an invalid value into a select input,.

eg, prevent this:

image

## How Has This Been Tested?

Storybook

## Types of changes:

Fix

## Checklist:

### DIFF
--- a/renderer/components/Form/Select.js
+++ b/renderer/components/Form/Select.js
@@ -165,16 +165,28 @@ const Select = props => {
         openMenu,
         toggleMenu,
       }) => {
+        /**
+         * getInitialValue - Get the initial selected value.
+         *
+         * @returns {string} Initial selected value.
+         */
         const getInitialValue = () => {
           if (selectedItem) {
             return selectedItem.value
           }
-
           if (initialSelectedItem) {
-            initialSelectedItem.value
+            return initialSelectedItem.value
           }
-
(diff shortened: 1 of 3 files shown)


renderer/components/Form/Select.js
```diff
@@ -165,18 +165,30 @@ const Select = props => {
...
+         *
+         * @returns {string} Initial selected value.
+         */
         const getInitialValue = () => {
           if (selectedItem) {
             return selectedItem.value
           }
-
           if (initialSelectedItem) {
             initialSelectedItem.value
```
korhaliv: what's this? probably `return` is missing?

renderer/components/Form/Select.js
```diff
@@ -165,18 +165,30 @@ const Select = props => {
...
+         * getValue - Get the current selected value.
+         *
+         * @returns {string} Current selected value.
+         */
+        const getValue = () => {
+          return selectedItem ? selectedItem.value : ''
+        }
+
         return (
           <div style={{ position: 'relative' }}>
```
korhaliv: this probably needs migration to sx

mrfelton:
Updated

Issue #1387, opened 2019-01-18
security for wallet (pin / password)


ad2003:
It would be great if you would have a pin securing your zap desktop wallet for btcpay server. At the moment you get access to the complete lightning node, by just opening Zap Desktop without any security.

PR #2860, opened 2019-09-09
feat(app): basic password authentication

mrfelton:
## Description:

This PR adds basic password authentication for the app. The password is stored in plain text in the indexed database so it's not a strong protection, however it does provide a useful basic level of protection.

This is currently an opt-in feature that users can enable from the preferences page

Additional things we should consider:

1. Store password in more secure way (eg, in the keychain)
2. Require users to enter existing password in order to change password
3. Enable password to grant access to all wallets
4. Auto-lock app after a time limit

## Motivation and Context:

Fix https://github.com/LN-Zap/zap-desktop/issues/1387

## How Has This Been Tested?

Manually

image

image

## Types of changes:

Feature

## Checklist:


renderer/components/Settings/SettingsFieldsSecurity.js
```diff
@@ -0,0 +1,48 @@
...
+import { useIntl } from 'react-intl'
+import { DataRow } from 'components/UI'
+import { PasswordInput, Toggle } from 'components/Form'
+import { FieldLabel } from './SettingsFieldHelpers'
+import messages from './messages'
+
+const SettingsFieldsSecurity = ({ currentConfig }) => {
+  const { value: isPasswordActive } = useFieldState('password.active')
+  const { submits } = useFormState()
+  const shouldValidateInline = submits > 0
```
korhaliv: In some places it's `willValidateInline` is there any naming scheme behind these?

renderer/containers/Login/Login.js
```diff
@@ -0,0 +1,35 @@
...
+  return <Redirect to="/init" />
+}
+
+WrappedLogin.propTypes = {
+  isLoggedIn: PropTypes.bool,
+  isPasswordEnabled: PropTypes.bool,
+}
+
+const mapStateToProps = state => ({
+  isPasswordEnabled: settingsSelectors.currentConfig(state).password.active,
```
korhaliv: Consider introducing a standalone selector for this

renderer/containers/Root.js
```diff
@@ -53,11 +67,31 @@ const Root = ({
         setMounted(true)
         await initDatabase()
         await initSettings()
-        await initTheme()
+        initTheme()
```
korhaliv: maybe it's worth to introduce a composite init action outside of the `Root` component to avoid growing prop list ?

renderer/reducers/account.js
```diff
@@ -0,0 +1,154 @@
...
+/**
+ * initAccounts - Fetch the current account info from the database and save into the store.
+ * Should be called once when the app first loads.
+ *
+ * @returns {Function} Thunk
+ */
+export const initAccount = () => async (dispatch, getState) => {
+  dispatch({ type: INIT_ACCOUNT })
+  try {
+    // Auto login user of password feature is diabled.
```
korhaliv: typo diabled

korhaliv: typo of

renderer/components/Settings/SettingsForm.js
```diff
@@ -18,6 +19,11 @@ const SettingsForm = ({
 }) => {
   const handleSubmit = async values => {
     try {
+      // If password feature has been disabled, rest the password to null.
```
korhaliv: rest -> reset

renderer/components/Settings/SettingsPage.js
```diff
@@ -11,27 +11,44 @@ import ZapLogo from 'components/Icon/ZapLogo'
 import SettingsForm from 'containers/Settings/SettingsForm'
 import SettingsFieldsWallet from './SettingsFieldsWallet'
 import SettingsFieldsGeneral from './SettingsFieldsGeneral'
+import SettingsFieldsSecurity from './SettingsFieldsSecurity'
 import messages from './messages'
 
-const SettingsMenu = ({ group, setGroup, ...rest }) => {
-  const items = [
+const SettingsMenu = ({ group, setGroup, isLoggedIn, ...rest }) => {
+  // Items accessle to unauthenticated users.
```
korhaliv: accessle

renderer/components/Login/Login.js
```diff
@@ -0,0 +1,72 @@
...
+              <Box mx="auto" width={5 / 8}>
+                <Flex justifyContent="center" mb={4}>
+                  <ZapLogo height={34} width={34} />
+                </Flex>
+                <Text mb={4} textAlign="center">
+                  <FormattedMessage {...messages.intro} />
+                </Text>
+                {error && (
+                  <Flex alignItems="center" flexDirection="column">
+                    <Message mb={3} variant="error" width="auto">
```
korhaliv: Why do we need to set `width` explicitly to `auto` ?

renderer/components/Settings/SettingsPage.js
```diff
@@ -108,6 +125,7 @@ const SettingsPage = ({ currentConfig, ...rest }) => {
         <SettingsForm>
           {group === 'general' && <SettingsFieldsGeneral currentConfig={currentConfig} />}
           {group === 'wallet' && <SettingsFieldsWallet currentConfig={currentConfig} />}
+          {group === 'security' && <SettingsFieldsSecurity currentConfig={currentConfig} />}
```
korhaliv: since this list keeps growing and all components share the same prop list, consider converting this into a map of `<group, component>` to reduce boilerplate code.

renderer/components/Settings/SettingsFieldsSecurity.js
mrfelton: We have typically have called this `shouldValidateInline`, however:

1) This fails our `react/boolean-prop-naming` lint rule
2) `will` is semantically more correct that `should` since the former is imperative statement rather than `should` which has a a probability associated with it.

I've added a refactor commit to convert all instances of `shouldValidateInline` to `willValidateInline`

renderer/components/Login/Login.js
```diff
@@ -0,0 +1,72 @@
...
+import ArrowRight from 'components/Icon/ArrowRight'
+import ZapLogo from 'components/Icon/ZapLogo'
+import messages from './messages'
+
+const Login = ({ login, loginError, clearLoginError, ...rest }) => {
+  const formApiRef = useRef(null)
+
+  useEffect(() => {
+    const { current: formApi } = formApiRef
+    loginError && formApi.setFormError(loginError)
```
korhaliv: if `loginError` is not set, do we need to call clear ?
maybe frame this two into `if (loginError)...`

renderer/components/Login/Login.js
```diff
@@ -0,0 +1,72 @@
...
+  const handleSubmit = ({ password }) => login(password)
+
+  return (
+    <Form getApi={api => (formApiRef.current = api)} onSubmit={handleSubmit} width={1}>
+      {({ formState: { submits, error } }) => {
+        const willValidateInline = submits > 0
+        return (
+          <CenteredContent {...rest} mx="auto" width={11 / 16}>
+            <Box px={0} py={4} width={1} {...rest}>
+              <Box mx="auto" width={5 / 8}>
```
korhaliv: Can we use one `Flex` instead ? (to center if it's needed)

renderer/components/Login/Login.js
```diff
@@ -0,0 +1,72 @@
...
+                <Flex justifyContent="center" mb={4}>
+                  <ZapLogo height={34} width={34} />
+                </Flex>
+                <Text mb={4} textAlign="center">
+                  <FormattedMessage {...messages.intro} />
+                </Text>
+                {error && (
+                  <Flex alignItems="center" flexDirection="column">
```
korhaliv: Maybe center everything at the top level container to keep JSX more concise?

renderer/components/Settings/SettingsForm.js
```diff
@@ -18,6 +19,11 @@ const SettingsForm = ({
 }) => {
   const handleSubmit = async values => {
     try {
+      // If password feature has been disabled, reset the password to null.
+      if (!get(values, 'password.value')) {
+        values.password.value = null
```
korhaliv: Should it be `null` or removed all together with `delete` ?

renderer/components/Settings/SettingsPage.js
```diff
@@ -11,27 +11,44 @@ import ZapLogo from 'components/Icon/ZapLogo'
...
+    {
+      id: 'security',
+      title: <FormattedMessage {...messages.fieldgroup_security} />,
+      onClick: () => setGroup('security'),
+    },
   ]
 
+  let items = [...anonItems]
+  if (isLoggedIn) {
+    items = [...items, ...authItems]
```
korhaliv: consider just doing `items = items.concat(authItems)` or `items.push(...authItems)` so it's cleared that we just adding authItems here

renderer/components/Settings/SettingsForm.js
mrfelton: It has to be set back to null. Only keys that exist in the values object will be applied as config overrides when the form is saved.

mrfelton:
Updated

mrfelton:
Closing in favour of https://github.com/LN-Zap/zap-desktop/pull/2940

PR #2882, opened 2019-09-19
feat(ui): add ability to configure neutrino nodes for local wallets

korhaliv:
## Description:
This PR adds ability to edit list of neutrino-compliant nodes local wallet connects to

## Motivation and Context:
Currently neutrino peers are hardcoded (to Zap hq nodes). If for any reason they are down wallet becomes not functional. Using user defines node list also improves privacy.

## How Has This Been Tested?
Manually

image

## Types of changes:
New feature

## Checklist:


renderer/components/Home/WalletSettingsFormLocal.js
```diff
@@ -48,6 +61,32 @@ class WalletSettingsFormLocal extends React.Component {
...
           <DataRow left={<FormattedMessage {...messages.chain} />} py={2} right={chain} />
           <DataRow left={<FormattedMessage {...messages.network} />} py={2} right={network} />
+
+          <Bar variant="light" />
+          <DataRow
+            left={<FieldLabel itemKey="neutrinoUrl" tooltip="neutrinoUrl_tooltip" />}
+            right={
+              <TextArea
+                field="neutrinoNodes"
+                format={value => value.join('\n')}
```
mrfelton: Consider extracting this method

renderer/components/Home/WalletSettingsFormLocal.js
```diff
@@ -48,6 +61,32 @@ class WalletSettingsFormLocal extends React.Component {
...
+                highlightOnValid={false}
+                initialValue={config.lnd.neutrino[chain][network]}
+                parse={value => {
```
mrfelton: also extract parse method

renderer/components/Settings/SettingsFieldHelpers.js
```diff
@@ -1,28 +1,5 @@
...
-    </Box>
-  )
-}
-
-FieldLabel.propTypes = {
-  itemKey: PropTypes.string.isRequired,
-}
+import { IntegerInput } from 'components/Form'
 
 export const PositiveIntegerField = props => (
```
mrfelton: Perhaps we should move `PositiveIntegerField` into `/components/Form`

renderer/components/Home/WalletSettingsFormLocal.js
```diff
@@ -48,6 +61,32 @@ class WalletSettingsFormLocal extends React.Component {
 
           <DataRow left={<FormattedMessage {...messages.chain} />} py={2} right={chain} />
           <DataRow left={<FormattedMessage {...messages.network} />} py={2} right={network} />
+
+          <Bar variant="light" />
+          <DataRow
```
mrfelton: Lets move this whole section to just above or below the autopilot settings and give the section a heading like `Connections` or `Advanced`

renderer/components/Home/WalletSettingsFormLocal.js
```diff
@@ -48,6 +61,32 @@ class WalletSettingsFormLocal extends React.Component {
...
+            left={<FieldLabel itemKey="neutrinoUrl" tooltip="neutrinoUrl_tooltip" />}
+            right={
+              <TextArea
```
mrfelton: Can we add some kind of validation to this textarea to ensure that what is entered is a contactable address? Perhaps use `dns.lookup` like we do when validation remove node addresses?

Otherwise, I can enter something like `w` on a newline and then the wallet fails to start with an error like `Unable to initialize neutrino backend: unable to create neutrino light client: lookup w: no such host`

mrfelton: Also, I can do something like:
```
mainnet1-btcd.zaphq.io
mainnet2-btcd.zaphq.io
mainnet2-btcd.zaphq.io
```

And end up with a startup command like `--neutrino.connect=mainnet1-btcd.zaphq.io --neutrino.connect=mainnet2-btcd.zaphq.io --neutrino.connect=mainnet2-btcd.zaphq.io`. Whilst this is valid, I'm not sure that it's desirable. Might be an idea to validate against that sort of things

renderer/components/Home/WalletSettingsFormLocal.js
```diff
@@ -27,6 +34,12 @@ class WalletSettingsFormLocal extends React.Component {
     })
   }
 
+  resetNeutrinoSettings = () => {
+    const { formApi, wallet } = this.props
+    const { chain, network } = wallet
+    formApi.setValue('neutrinoNodes', config.lnd.neutrino[chain][network])
```
mrfelton: The behaviour here is a little different to other elements of the wallet config and potentially problematic. The way that the autopilot settings work for example, is as default overrides - meaning that if you have customised the settings then they are saved in the database but if you have not or you have set them back to the defaults then they are removed from the database.

That ensures that if we update our defaults in a future release, users will automatically assume those new defaults unless they have specifically overridden the settings and set them to some non-default values.

However with the neutrino configs they are saved into the database and never removed again even after restoring the defaults. This could be problematic as it could mean that users that have simply changed the name of their wallet (for example) would no longer receive changes to our default configs if we update them for a future release (since they would still have hard coded overrides in their database) and they could be left stuck with a connection to a retired node.

renderer/components/Home/WalletSettingsFormLocal.js
korhaliv: done

renderer/components/Home/WalletSettingsFormLocal.js
korhaliv: done

renderer/components/Settings/SettingsFieldHelpers.js
korhaliv: done

renderer/components/Home/WalletSettingsFormLocal.js
korhaliv: done

renderer/components/Home/WalletSettingsFormLocal.js
korhaliv: Validation is quite tricky because it is async. But believe i made a lot of progress towards this. Check the updated PR

renderer/components/Home/WalletSettingsFormLocal.js
korhaliv: Made a number of improvements towards this goal

renderer/components/Home/WalletLauncher.js
```diff
@@ -7,12 +7,16 @@ import { withRouter } from 'react-router-dom'
 import { Box, Flex } from 'rebass/styled-components'
 import styled from 'styled-components'
 import parse from 'lndconnect/parse'
+import { isEqual } from 'lodash'
```
mrfelton: This should use `import isEqual from 'lodash/isEqual` syntax in order to prevent bundle bloat.

renderer/components/Home/WalletSettingsFormLocal.js
```diff
@@ -1,15 +1,60 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import config from 'config'
 import { compose } from 'redux'
 import { withFormApi, withFormState } from 'informed'
 import { FormattedMessage, injectIntl } from 'react-intl'
 import { Box, Flex } from 'rebass/styled-components'
+import { uniq } from 'lodash'
```
mrfelton: Also here - `import uniq from 'lodash/uniq'`

PR #2936, opened 2019-10-06
fix(ui): set delete wallet header width to same as content width

### LINKED ISSUE #2926: Design issue: Position of the title in the "Delete wallet" view too far to the left
If the title text is too long it starts too far to the left. See here in my printscreen: http://prntscr.com/peuj2s
"Bist du sicher..." and "werden soll?" is too far to the left. 

Desktop version. Windows. 

0.5.3-beta

thx

bolatovumar:
## Motivation and Context:

fix #2926

## How Has This Been Tested?

Manually

Screen Shot 2019-10-06 at 2 17 40 PM

## Checklist:

### DIFF
--- a/renderer/components/Channels/ChannelCloseDialog.js
+++ b/renderer/components/Channels/ChannelCloseDialog.js
@@ -53,7 +53,7 @@ const DialogWrapper = ({ intl, isForceClose, isOpen, onClose, onCancel, csvDelay
       <Form onSubmit={handleSubmit}>
         <Dialog buttons={buttons} header={header} onClose={onCancel} width={640}>
           <Flex alignItems="center" flexDirection="column">
-            <Text color="gray" mb={2} textAlign={isForceClose ? 'left' : 'center'} width={500}>
+            <Text color="gray" mb={2} textAlign={isForceClose ? 'left' : 'center'}>
               <FormattedMessage
                 {...(isForceClose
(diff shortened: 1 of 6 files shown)


bolatovumar:
@mrfelton I set the width instead of padding because the content below the header also sets the width but not the padding. Should I change both to use padding instead of width?

mrfelton:
> Should I change both to use padding instead of width?

No, I don't think anything should be set on both - that's kinda the point I'm making. We don't want to have to edit the value in 2 places in order to set the width correctly.

So rather than settings the additional width when creating the dialog instance, why not edit the `Dialog` component directly so that it always has some padding.

e.g.

```diff
diff --git a/renderer/components/UI/Dialog.js b/renderer/components/UI/Dialog.js
index 32899bad1..85fda6197 100644
--- a/renderer/components/UI/Dialog.js
+++ b/renderer/components/UI/Dialog.js
@@ -32,6 +32,7 @@ const Dialog = ({ header, onClose, buttons, width, children }) => {
         flexDirection="column"
         justifyContent="center"
         pb={4}
+        px={3}
       >
         <Flex alignItems="center" flexDirection="column" justifyContent="flex-start">
           {headerLayout}
```

That way, you just set the width once when creating ta dialog instance and it will always have padding of `3` regardless of the width.

bolatovumar:
@mrfelton ok, my bad, didn't read your comment correctly.

PR #2940, opened 2019-10-07
feat(app): basic password authentication

### LINKED ISSUE #1387: security for wallet (pin / password)

korhaliv:
## Description:
Continues #2860 

This PR adds a number of improvements:
1. Stores hashes of password in OS secure storage (keychain on macOS, credentials manager on Windows, gnome keyring on Ubuntu)
2. Adds ability to change password via dialog
3. Adds ability to disable/enable password via dialog
4. Adds infrastructure for password protected functionality (password prompt)
Resolves #1387

## How Has This Been Tested?
Manually

image
image
image

## Types of changes:
New feature

## Checklist:


renderer/reducers/modal.js
```diff
@@ -139,6 +168,14 @@ const ACTION_HANDLERS = {
   [SET_MODALS]: (state, { modals }) => {
     state.modals = modals
   },
+
+  [OPEN_DIALOG]: (state, { id }) => {
```
mrfelton: Lets add reducer snapshot tests for these new actions

renderer/reducers/modal.js
```diff
@@ -39,6 +42,32 @@ const createModalData = (type, options) => ({
 // Actions
 // ------------------------------------
 
+/**
+ * openDialog - Open dialog.
+ *
+ * @param  {string} id Name of modal to open
+ * @returns {Function} Thunk
+ */
+export const openDialog = id => dispatch => {
```
mrfelton: What do you think about moving these dialog methods to a new `dialog` reducer? (I know that these particular dialogs are technically modals too, but they don't live in the modal stack and aren't related to the non-dialog modals)

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+ * requirePassword - Password protect routine. Should be placed before protected code.
+ *
+ * @param {string} password current password.
+ * @returns {Promise} Promise that fulfills after login attempt (either successful or not)
+ */
+const requirePassword = password => dispatch => {
+  return new Promise((resolve, reject) => {
+    dispatch(send('getPassword'))
+    // compare hash received from the main thread to a hash of a password provided
+    window.ipcRenderer.once('getPassword', async (event, { password: hash }) => {
```
mrfelton: Probably should attach handler before dispatching the message

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+    )
+    dispatch(closeDialog(PASSWORD_SET_DIALOG_ID))
+    dispatch(showNotification(intl.formatMessage(messages.account_password_enabled)))
+  } catch (error) {
+    dispatch({ type: LOGIN_FAILURE, error: error.message })
+  }
+}
+
+/**
+ * enablePassword - Disables app-wide password protection.
```
mrfelton: s/enablePassword/disablePassword/

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+/**
+ * enablePassword - Enables app-wide password protection.
+ *
+ * @param {string} password to be used further on.
+ * @returns {Function} Thunk
+ */
+export const enablePassword = ({ password }) => async dispatch => {
+  try {
+    const intl = getIntl()
+    dispatch(setPassword(password))
```
mrfelton: Should probably `await` here since `setPassword` is async

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+    dispatch(
+      saveConfigOverrides({
```
mrfelton: This can be simplified using `putConfig`:

```js
putConfig('password.active', true)
```

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+}
+
+/**
+ * setPassword - Updates wallet password.
+ *
+ * @param {string} password new password
+ * @returns {Function} Thunk
+ */
+const setPassword = password => async dispatch => {
+  dispatch(send('setPassword', { password: await sha256digest(password) }))
```
mrfelton: We should probably attach handlers for `setPasswordSuccess` and `setPasswordFailure` and resolve or reject this method accordingly based on the responses from those.

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+ * enablePassword - Disables app-wide password protection.
+ *
+ * @param {string} password current password.
+ * @returns {Function} Thunk
+ */
+export const disablePassword = ({ password }) => async dispatch => {
+  try {
+    const intl = getIntl()
+    await dispatch(requirePassword(password))
+    dispatch(send('deletePassword'))
```
mrfelton: Here we are sending the message directly to the electron process and not waiting to know the result. If `deletePassword` fails for some reason we would end up with an inaccurate state.

Perhaps we should create a wrapper for `deletePassword` as we do for `setPassword` and handle the responses within that method.

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+    dispatch(
+      saveConfigOverrides({
```
mrfelton: Can also use `putConfig`

renderer/reducers/account.js
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * requirePassword - Password protect routine. Should be placed before protected code.
+ *
+ * @param {string} password current password.
+ * @returns {Promise} Promise that fulfills after login attempt (either successful or not)
+ */
+const requirePassword = password => dispatch => {
+  return new Promise((resolve, reject) => {
+    dispatch(send('getPassword'))
```
mrfelton: I'd suggest extracting this to a `getPassword` method which sends the message and listens/handles the relevant responses.

renderer/components/Settings/Security/hooks.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * useLoginError - Sets loginError with informed form and calls `clearLoginError` afterwards
+ * to clean redux state.
+ *
+ * @param {string} loginError login error object compatible with informed api
+ * @param {string} clearLoginError action to call after error is set
+ * @param {object} formApi React ref that references informed form api
+ */
+
+export const useLoginError = (loginError, clearLoginError, formApi) => {
```
mrfelton: I might be good to rename this as `useFormError` and move to global hooks dir. This could be useful for other forms that want to apply an error for a form and do something like clearing it from the state immediately after.

renderer/components/Settings/Security/PasswordState.js
```diff
@@ -0,0 +1,37 @@
...
+    </Flex>
+  )
+}
+
+PasswordState.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  onDisable: PropTypes.func.isRequired,
+  onEnable: PropTypes.func.isRequired,
+  // eslint-disable-next-line react/boolean-prop-naming
+  value: PropTypes.bool,
```
mrfelton: How about renaming this something like `isActive` instead of working around the lint rule?

renderer/containers/Initializer.js
```diff
@@ -98,17 +67,7 @@ const mapStateToProps = state => ({
...
-  initLocale,
-  initWallets,
-  initAutopay,
-  initChannels,
-}
-
 export default connect(
   mapStateToProps,
-  mapDispatchToProps
+  null
```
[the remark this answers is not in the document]
korhaliv: I think that was inherited from the parent PR or is a merge bug, but to answer you question - in our case it doesn't matter probably, but there is a difference between passing `null` and `undefined` in `dispatch` being passed as a prop. If we pass `null` dispatch won't implicitly get passed into component

renderer/reducers/account.js
korhaliv: oh nice. this is exactly why i was looking for but forgot `putConfig` exists

utils/crypto.js
```diff
@@ -232,3 +232,18 @@ export const getFeeRange = (routes = []) => ({
...
   max: getMaxFee(routes),
 })
+
+/**
+ * sha256digest - Generates a digest of the `message`
+ *
+ * @param {string} message message to hash
+ * @returns {string} sha256 hex hash of a `message`
+ */
+export const sha256digest = async message => {
```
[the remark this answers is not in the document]
korhaliv: done

test/unit/components/UI/__snapshots__/Countdown.spec.js.snap
```diff
@@ -16,7 +16,7 @@ exports[`component.UI.Countdown should render correctly with default props 1`] =
   fontSize="m"
 >
   Expired
-   
+
```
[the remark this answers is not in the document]
korhaliv: sry, this is local issue with tests

renderer/store/schema/account.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @class Account
+ * Account helper class.
+ */
+export default class Account {
```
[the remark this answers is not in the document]
korhaliv: removed

renderer/reducers/modal.js
korhaliv: added

renderer/reducers/account.js
korhaliv: the approach has been refactored based on slack convo

renderer/reducers/account.js
korhaliv: this was removed all together

renderer/reducers/account.js
korhaliv: wasn't addressed directly as per slack convo, but some refactoring was done

renderer/components/Settings/Security/hooks.js
korhaliv: good call!

renderer/components/Settings/Security/PasswordState.js
korhaliv: sure :)

korhaliv:
@mrfelton PR updated

renderer/components/Login/Login.js
```diff
@@ -0,0 +1,66 @@
...
+
+  const handleSubmit = ({ password }) => login(password)
+
+  return (
+    <Form getApi={api => (formApiRef.current = api)} onSubmit={handleSubmit} width={1}>
+      {({ formState: { submits, error } }) => {
+        const willValidateInline = submits > 0
+        return (
+          <CenteredContent {...rest} mx="auto" width={11 / 16}>
+            <Flex alignItems="center" flexDirection="column" px={0} width={5 / 8} {...rest}>
```
mrfelton: `...rest` is being applied twice - once to the outer `CenteredContent` and then again to this `Flex`. It should probably only be applied on the outer most element.

electron/secureStorage/service.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * createStorageService - Create secure storage service.
+ *
+ * @param  {object} mainWindow Browser window
+ */
+export default function createStorageService(mainWindow) {
+  const storage = createStorage(config.secureStorage.namespace)
+  // helper func to send messages to the renderer process
+  const send = (msg, params) => mainWindow.webContents.send(msg, params)
+  const key = 'app-password'
```
mrfelton: Rather than having a hardcoded key here, what if we extracted this so that this whole thing was more reusable:

So rather than `getPassword` you'd have `get(key)`. Rather than `deletePassword` you would have `delete(key)` etc.

We still don't know how we will store additional sensitive data (new keys under the store, in IDB encrypted with the password etc), but it would probably be good to make this method more generic anyway so that we have an easy way to manipulate multiple things in the key store.

We can defer this to a follow up refactor PR if you prefer though.

renderer/reducers/utils/waitForIpc.js
```diff
@@ -0,0 +1,52 @@
...
+  responseEvent,
+  failureEvent,
+  timeout = 5000
+) {
+  return dispatch => {
+    const failureEventName = failureEvent || `${message}Failure`
+    const successEventName = responseEvent || message
+
+    const removeFailureEventListeners = () =>
+      window.ipcRenderer.removeAllListeners(failureEventName)
```
mrfelton: Calling `removeAllListeners` could have unintended consequences, removing listeners unrelated to this particular call to `waitForIpcEvent`.

We should take care to only remove the specific listeners that were added by us.

test/unit/utils/sha256.spec.js
```diff
@@ -0,0 +1,9 @@
+import sha256digest from '@zap/utils/sha256'
+
+describe('sha256digest', () => {
+  it('should produce correct sha256 hashes of a string', () => {
+    expect(sha256digest('chancellor on the brink of second bailout for banks')).toEqual(
```
mrfelton: nice ;)

translations/de-DE.json
```diff
@@ -298,13 +298,14 @@
...
   "components.Home.wallets_menu_local_title": "Deine Konten",
   "components.Home.wallets_menu_other_title": "Mehr",
-  "components.Loading.connecting_to_lnd": "Verbindung zu Lnd gRPC-Schnittstelle",
-  "components.Loading.connecting_to_unlocker": "Verbindung zum Entsperr-Interface",
-  "components.Loading.fetching_tickers": "Lade Zeichner",
-  "components.Loading.loading": "Lädt",
-  "components.Loading.starting_lnd": "Starte Lnd",
-  "components.Loading.starting_neutrino": "Neutrino starten",
-  "components.Loading.starting_wallet_unlocker": "Wallet-Entsperrung starten",
+  "components.Loading.connecting_to_lnd": "",
```
mrfelton: Looks like we have lost some translations here.

renderer/reducers/account.js
```diff
@@ -0,0 +1,272 @@
...
+    mainLog.warn('checkAccountPasswordEnabled error: %o', e)
+    dispatch({ type: LOGIN_FAILURE, error: LOGIN_NOT_ALLOWED })
+  }
+}
+
+/**
+ * setIsPasswordEnabled - Whether password is set for the account.
+ *
+ * @param {string} value value
+ * @returns {Function} Thunk
```
mrfelton: This returns an `Action` not a `Thunk`

renderer/reducers/account.js
```diff
@@ -0,0 +1,272 @@
...
+    dispatch({ type: LOGIN, password })
+    await dispatch(requirePassword(password))
+    dispatch({ type: LOGIN_SUCCESS })
+  } catch (error) {
+    dispatch({ type: LOGIN_FAILURE, error: error.message })
+  }
+}
+
+/**
+ * clearVerifyUserError - Clear verify user error.
```
mrfelton: `Clear verify user error.` -> `Clear login user error.`

renderer/reducers/account.js
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * @returns {object} Action
```
mrfelton: This one returns a `Thunk` not an `Action`

PR #2997, opened 2019-10-14
fix(ui): make sure channel create button doesn't overflow

### LINKED ISSUE #2989: Open Channel form submit button text overrun with large satoshi amounts
## Description

Open Channel form submit button text overrun with large satoshi amounts

## Expected Behavior

Button length should adapt

## Actual Behavior

Button length remains constant and text overflows

## Steps to Reproduce

Enter large satoshi amount in create channel form, click the next button, and see how the submit button text looks

## Context

Channels UI

## Screenshots

image

bolatovumar:
## Motivation and Context:

Fix #2989

## How Has This Been Tested?

Manually

Screen Shot 2019-10-14 at 4 05 22 PM

## Checklist:

### DIFF
--- a/renderer/components/Channels/ChannelCreateForm.js
+++ b/renderer/components/Channels/ChannelCreateForm.js
@@ -46,10 +46,10 @@ const ShowHide = Keyframes.Spring({
 
 const FormButtons = ({ isNextButtonDisabled, nextButtonText, onBack }) => (
   <Flex alignItems="center" mb={3}>
-    <Box width={1 / 4}>
+    <Box width={1 / 6}>
       <ChannelBackButton mr="auto" onClick={onBack} />
     </Box>
-    <Flex justifyContent="center" width={2 / 4}>
+    <Flex justifyContent="center" width={4 / 6}>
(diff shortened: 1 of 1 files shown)


bolatovumar:
@mrfelton, hmm... It shouldn't have messed up the button positioning. It was fine when I tested it. Anyway, I like your solution better as it's simpler even though it doesn't entirely solve the issue. You could still run into the overflow if the satoshi amount is large enough but it would have to be a really large amount.

bolatovumar:
Oh, I think I know why it looks messed up for you. This bug is not present on the `master` branch and I branched off from the `next` branch where this bug is present.

bolatovumar:
Looks like the reason this issue is not present on the `master` branch but is present on the `next` branch is because the button component now has `min-width: 0;` applied to it while previously it didn't. I guess it's a change introduced in the new version of rebass.

Given that this issue is not present in `master`, should the fix be rebased against `next` instead then?

mrfelton:
> Given that this issue is not present in master, should the fix be rebased against next instead then?

Sorry, I missed this comment. The fix has been merged against master branch, which is fine since it doesn't introduce any new features.

PR #2998, opened 2019-10-15
fix(ui): retain channel view mode on logout

### LINKED ISSUE #2988: Preferred channel list view mode not retained through wallet logout
## Description

Channel list view mode not retained 

## Expected Behavior

Preferred channel list view mode should be remembered across logins

## Actual Behavior

Channel list view mode always resets to `card` view after logging into a wallet

## Steps to Reproduce

1. Log into a wallet
2. Open Manage Channels screen and set view mode to list view
3. log out of the wallet
4. Log back in and go back to the Manage Channels page

Note that it's back on the card view again.

## Context

Channes UI

bolatovumar:
## Motivation and Context:

fix #2988

## How Has This Been Tested?

Manually

## Checklist:

### DIFF
--- a/renderer/reducers/channels.js
+++ b/renderer/reducers/channels.js
@@ -1,11 +1,11 @@
 import { createSelector } from 'reselect'
 import { proxyValue } from 'comlinkjs'
+import get from 'lodash/get'
 import orderBy from 'lodash/orderBy'
 import throttle from 'lodash/throttle'
 import { requestSuggestedNodes } from '@zap/utils/api'
 import truncateNodePubkey from '@zap/utils/truncateNodePubkey'
 import { grpcService } from 'workers'
-import config from 'config'
 import { updateNotification, showWarning, showError } from './notification'
 import { fetchBalance } from './balance'
 import { walletSelectors } from './wallet'
@@ -53,7 +53,6 @@ const initialState = {
   ],
 
   selectedChannelId: null,
-  viewMode: config.channels.viewMode,
 
   suggestedNodes: {
     mainnet: [],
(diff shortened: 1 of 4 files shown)


renderer/reducers/channels.js
```diff
@@ -843,7 +844,12 @@ const channelSortSelector = state => state.channels.sort
 const channelSortOrderSelector = state => state.channels.sortOrder
 const filterSelector = state => state.channels.filter
 const nodesSelector = state => networkSelectors.nodes(state)
-const viewModeSelector = state => state.channels.viewMode
+const viewModeSelector = createSelector(
+  settingsSelectors.currentConfig,
+  currentConfig => {
+    return get(currentConfig, 'channels.viewMode', config.channels.viewMode)
```
mrfelton: `currentConfig` contains the default config merged with local config overrides, and with database overrides applied. So, you shouldn't need to set the default here as it's redundant.

bolatovumar: OK, will update.

bolatovumar: Alright, done.

renderer/reducers/channels.js
```diff
@@ -843,7 +844,10 @@ const channelSortSelector = state => state.channels.sort
 const channelSortOrderSelector = state => state.channels.sortOrder
 const filterSelector = state => state.channels.filter
 const nodesSelector = state => networkSelectors.nodes(state)
-const viewModeSelector = state => state.channels.viewMode
```
mrfelton: Since `state.channels.viewMode` is no longer used, I'd suggest that we remove it from the state altogether to avoid confusion.

bolatovumar: @mrfelton I hope I understood your correctly. See the follow-up commit.

mrfelton: Yes, that's right @bolatovumar . However, the seconds commit should have the scope `refactor` instead of `fix` since it does not fix an actual user facing bug and therefore is not something that we'd want to include in the changelog.

mrfelton:
Also note tests seem to be failing @bolatovumar . Probably need to update the test snapshots since you have modified the state.

bolatovumar:
@mrfelton ok, updated test snap files and squashed the commits into one

PR #3024, opened 2019-10-20
fix(wallet): increase fee limit on payment retries

mrfelton:
## Description:

In the case the queryRoutes gives us an unplayable route we retry the payment a couple of times. When doing the retries, increase the fee limit to increase the chance of payment success.

## Motivation and Context:

fix 32952

## How Has This Been Tested?

Manually

## Types of changes:

Fix

## Checklist:

### DIFF
--- a/utils/crypto.js
+++ b/utils/crypto.js
@@ -1,4 +1,6 @@
 import get from 'lodash/get'
+import config from 'config'
+import range from 'lodash/range'
 import { address } from 'bitcoinjs-lib'
 import lightningRequestReq from 'bolt11'
 import coininfo from 'coininfo'
@@ -172,9 +174,9 @@ export const getNodeAlias = (pubkey, nodes = []) => {
 }
 
 /**
- * getMinFee - Given a list of routest, find the minimum fee.
+ * getMinFee - Given a list of routes, find the minimum fee.
  *
- * @param {*} routes List of routes
+ * @param {Array} routes List of routes
  * @returns {number} minimum fee rounded up to the nearest satoshi
  */
 export const getMinFee = (routes = []) => {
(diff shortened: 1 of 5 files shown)


utils/crypto.js
```diff
@@ -198,9 +199,30 @@ export const getMaxFee = routes => {
...
+    return null
+  }
 
-  // Add one to the fee to add room for accuracy error when using as a fee limit.
-  return fee + 1
+  // Factor in expontnial increment that could be applied in retry attempts.
+  const {
+    invoices: { retryCount, feeIncrementExponent },
+  } = config
+  range(retryCount).forEach(() => {
```
mrfelton: fee*Math.pow(feeIncrementExponent,retryCount)

utils/crypto.js
```diff
@@ -198,9 +199,30 @@ export const getMaxFee = routes => {
...
+}
+
+/**
+ * getMaxFee - Given a list of routes, find the maximum fee factoring in all possible payment retry attempts.
+ *
+ * @param {*} routes List of routes
+ * @returns {number} maximum fee
+ */
+export const getMaxFeeInclusive = routes => {
+  let fee = getMaxFee(routes)
```
mrfelton: move this down

utils/crypto.js
```diff
@@ -198,9 +199,30 @@ export const getMaxFee = routes => {
     return null
   }
   const fee = routes.reduce((max, b) => Math.max(max, b.total_fees), routes[0].total_fees)
+  return Math.ceil(Math.max(fee, 1))
+}
+
+/**
+ * getMaxFee - Given a list of routes, find the maximum fee factoring in all possible payment retry attempts.
+ *
+ * @param {*} routes List of routes
```
mrfelton: *->array

utils/crypto.js
```diff
@@ -183,12 +185,11 @@ export const getMinFee = (routes = []) => {
   }
   const fee = routes.reduce((min, b) => Math.min(min, b.total_fees), routes[0].total_fees)
```
mrfelton: fee =Math.celil(Math.max(1, Math.min(...b.total_fees)))

utils/crypto.js
mrfelton: This actually gives a slightly different result to what I wanted to achieve because we want to apply `Math.ceil` at each retry attempt.

PR #3036, opened 2019-10-22
fix(ui): don't open channel management modal when not needed

### LINKED ISSUE #3035: Closing "Open a Channel" modal opens channel management modal
This issue is on the `next` branch.

## Description

When you open "Open a Channel" modal and then try to close it you see the channel management modal. See video here: https://youtu.be/Cf7fG7Dl9M8

## Expected Behavior

See main wallet window when closing the "Open a channel" modal.

## Actual Behavior

See channel management UI instead.

## Possible Fix

Not sure but I think how modal are handle changed recently? @korhaliv probably knows.

## Your Environment

- Zap version: 0.6.0-alpha

bolatovumar:
## Motivation and Context:

fix #3035

## How Has This Been Tested?

Manually

### DIFF
--- a/renderer/components/Channels/ChannelsMenu/ChannelsMenu.js
+++ b/renderer/components/Channels/ChannelsMenu/ChannelsMenu.js
@@ -55,10 +55,7 @@ const ChannelsMenu = ({
       id: 'create',
       title: <FormattedMessage {...messages.menu_item_channel_create_title} />,
       description: <FormattedMessage {...messages.menu_item_channel_create_description} />,
-      onClick: () => {
-        openModal('CHANNELS')
-        openModal('CHANNEL_CREATE')
-      },
+      onClick: () => openModal('CHANNEL_CREATE'),
     },
   ]
 
--- a/test/e2e/basic-ui.spec.js
+++ b/test/e2e/basic-ui.spec.js
@@ -114,7 +114,6 @@ test('provide access to basic wallet functionality', async t => {
     .expect(wallet.createChannels.modal.exists)
     .ok()
     .click(wallet.createChannels.closeButton)
-    .click(wallet.channelsForm.closeButton)
 
     // test profile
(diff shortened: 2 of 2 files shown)


bolatovumar:
@mrfelton, ok I will take a look. What I'm wondering is how this broke in the first place.

bolatovumar:
@mrfelton now that I'm thinking about it, would you really want to land on the channel management screen after you initiated a channel. Once you have more than 2 channels if you are in card view or 4 channels if you are in list view, you have to scroll down to see the pending channel you just created. If you already have a couple of channels open, you probably don't really care to scroll down to see the "pending" channel and will just close the channel management pane. Does this make sense?

mrfelton:
After creating a new channel, there is a period where we show the channel in ‘loading’ state prior to it becoming pending. In this state I think it shows at the top of the list - so newly opened channels should generally always show at the top of the list. I’m not 100% on how the selected sort mode affects that though.

bolatovumar:
@mrfelton this doesn't seem to be the case (using the `master` branch). See video here: https://youtu.be/UOn0oO2lp0s

Is this a bug? Would you expect to show the newly "opened" channel at the top of the list? In that case it would make sense to show the channel management view once you initiate opening a channel. If that's not the case then going back to the wallet view makes sense.

mrfelton:
Hmm. Interesting. This was the behavior in the past, but it looks like it no longer works this way since adding the ability to filter and sort the results.

This makes sense, because if you have your list set to show only online channels, ordered by activity (for example) then it wouldn't make sense to show loading or pending channels at the top of the list.

So, given this, and the fact that we do now also have a success notification that notifies the user if the channel was opened or not perhaps it does make sense to return the user to the wallet page if that is where they accessed the create channel form from.

bolatovumar:
@mrfelton OK, I will fix up the issue with the tests and push it.

Issue #3045, opened 2019-10-24
Unlocalizable strings


sohomang:
## Description

The following strings do not exist in `translations/en.json`, so they're not showing up in Crowdin:

- "Channels" menu's label on the main screen
- "onchain" and "in channels" on the `Pay` screen
- "or" on the `Connect` screen during the onboarding
- Some error message that shows up when trying to unlock a remote node

Bonus:
- Shouldn't the `open` status label next to each channel say `online` instead? In contrast to the `offline` status.

## Screenshots

Respectively:
<img width="284" alt="Channels" src="https://user-images.githubusercontent.com/28465517/67442893-e1319000-f602-11e9-843f-8606c270f3e7.png">
<img width="196" alt="On-chain:In channels" src="https://user-images.githubusercontent.com/28465517/67442898-e68eda80-f602-11e9-85e3-0f9b3e29e757.png">
<img width="560" alt="Or" src="https://user-images.githubusercontent.com/28465517/67442907-ec84bb80-f602-11e9-94e5-745ea83bef0c.png">
<img width="1193" alt="Screen Shot 2019-10-24 at 02 33 42" src="https://user-images.githubusercontent.com/28465517/67443826-c06b3980-f606-11e9-9a44-e0038ec923c2.png">
<img width="179" alt="Screen Shot 2019-10-24 at 02 36 35" src="https://user-images.githubusercontent.com/28465517/67443917-1f30b300-f607-11e9-9f95-bbd81f1e4585.png">

bolatovumar:
I can take a look and include missing strings if they are hardcoded.

bolatovumar:
@sohomang I addressed all issues except for the last one and the "bonus" one. I didn't address the last one 
> Some error message that shows up when trying to unlock a remote node

As these are error messages, do we wanna have them localizable @mrfelton?

As for the bonus one
>Shouldn't the open status label next to each channel say online instead? In contrast to the offline status.

This makes sense to me. Is there a reason why we would want to keep it as "open" instead of "online" @mrfelton?

bolatovumar:
> Shouldn't the open status label next to each channel say online instead? In contrast to the offline status.

According @sohomang, there was a consensus that we want to change the wording to "online" here. Should we go ahead with it @mrfelton?

bolatovumar:
[...]

PR #3050, opened 2019-10-25
fix(i18n): extract hardcoded strings to translation files

bolatovumar:
Replaced some hardcoded string identified in #3045.

## Motivation and Context:

Address #3045

## How Has This Been Tested?

Manually

## Checklist:


renderer/components/Pay/Pay.js
```diff
@@ -738,12 +738,12 @@ class Pay extends React.Component {
                           <Text fontSize="xs" textAlign="center">
                             <CryptoValue value={walletBalanceConfirmed} />
                             {` `}
-                            {cryptoUnitName} (onchain),
+                            {cryptoUnitName} <FormattedMessage {...messages.onchain_label} />
```
mrfelton: @bolatovumar I'd suggest combining these into a single translation string that includes both `cryptoUnitName` and the `onchain_label`.

So that the translation strings looks more like this:

```jsx
{amont} {cryptoUnitName} (onchain)
```

And then use like:

```jsx
<FormattedMessage {...messages.onchain_balance} values={{ amount, cryptoUnitName }} />
```
This will provide greater flexibility for languages that don't follow the latin text formatting rules.

renderer/components/Pay/Pay.js
```diff
@@ -738,12 +738,12 @@ class Pay extends React.Component {
...
                           </Text>
                           <Text fontSize="xs" textAlign="center">
                             <CryptoValue value={channelBalance} />
                             {` `}
-                            {cryptoUnitName} (in channels)
+                            {cryptoUnitName} <FormattedMessage {...messages.in_channels_label} />
```
mrfelton: Same goes for this one.

renderer/components/Pay/Pay.js
bolatovumar: Makes sense. Will do.

bolatovumar: Done.

renderer/components/Pay/Pay.js
bolatovumar: Done.

Issue #3065, opened 2019-10-27
Enable jsx-no-hardcoded-content lint rule


bolatovumar:
We should have a lint rule which detects hardcoded strings. We don't want them to end up in the UI since they are non-localizable. `eslint-plugin-shopify` package has a `jsx-no-hardcoded-content` rule which helps wit this.

See discussion here: https://github.com/LN-Zap/zap-desktop/issues/3045

We decided to enable this rule: https://github.com/LN-Zap/zap-desktop/issues/3045#issuecomment-546742395

bolatovumar:
I will follow this up with a PR.

mrfelton:
Closed via https://github.com/LN-Zap/zap-desktop/pull/3066

PR #3066, opened 2019-10-28
Externalize hardcoded UI strings

bolatovumar:
## Motivation and Context:

Address #3065

## How Has This Been Tested?

Manually

## Checklist:


bolatovumar:
@mrfelton I enabled the rule and fixed the obvious violations. The rest of the are mostly due to cases where we combine multiple values like `amount` and `cryptoUnitName` together with a space to form a string. There are also a few issues where we have hardcoded signs like `+`, `-`, `~` and `:`. I'm wondering what your opinion on addressing these issue is. I guess it's worth it for the languages that don't follow the word order conventions of English?

test/.eslintrc
```diff
@@ -2,9 +2,10 @@
   "env": {
     "jest/globals": true
   },
-  "plugins": ["jest"],
+  "plugins": ["jest", "shopify"],
```
mrfelton: Do we really need to enable this for the tests? Doesn't seem like it would provide much benefit here.

.eslintrc.js
```diff
@@ -85,6 +85,13 @@ module.exports = {
       },
     ],
     'prefer-promise-reject-errors': 0,
+    'shopify/jsx-no-hardcoded-content': [
+      'warn',
```
mrfelton: Ideally, we should make this an `error` rather than a warning as if we are going to user rules I'd rather they be enforced as a standard rather than left as something that cal slip through the cracks unaddressed.

test/.eslintrc
bolatovumar: Makes sense. We don't really need it here.

renderer/components/Channels/messages.js
```diff
@@ -105,4 +105,5 @@ export default defineMessages({
   channel_desc_sort: 'Switch to descending order',
   channel_sort_label: 'Sort by:',
   channel_filter_label: 'Show:',
+  no_search_results: 'Your seach did not return any results.',
```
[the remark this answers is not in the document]
bolatovumar: Oops, I just copy-pasted existing string and I guess it had a typo in it.

.eslintrc.js
bolatovumar: @mrfelton I set it as a error at first but because I still had unresolved issues, I couldn't push my branch out lol. I will keep it as a warning until all issues are resolved.

renderer/components/Channels/messages.js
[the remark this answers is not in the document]
bolatovumar: This is resolved now.

bolatovumar:
@mrfelton alright, I rebased against  the `next` branch. Looks like lint rules can't be added directly anymore, only through `zap-eslint-config` package. Should I add this new rule to that package and leave this PR as is and just have it fix the issues that it fixes.

mrfelton:
@bolatovumar `zap-eslint-config` contains rules that we use across all projects but you can add rules that are more specific to zap-desktop such as this one directly to this project.

bolatovumar:
@mrfelton ok, sweet. I will just add this one rule to this project then.

bolatovumar:
@mrfelton by `signDisplay` prop do you mean this (https://v8.dev/features/intl-numberformat#sign)? Looks like that option is only available in Chrome 77+ while Electron 6 which we are using is based on Chrome 76.

bolatovumar:
@mrfelton also, I'm not sure I understand what you mean by 

> We should leverage the rich text formatting features of intl to combine strings with special formatting into a single translatable string rather than piecing them together ourselves in the markup, as per https://github.com/formatjs/react-intl/blob/master/docs/Components.md#rich-text-formatting

For example, one of the things this lint rule points out as an error is this:
```js
value: (
        <Text>
          <CryptoValue value={local_chan_reserve_sat} />
          {` `}
          {cryptoUnitName}
        </Text>
      ),
```

I'm thinking of replacing it with this which will allow rearranging the words for languages that don't follow English word order rules:
```js
value: (
        <FormattedMessage
          {...messages.crypto_amount_with_unit}
          values={{ amount: local_chan_reserve_sat, cryptoUnitName }}
        />
      ),
```

However, that means every language message file will have to contain this message, right?

Similarly, what about something like this:
```js
<Text color="gray">
    ≈&nbsp;
    <FiatValue style="currency" value={amount} />
</Text>
```

The hardcoded "approximately equal" sign triggers this lint rule. I'm wondering if something like this should be extracted into its own translatable string or leave it as is.

bolatovumar:
Other examples that trigger this rule:

```js
<Flex alignItems="baseline">
    <Text color="gray">=</Text>
    <FiatValue color="gray" style="currency" value={limit || defaultValue} />
</Flex>
```

```js
<Flex justifyContent="center" mt={42} width={1 / 11}>
    =
</Flex>
```

```js
const RequiredMark = () => (
  <Supertext fontSize="s" ml={1}>
    *
  </Supertext>
)
```

```js
<Heading.h4 my={1}>
    - <FormattedMessage {...messages.or} /> -
</Heading.h4>
```

```js
<Text>
    <FormattedMessage {...messages.verify_host_title} />{' '}
    <Span color="superGreen">{window.Zap.splitHostname(hostname).host}</Span>?{' '}
</Text>
```

```js
<StyledTooltipIconBox
    bg="primaryText"
    ml={1}
    onMouseEnter={this.hoverOn}
    onMouseLeave={this.hoverOff}
>
    <Text color="primaryColor" fontSize="s" fontWeight="bold">
    ?
    </Text>
</StyledTooltipIconBox>
```

I feel like the above can be counted as exceptions and the rule ignored for them, no?

mrfelton:
> However, that means every language message file will have to contain this message, right?

Now, translation files for other languages only need to add the string if they want to override the defaults. In this case, they would only need to include it if they wanted to change the display order of the two variables.

> I feel like the above can be counted as exceptions and the rule ignored for them, no?

Yes, I guess bypassing the rule for these bits of code probably makes the most sense.

bolatovumar:
@mrfelton alright, I enabled the rule and fixed the issues where appropriate and disabled the rule in cases where it was not needed (the examples I listed above).

bolatovumar:
@mrfelton oh, right, it's probably missing a non-breaking space character

bolatovumar:
@mrfelton alright, fixed the issue with the detail modal

PR #3138, opened 2019-11-07
Fix activity filters

korhaliv:
## Description:
Fix a bug that resulted in "Sent" filter being removed after sending a payment. This PR also removes sending activity from "Sent" filter to avoid duplicated when it's used together with "Pending" activity

## How Has This Been Tested?
Manually

## Types of changes:
Bugfix

## Checklist:

### DIFF
--- a/renderer/components/Pay/Pay.js
+++ b/renderer/components/Pay/Pay.js
@@ -14,9 +14,9 @@ import { intlShape } from '@zap/i18n'
 
 class Pay extends React.Component {
   static propTypes = {
+    addFilter: PropTypes.func.isRequired,
     chain: PropTypes.string.isRequired,
     chainName: PropTypes.string.isRequired,
-    changeFilter: PropTypes.func.isRequired,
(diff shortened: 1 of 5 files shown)


renderer/components/Pay/Pay.js
```diff
@@ -238,7 +238,7 @@ class Pay extends React.Component {
     }
 
     // Change the transaction filter to ALL transactions
```
mrfelton: This comment probably needs to be updated as it's out of date now.

renderer/components/Pay/Pay.js
```diff
@@ -238,7 +238,7 @@ class Pay extends React.Component {
...
-    changeFilter('SENT_ACTIVITY')
+    addFilter('SENT_ACTIVITY')
```
mrfelton: After sending a new payment, the payment will be in pending state - so will not be included in the `SENT_ACTIVITY` filter. This should probably also add `PENDING_ACTIVITY`

PR #3145, opened 2019-11-08
Ensure failed payments marked as failed

### LINKED ISSUE #3142: Retried failed payment stuck in processing state
## Description

Retried failed payment stuck in processing state 

## Expected Behavior

If you retry a failed payment and it also fails the second time the second attempt should be marked as failed just like the first one is.

## Actual Behavior

The second failed payment attempt gets stick in payments sending state.

## Possible Fix

Adjust payments sending logic

## Steps to Reproduce

1. Create an LN invoice
2. Try to pay your own invoice - after 3 failed attempts it gets marked as failed
3. Try to pay the same invoice again - this time it gets stuck in sending state

mrfelton:
## Description:

Ensure failed payments marked as failed

Generate a unique payment id for sending payments to enable more accurate tracking of payments across retry attempts.

## Motivation and Context:

Fix #3142

## How Has This Been Tested?

Manually

## Types of changes:

Fix

## Checklist:

### DIFF
--- a/renderer/reducers/payment.js
+++ b/renderer/reducers/payment.js
@@ -1,11 +1,13 @@
 import config from 'config'
 import { createSelector } from 'reselect'
 import uniqBy from 'lodash/uniqBy'
+import find from 'lodash/find'
 import errorToUserFriendly from '@zap/utils/userFriendlyErrors'
 import { getIntl } from '@zap/i18n'
 import { decodePayReq, getNodeAlias } from '@zap/utils/crypto'
 import { convert } from '@zap/utils/btc'
 import delay from '@zap/utils/delay'
+import genId from '@zap/utils/genId'
 import { grpc } from 'workers'
 import createReducer from './utils/createReducer'
 import { fetchBalance } from './balance'
@@ -91,18 +93,6 @@ const decoratePayment = (payment, nodes = []) => {
   }
 }
 
-/**
- * getLastSendingEntry - Find the latest temporary paymentsSending entry for the payment.
- *
- * @param  {object} state Redux state
- * @param  {string} paymentRequest Payment request
- * @returns {object} sendingPayments entry
- */
-const getLastSendingEntry = (state, paymentRequest) =>
-  [...state.payment.paymentsSending]
-    .sort((a, b) => b.creation_date - a.creation_date)
-    .find(p => p.paymentRequest === paymentRequest)
-
 // ------------------------------------
 // Actions
 // ------------------------------------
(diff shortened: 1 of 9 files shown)


services/grpc/lightning.methods.js
```diff
@@ -260,73 +260,64 @@ async function closeChannel(payload = {}) {
...
-          this.emit('sendPayment.error', error)
-          reject(error)
         }
 
+        // In case of an error, notify the client if there was a problem sending the payment.
+        else {
+          grpcLog.error('PAYMENT ERROR', res)
+          const error = new Error(res.payment_error)
+          error.details = res
+          reject(error)
```
korhaliv: Is there a reason why here we reject and in `catch` we re-throw?

renderer/reducers/payment.js
```diff
@@ -282,14 +277,15 @@ export const paymentSuccessful = ({ payment_request }) => async (dispatch, getSt
...
+ * @param {Error} error Error
+ * @param {object} details Failed payment details
+ *
  * @returns {Function} Thunk
  */
-export const paymentFailed = ({ payment_request, error }) => async (dispatch, getState) => {
+export const paymentFailed = (error, { paymentId }) => async (dispatch, getState) => {
   const state = getState()
-  const paymentSending = getLastSendingEntry(state, payment_request)
+  const paymentSending = find(state.payment.paymentsSending, { paymentId })
```
korhaliv: Consider using `paymentsSendingSelector`

services/grpc/lightning.methods.js
```diff
@@ -260,73 +260,64 @@ async function closeChannel(payload = {}) {
...
-              ? invoice.tags.find(t => t.tagName === 'payment_hash')
-              : null
-            paymentHash = paymentHashTag ? paymentHashTag.data : null
-          }
-
-          // Convert the preimage to a hex string.
-          const paymentPreimage = data.payment_preimage
-            ? data.payment_preimage.toString('hex')
+        // Convert payment_hash to hex string.
+        data.payment_hash = data.payment_hash ? data.payment_hash.toString('hex') : null
```
korhaliv: Minor but could be a little cleaner if you do 
```
if (data.payment_hash) {
}
else {
}
```
so two cases are more obvious and no extract checks are performed

services/grpc/lightning.methods.js
mrfelton: The rejection here is used to reject the Promise within the `data` event handler in the case that the `sendPayment` lightning call returns an error (which it does by returning a success but with `payment_error` set).

The catch at the end wraps the entire method and is triggered in the event that an unexpected error occurs anywhere within our `sendPayment` wrapper. It doesn't re throw the rejection from the `data` event handler.

PR #3146, opened 2019-11-08
Ensure correct expire date for invoices with far future expiry

### LINKED ISSUE #3144: Invoice with expiry in far future shows as expired on invoice detail page
## Description

Invoice with expiry in far future shows as expired on invoice detail page

## Expected Behavior

Should show actual expiry date.

## Actual Behavior

Shows as expired

## Possible Fix

This is because we use the `Countdown` component to display the expiry date which uses `setInterval` which can not handle large numbers.

Consider not using the Countdown component if the date is in the far future.

## Steps to Reproduce

Create invoice with expiry in 1 year and view the invoice detail page

mrfelton:
## Description:

The largest number (ms) supported by setInterval is 2147483647. If countdown expire date is outside of this range do not use a timer.

Note: Technically if a user was to keep a countdown on the screen for over 24 days it would display an inaccurate expired message, but in practice this is probably such an edge case that it's not worth investing time into addressing at this point.

## Motivation and Context:

Fix #3144

## How Has This Been Tested?

Manually

## Types of changes:

Fix

## Checklist:

### DIFF
--- a/renderer/components/UI/Countdown.js
+++ b/renderer/components/UI/Countdown.js
@@ -1,94 +1,70 @@
-import React from 'react'
+import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import { FormattedMessage, FormattedRelativeTime } from 'react-intl'
+import { useTimeout } from 'hooks'
 import Text from './Text'
 import messages from './messages'
 
-class Countdown extends React.Component {
-  state = {
-    isExpired: null,
-    timer: null,
+const Countdown = ({
+  offset,
+  onExpire,
+  colorActive,
+  colorExpired,
+  countdownStyle,
+  isContinual,
+  updateInterval,
+  ...rest
+}) => {
+  let expiresIn = offset
+  if (offset instanceof Date) {
+    expiresIn = offset - Date.now()
+    expiresIn = Math.round(expiresIn / 1000)
   }
 
-  static propTypes = {
-    colorActive: PropTypes.string,
-    colorExpired: PropTypes.string,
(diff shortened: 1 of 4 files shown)


renderer/components/UI/Countdown.js
```diff
@@ -39,15 +40,17 @@ class Countdown extends React.Component {
       this.expiresIn = expiresIn
     }
 
-    if (expiresIn >= 0) {
+    // The largest number (ms) supported by setInterval is 2147483647.
+    // If countdown expire date is within this range setup a timer that updates isExpired state on expiry.
+    if (inRange(this.expiresIn, 0, 2147483)) {
```
korhaliv: And if it's not in range (>~25 days?) That's a rare case, but if expiry is out of range we could setup a timer that will check in 2147483 secs again and if it's again out of range setup the timer again and so on, until expiry is in range. This probably has a very limited real world application, but would bug free. What do you think?

mrfelton: I've done a refactor of this component to convert it to a functional component, simplify it somewhat, and use the `useTimeout` hook to handle updating the isExpired state.

This has resolved the visible bugs that were present and that I was seeking to fix in this PR. Technically, the `useTimeout` hook itself doesn't properly support far future timeouts but in reality this is an edge case, not something that really affects us, and not something that I want to address here.

I've created a separate issue to enhance the `useTimeout` hook to support far future expiry dates. See https://github.com/LN-Zap/zap-desktop/issues/3159

PR #3155, opened 2019-11-09
fix(wallet): present correct fee when no route

### LINKED ISSUE #3154: Fee rate displayed as less than 1 satoshi when route and fee is actually unknown
## Description

Fee rate displayed as 1 satoshi when route and fee is actually unknown.

## Expected Behavior

If no routes are found when running query routes we should dispkay the fee as `unknown`

## Actual Behavior

We display the fee as `less than 1 satoshi`

## Steps to Reproduce

Try to pay an invoice where a route can not be found (eg, create an invoice and try to pay your own invoice).

mrfelton:
## Description:

Ensure fee rate is displayed as `unknown` if no routes are found.

## Motivation and Context:

Fix #3154

## How Has This Been Tested?

Manually

## Types of changes:

Bug fix

## Checklist:

### DIFF
--- a/renderer/components/Pay/PaySummaryLightning.js
+++ b/renderer/components/Pay/PaySummaryLightning.js
@@ -76,7 +76,7 @@ class PaySummaryLightning extends React.Component {
     let feeMessage = messages.fee_unknown
 
     // If thex max fee is 0 or 1 then show a message like "less than 1".
-    if (maxFee === 0 || maxFee === 1) {
+    if (maxFee >= 0 && maxFee < 1) {
       feeMessage = messages.fee_less_than_1
     }
     // Otherwise, if we have both a min and max fee that are different, present the fee range.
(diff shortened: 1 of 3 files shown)


renderer/components/Pay/PaySummaryLightning.js
```diff
@@ -76,7 +76,7 @@ class PaySummaryLightning extends React.Component {
     let feeMessage = messages.fee_unknown
 
     // If thex max fee is 0 or 1 then show a message like "less than 1".
-    if (maxFee === 0 || maxFee === 1) {
+    if (maxFee >= 0 && maxFee <= 1) {
```
korhaliv: message says less than 1 while we use `<=1` ?

renderer/reducers/pay.js
```diff
@@ -225,7 +225,7 @@ export const queryRoutes = (pubKey, amount) => async dispatch => {
     })
     dispatch({ type: QUERY_ROUTES_SUCCESS, routes })
   } catch (e) {
-    dispatch({ type: QUERY_ROUTES_FAILURE })
+    dispatch({ type: QUERY_ROUTES_FAILURE, error: e.message })
```
korhaliv: Not something that the PR has touched but in 
queryFees error serialization is a bit different.
`const error = get(e, 'response.statusText', e)`
Should probably change to
```
    const error = get(e, 'response.statusText', e.message)
```
there

PR #3168, opened 2019-11-12
Fix getInfo subscription cornercases

korhaliv:
## Description:
Improve `synced_to_chain` status polling
Other:
Add ability to execute tasks immediately after they are added via `scheduler` (`runImmediately` flag)
Fix react hook warning and comments typos

## Motivation and Context:
If the user connects to a wallet that is already synced, it will continue to do the fast polling for up to 10 minutes or so - until the next block comes in.
Problem is that the info stream only emits data when something in the result changes, so when we first set it up if the chain is already synced then no event will be emitted until block_hash changes

## How Has This Been Tested?
Manually. Connecting to different types of wallets

## Types of changes:
Bugfix/Refactor

## Checklist:

### DIFF
--- a/services/grpc/grpc.js
+++ b/services/grpc/grpc.js
@@ -154,35 +154,39 @@ class ZapGrpc extends EventEmitter {
     return proxyValue(this.grpc.waitForState(args))
   }
 
+  isSubscribed(subscription) {
+    return Boolean(this.activeSubscriptions[subscription])
+  }
+
   /**
    * subscribeAll - Subscribe to all gRPC streams.
    */
   async subscribeAll() {
     this.subscribe('invoices', 'transactions', 'backups')
 
-    const finalizeSubscriptions = () => {
-      this.subscribe({ name: 'info', params: { pollInterval: 60000 } })
-      this.subscribeChannelGraph()
+    // Finalize subscriptions if `data.synced_to_chain` is true.
+    // Returns true if subscriptions were finalized and false otherwise
+    const finalizeSubscriptions = async data => {
+      if (data && data.synced_to_chain) {
+        if (this.isSubscribed('info')) {
+          await this.unsubscribe('info')
+        }
+        this.subscribe({ name: 'info', params: { pollInterval: 60000 } })
+        this.subscribeChannelGraph()
+        return true
+      }
+      return false
     }
+
+    const data = await this.services.Lightning.getInfo()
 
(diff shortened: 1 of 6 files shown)


test/unit/utils/scheduler.spec.js
```diff
@@ -154,4 +154,16 @@ describe('createScheduler tasks execution', () => {
...
+  test('handles runImmediately correctly', () => {
+    const scheduler = createScheduler()
+
+    const promise = new Promise(resolve => {
+      scheduler.addTask({ task: () => resolve(true), baseDelay: 10, runImmediately: true })
+    })
+
+    scheduler.removeAllTasks()
+
+    return expect(promiseTimeout(100, promise)).resolves.toBe(true)
```
mrfelton: Shouldn't this be set to `5ms` in order to verify that the task is run and resolves before the `10ms` base delay?

services/grpc/grpc.js
```diff
@@ -154,23 +154,38 @@ class ZapGrpc extends EventEmitter {
...
-    // sync to complete before accepting subscriptions.
-    this.on('subscribeGetInfo.data', async data => {
-      const { synced_to_chain } = data
-      if (synced_to_chain && !this.activeSubscriptions.channelgraph) {
-        // reduce poll interval to once a minute after synced_to_chain is true
-        await this.unsubscribe('info')
+  async subscribeAll() {
+    this.subscribe('invoices', 'transactions', 'backups')
+
+    // Finalized subscriptions if `data.synced_to_chain` is true.
```
mrfelton: Should read `Finalize subscriptions if...` (not `Finalized`(

services/grpc/grpc.js
```diff
@@ -154,23 +154,38 @@ class ZapGrpc extends EventEmitter {
...
+        }
         this.subscribe({ name: 'info', params: { pollInterval: 60000 } })
         this.subscribeChannelGraph()
+        return true
       }
-    })
+      return false
+    }
+
+    // If we are already fully synced set up finalize the subscriptions right away.
```
mrfelton: This comment seems out of place.

test/unit/utils/scheduler.spec.js
korhaliv: i guess 5ms wouldn't hurt, but this test assumes `removeAllTasks` actually works. Otherwise thing thing should never resolve and 100ms is just to not hang the test process. So tldr i think using 5ms is even better

PR #3177, opened 2019-11-13
Fix number of issues in pay form

### LINKED ISSUE #3175: Redundant fee fetch when paying LN invoice
## Description
App fetches fees when user tries to pay LN invoice which is redundant

### LINKED ISSUE #3176: Next button is disabled for a valid bitcoin address
## Description
image
Next button is disabled for a valid bitcoin address
Steps to repro:
1. Press Pay
2. Paste valid bitcoin address
3. Hit back

korhaliv:
## Description:
Fix number of issues in Pay form:
- resolves #3175 
- resolves #3176 
- Adds trimming to pay form address/PR input

## How Has This Been Tested?
Manually

## Types of changes:
Bugfixes/Minor perf improvements

## Checklist:

### DIFF
--- a/renderer/components/Pay/PayPanelFooter.js
+++ b/renderer/components/Pay/PayPanelFooter.js
@@ -1,7 +1,8 @@
-import React from 'react'
+import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import { Flex } from 'rebass/styled-components'
 import { FormattedMessage, injectIntl } from 'react-intl'
+import { usePrevious } from 'hooks'
 import { intlShape } from '@zap/i18n'
 import { convert } from '@zap/utils/btc'
 import { CryptoValue } from 'containers/UI'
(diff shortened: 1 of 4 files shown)


renderer/components/Pay/PayPanelFooter.js
```diff
@@ -100,14 +98,13 @@ const PayPanelFooter = props => {
...
   const hasBackButton = currentStep !== PAY_FORM_STEPS.address
-  const hasSubmitButton = currentStep !== PAY_FORM_STEPS.address || (isOnchain || isLn)
 
   return (
     <Flex flexDirection="column">
       {renderLiquidityWarning(props)}
       <PayButtons
         hasBackButton={hasBackButton}
-        hasSubmitButton={hasSubmitButton}
+        hasSubmitButton
```
mrfelton: This seems wrong. The next button is never supposed to be visible on the first step of the pay form since the form automatically moves you to the next step after entering a valid address.

korhaliv: what if you press back?

mrfelton: Ok, so perhaps we should make it visible only after going back?

Right now you have to edit the value for it to pick up the change and move you forwards

korhaliv: Not sure i understand. If you don't hit back then you can't see  it because you can't see that  form

mrfelton: When the form first loads, the button should not be visible because the form is empty. Then you enter an address and it automatically moves you forwards to the next step. If you go back then the button could show in order to address the issue you are observing with the button not being visible when you go back.

korhaliv: I see makes sense. Still maybe this is not actually desired. To my knowledge this is the only place where we would hide the button. In all other places we usually allow user to do one attempt and then do inline validation. I understand that here we have some sort of automatic transition, but i'm not sure this is enough to justify hiding of the button. What do u think

mrfelton: That's true. I think that the thought process here was that there would be no possible reason for anyone to do anything on this form other than to paste an address, so why clutter the ui with distractions? Instead, lets make the page as basic and single-purpose as possible. Not that a submit button is probably too much of a distraction, but that was the logic.

I think the best thing would be to keep it hidden if the user is new to the form in order to keep the existing behaviou as designed, and show it if the user is coming back from another form step

PR #3461, opened 2020-04-12
feat: add TypeScript type definitions

bolatovumar:
## Description:

Adds a `tsconfig.json` file and some TS type definition packages for improved IDE experience. No changes to build steps are required.

## Motivation and Context:

Improve IDE experience with better autocompletion and type hinting.

## How Has This Been Tested?

No code was changed.

## Checklist:


.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "javascript.validate.enable": false,
+  "javascript.validate.enable": true,
```
mrfelton, line 2: @bolatovumar 'm guessing this tells vscode to validate .js files as if they are typescript? Are you able to make this work for Atom too?

bolatovumar: @mrfelton no, it doesn't actually tell VSCode to do anything. It just gives it an option to validate JS files IF you switch the `checkJS` flag to `true` in `tsconfig.json`. I didn't switch that flag on since it would produce a ton of errors in the editor window.

mrfelton: So, how does this PR do anything at all? If I rename files as `.ts` then I get expected typescript support. Otherwise it seems to have no effect at all.

mrfelton: Maybe it's just Atom that is the problem. Seems to work ok in vscode. 🤔 

https://github.com/TypeStrong/atom-typescript/issues/509

bolatovumar: @mrfelton oh, I see. I wasn't aware Atom couldn't detect/use TS type definitions when writing plain JS. I assumed it would be able to.

In VSCode you get slightly better IDE experience with type hinting and autocomplete with this PR. In Atom you get nothing it seems.

bolatovumar: @mrfelton do we want to merge this or should I close this PR?

mrfelton: I'm still considering it @bolatovumar . Conceptually I'm keen on it, but my hesitation is the number of additional dependencies it adds.

bolatovumar: @mrfelton yeah, that makes sense. I'm OK with closing it if needed. Adding these additional packages does add value for improved DX but if the cost of including new packages is too high then that's fine as well.

mrfelton:
Decided to merge and work with it for a while so that we can better gauge the pros vs the cons

bolatovumar:
@mrfelton cool, thanks.

Issue #3600, opened 2020-07-13
24 hour time format


klauslovgreen:
## Detailed Description
Please can you enable support for 24H the format for the Mac Desktop app (ideally just replicate whatever the system is currently using). It already works on iOS ;-)

## Context

Some people have issues figuring am/pm out.. :-)

## Your Environment

- Zap version:
- Operating System and version:

mrfelton:
Closed via https://github.com/LN-Zap/zap-desktop/pull/3688

PR #3688, opened 2020-11-09
Time display mode settings 

rahulbile:
## Description:

Currently, the time displayed in-app for transactions is in 12 hour mode, some users may prefer to change it to 24 hour display mode

## Motivation and Context:

The feature request was requested in #3600 and while doing this change, the display of dates across the app is refactored to have more control on formats and making sure its extensible in future easily

## Types of changes:

Feature

## Checklist:


renderer/components/Activity/PaymentModal/PaymentModal.js
```diff
@@ -1,12 +1,19 @@
...
 import { intlShape } from '@zap/i18n'
 import { getDisplayNodeName } from 'reducers/payment/utils'
 import { Bar, DataRow, Header, Panel, Text } from 'components/UI'
 import { getTag } from '@zap/utils/crypto'
-import { CopyButton, CryptoSelector, CryptoValue, FiatSelector, FiatValue } from 'containers/UI'
+import {
+  CopyButton,
+  CryptoSelector,
+  CryptoValue,
+  FormattedDateTime,
```
mrfelton, line 13: Lets keep these imports in alphabetical order

renderer/components/Activity/Row.js
```diff
@@ -1,15 +1,15 @@
...
 import { Bar, Heading } from 'components/UI'
+import { FormattedDateTime } from 'containers/UI'
 
 const Row = ({ style, item, RowComponent }) => (
   <div style={style}>
     {item.title ? (
       <Box mt={4} pl={4}>
         <Heading.H4 fontWeight="normal">
-          <FormattedDate day="2-digit" month="short" value={item.title} year="numeric" />
+          <FormattedDateTime format="date" month="short" value={new Date(item.title)} />
```
mrfelton, line 12: Why are we converting to `Date` here? Doesn't look like that is done elsewhere

config/default.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   theme: 'dark',
   currency: 'USD',
   locale: 'en',
+  clockFormat24: false,
```
mrfelton, line 23: I’d be more included to make this something like `timeDisplayMode` and make it an enum with values like `24hour` & `12Hour` and then make the widget on the settings page a dropdown so user can select the one they want. It’s a bit more explicit so that the user knows what they get if they don’t select 24 hour clock, and leaves the option open to introduce other time formats.

renderer/components/Settings/messages.js
```diff
@@ -63,4 +63,6 @@ export default defineMessages({
   payments_maxParts_label: 'Payment parts',
   payments_maxParts_description:
     'Maximum number of partial payments that may be used to send a payment.',
+  clockFormat24_label: 'Enable 24-hour clock',
```
mrfelton, line 66: After adding new message strings, you should run `yarn extract-messages` in order to extract into the translation files.

renderer/containers/UI/FormattedDateTime.js
```diff
@@ -0,0 +1,39 @@
...
+          month={month}
+          value={value}
+          year="numeric"
+        />
+      )
+  }
+}
+
+FormattedDateTime.propTypes = {
+  clockFormat24: PropTypes.string,
```
mrfelton, line 33: This should be `PropTypes.bool` in current setup, though as noted earlier Id suggest changing the setting from a bool to an enum string.

renderer/containers/UI/FormattedDateTime.js
```diff
@@ -0,0 +1,39 @@
...
+  format: PropTypes.string,
+  month: PropTypes.string,
+  value: PropTypes.instanceOf(Date),
```
mrfelton, line 36: This isn't always a `Date`. You might want to use `oneOfType` if this can also be things other than a `Date` object.

mrfelton: Also, should probably have `.isRequired`

mrfelton (changes requested):
Nice one @rahulbile . A few comments/suggestions

config/default.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   theme: 'dark',
   currency: 'USD',
   locale: 'en',
+  timeDisplayMode: '12Hour',
```
mrfelton, line 23: Can you add comment for this line that lists out the supported values please?

mrfelton (changes requested):

config/default.js
rahulbile: Done @mrfelton

mrfelton (changes requested):
Am seeing the following error in the console:

> Warning: Failed prop type: Invalid prop `value` supplied to `FormattedDateTime`

renderer/components/Activity/Row.js
```diff
@@ -1,15 +1,15 @@
...
 import { Bar, Heading } from 'components/UI'
+import { FormattedDateTime } from 'containers/UI'
 
 const Row = ({ style, item, RowComponent }) => (
   <div style={style}>
     {item.title ? (
       <Box mt={4} pl={4}>
         <Heading.H4 fontWeight="normal">
-          <FormattedDate day="2-digit" month="short" value={item.title} year="numeric" />
+          <FormattedDateTime format="date" month="short" value={item.title} />
```
mrfelton, line 12: Maybe this is what's causing the prop warning? What is `title`? Is it always a string or Date?

renderer/containers/UI/FormattedDateTime.js
```diff
@@ -0,0 +1,40 @@
...
+import { connect } from 'react-redux'
+import { FormattedDate, FormattedTime } from 'react-intl'
+import { settingsSelectors } from 'reducers/settings'
+
+const mapStateToProps = state => ({
+  timeDisplayMode: settingsSelectors.currentConfig(state).timeDisplayMode,
+})
+
+const FormattedDateTime = ({ timeDisplayMode, format, month = 'long', value }) => {
+  const hour12 = timeDisplayMode && timeDisplayMode === '12hour' ? 1 : 0
```
mrfelton, line 12: General style rule is that booleans should use format `isX` - I'd call this `is12Hour`

mrfelton: If you set default value in props above this can be simplified to something like:

```js
const is12Hour = timeDisplayMode === '12hour'
```

renderer/containers/UI/FormattedDateTime.js
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+import { FormattedDate, FormattedTime } from 'react-intl'
+import { settingsSelectors } from 'reducers/settings'
+
+const mapStateToProps = state => ({
+  timeDisplayMode: settingsSelectors.currentConfig(state).timeDisplayMode,
+})
+
+const FormattedDateTime = ({ timeDisplayMode, format, month = 'long', value }) => {
```
mrfelton, line 11: Probably you should set default here: `timeDisplayMode = '12hour'`

`value` is required so maybe bring it to the start of the prop list.

renderer/containers/UI/FormattedDateTime.js
```diff
@@ -0,0 +1,40 @@
...
+          year="numeric"
+        />
+      )
+  }
+}
+
+FormattedDateTime.propTypes = {
+  format: PropTypes.string,
+  month: PropTypes.string,
+  timeDisplayMode: PropTypes.string,
```
mrfelton, line 36: You can make this an enum like:

```js
timeDisplayMode: PropTypes.oneOf(['12hour', '24Hour']),
```

renderer/containers/UI/FormattedDateTime.js
```diff
@@ -0,0 +1,40 @@
...
+          month={month}
+          value={value}
+          year="numeric"
+        />
+      )
+  }
+}
+
+FormattedDateTime.propTypes = {
+  format: PropTypes.string,
```
mrfelton, line 34: Since `format` can only take specific values you can make an enum using `oneOf` and list out the specific string values.

mrfelton (changes requested):

renderer/containers/UI/FormattedDateTime.js
rahulbile: Done

renderer/containers/UI/FormattedDateTime.js
rahulbile: Done

renderer/containers/UI/FormattedDateTime.js
rahulbile: Have set default timeDisplayMode to '12hour'
linter looks for alphabetical order of prop list, so have sorted accordingly.

renderer/containers/UI/FormattedDateTime.js
rahulbile: Done.

renderer/components/Activity/Row.js
rahulbile: Yeah looks like in some cases its not a date, so have updated the value validation to be: PropTypes.any, will that be fine? Doing so the error does go away, shall we format the value before sending or any validation is fine?

mrfelton: I guess really the component supports dates passed as `string` (which is what `item.title` is). Probably should update component prop type definition to `oneOfType([PropTypes.number, PropTypes.string, PropTypes.instanceOf(Date)])`

rahulbile: Yep, adding that works fine and also fixes the validation issue. Updated the PR

mrfelton (approved):
Tested ACK https://github.com/LN-Zap/zap-desktop/pull/3688/commits/aae38385e27cac1dc06ffd3aa81aeb608c16fc74

PR #3691, opened 2020-11-14
Rb/electron update 3670

rahulbile:
Security update for electron

## How Has This Been Tested?

Manual and by running tests

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     }
   },
   "browserslist": [
-    "electron 8.5.2"
+    "electron 8.5.3"
   ],
   "engines": {
     "node": ">=12.0.0",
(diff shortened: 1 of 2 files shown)


rahulbile:
@mrfelton I tested this on new ubuntu  system, and it did threw the error as below on first try, but was able to connect later for all subsequent attempts.  Also, yarn e2e passed for connecting to the remote wallet. Can you once try to test, please? If it still fails, I think increasing https://github.com/LN-Zap/node-lnd-grpc/blob/master/src/utils/constants.js#L12 to 25 might help. Can you please trigger a build on travis?

mrfelton:
@rahulbile we're out of Travis credits atm - working on getting that renewed and will try re-running when possible. But, I have previously tried running this build on travis multiple times and it always failed.

I think it might be an idea to update lnd-grpc repo. Best thing to do might be to make it possible to override the timeouts as part of the `LndGrpc` constructor and then we can set custom values in zap-desktop. But before doing that we could try a simpler approach to just update the default timeout values.

Issue #3730, opened 2021-05-08
WalletUnlocker service is no longer available (pending lnd 0.13.0 breaking change)


musdom:
## Description

zap-desktop no longer works with lnd built from recent master branch . My zap-iOS however still works with the same lnd.

## Expected Behavior

zap-desktop connects to lnd as usual

## Actual Behavior

Displays a password prompt to unlock my already unlocked lnd wallet. When password is entered, encountered error: `2 UNKNOWN: wallet already unlocked, WalletUnlocker service is no longer available`

## Possible Fix

This merged lnd PR may be relevant: https://github.com/lightningnetwork/lnd/issues/3982

## Context

I accidentally built and ran lnd from master instead of the 0.12.1 tag, and now can't go back as it did a channel.db migration. 

## Your Environment

- Zap version: 0.7.2
- Operating System and version: Ubuntu 20.04

sortofasian:
Same issue on Mac and Windows, remote or local using prebuilt binaries of lnd 0.13.0-beta.rc2. Still works normally on mobile.

rbrooklyn:
Issue is unfortunately persisting in the latest 0.7.3 release.

mrfelton:
We do not currently support lnd 0.13 though this is being worked on.

rahulbile:
This is fixed in latest release 0.7.4-beta

PR #3743, opened 2021-06-25
Update lnd to v0.13.1-beta

rahulbile:
## Description:

Update to lnd 0.13.0 beta

## Motivation and Context:
Closes: #3730
Update lnd

## Types of changes:

Enhancement

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,8 +44,7 @@
   },
   "config": {
     "lnd-binary": {
-      "binaryVersion": "0.12.0-beta-2-g2dd779205",
-      "binarySite": "https://github.com/LN-Zap/lnd/releases/download"
+      "binaryVersion": "0.13.1-beta"
     }
   },
   "browserslist": [
@@ -255,7 +254,7 @@
     "cross-env": "7.0.2",
     "cross-spawn": "7.0.3",
     "csp-html-webpack-plugin": "4.0.0",
-    "electron": "9.4.0",
+    "electron": "9.4.4",
     "electron-builder": "22.9.1",
     "electron-devtools-installer": "3.1.1",
     "electron-notarize": "1.0.0",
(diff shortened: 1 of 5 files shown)


renderer/components/Channels/ChannelCreateForm.js
```diff
@@ -195,14 +195,14 @@ class ChannelCreateForm extends React.Component {
     const [pubkey, host] = nodePubkey.split('@')
 
     // Determine the fee rate to use.
-    const satPerByte = this.getFee()
+    const satPerVByte = this.getFee()
```
mrfelton, line 198: I don't think you can do this, because `getFee` is not returning `satPerVByte` - t's a different unit to `satPerByte`

mrfelton, line 198: I don't think you can do this, because `getFee` is not returning `satPerVByte` - its returning `satPerByte` which is a different unit to `satPerVByte`

mrfelton (changes requested):

renderer/components/Channels/ChannelCreateForm.js
rahulbile: Done, reverted.

package.json
```diff
@@ -288,7 +287,7 @@
     "husky": "4.3.0",
     "jest": "26.5.3",
     "jest-styled-components": "6.3.4",
-    "lnd-binary": "0.3.18",
+    "lnd-binary": "^0.3.22",
```
mrfelton, line 290: Should use exact version.

mrfelton (changes requested):

package.json
rahulbile: Done.

mrfelton (approved):

